PR #3, opened 2016-06-17
Prevent file-level comments from being confused with package godoc

yugui:
It improves the package documentation in https://godoc.org/github.com/bazelbuild/buildifier/core


pmbethe09:
LGTM

PR #5, opened 2016-06-24
defer acquisition of testpath

pmbethe09:
this unblocks unforking with Google internal.


damienmg:
Oh sorry I totally missed this PR. Thanks Laurent for reviewing it.

PR #8, opened 2016-07-26
fix rendering of MarkDown file

tfarina:
There is a missing ` at the end of the buildifier command that is
breaking the layout when GitHub renders the README.md file.

Instead, just use the syntax $() to fix this problem.


bazel-io:

damienmg:
@pmbethe09 this PR LGTM but I don't remember the workflow here, can we just squash the PR?

pmbethe09:
LGTM.

(sorry I was away).

PR #10, opened 2016-08-18
Allow forcing multiline for dict and tuple comprehensions

nesciens:
Also:
* Add .gitignore, as spurious files started appearing in `git status`.
* Fix a panic in the test caused by using `[:len(".txt")]` in the testdata, which creates a zero value for `reflect` on which `Type()` cannot be called.

It'd be nice if long comprehensions would be broken automatically, but being able to force this takes away the immediate pain.

### DIFF
--- a/core/parse.y
+++ b/core/parse.y
@@ -255,6 +255,7 @@ expr:
 	}
 |	'(' expr for_clauses if_clauses_opt ')'
 	{
+		exprStart, _ := $2.Span()
 		$$ = &ListForExpr{
 			Brack: "()",
 			Start: $1,
@@ -262,10 +263,12 @@ expr:
 			For: $3,
 			If: $4,
 			End: End{Pos: $5},
+			ForceMultiLine: $1.Line != exprStart.Line,
 		}
 	}
 |	'{' keyvalue for_clauses if_clauses_opt '}'
 	{
+		exprStart, _ := $2.Span()
 		$$ = &ListForExpr{
 			Brack: "{}",
 			Start: $1,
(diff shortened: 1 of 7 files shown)


bazel-io:

core/parse.y
```diff
@@ -255,24 +255,28 @@ expr:
...
 	{
+		exprStart, _ := $2.Span()
 		$$ = &ListForExpr{
 			Brack: "()",
 			Start: $1,
 			X: $2,
 			For: $3,
 			If: $4,
 			End: End{Pos: $5},
+			ForceMultiLine: $1.Line != exprStart.Line,
```
pmbethe09: should this also have a line-length component?  If I have a short but wrapped line I would like it to be made smaller
[
do_this
for f in files
]

--- should become ---
[do_this for f in files]

nesciens: Automatically collapsing or expanding depending on line-length would indeed be a natural feature request. However, this seems too complicated for a drive-by contributor (i.e. me) to implement as heuristics in the presence of many possible line-breaks are likely to get complicated. In the test I have used long lines to make sure that it does not break if such a feature gets introduced.

Currently, as far as I can tell, () and {} comprehensions never get newlines after (/{, before "for" and before )/}. With this change, the user can request those newlines by adding a newline after the opening bracket.

You can have buildifier do the change from the example by deleting the newline after [.

Note that this change does not affect the behavior of []-comprehensions (including the behavior surrounding the example). It merely extends the current behavior for [... for ....] to (... for ...) and {...: ... for ...}.

pmbethe09:
Sounds good LGTM

PR #13, opened 2016-10-10
Fix bug where 'x = a if cond else b' was incorrectly parsed

pmbethe09:
Added a test to confirm.

### DIFF
--- a/core/testdata/045.golden
+++ b/core/testdata/045.golden
@@ -1,0 +1,10 @@
+[
+    cc_library(
+        name = "Prefix" + suffix,
+        deps = [dep] if dep else [],
+    )
+    for suffix, dep in [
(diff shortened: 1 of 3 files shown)


core/testdata/045.golden
```diff
@@ -0,0 +1,10 @@
+[
```
niltonvolpato: Maybe you should also add a test case checking that "if a = b: ..." remains the same

pmbethe09: Laurent reminded me that 
"if a = b: ..."
is not valid in Python

niltonvolpato: Oh, true :)

laurentlb:
LGTM

PR #18, opened 2016-10-17
Reference rules_go using reliable http_archive()

jart:
Fetching via git is slow, buggy, and wastes disk space. It caused a problem for a TensorFlow user in tensorflow/tensorflow#4999. Using http_archive is the optimal solution. I've slowly but surely been updating all the Bazel projects to follow this practice. I've also taken the liberty of mirroring this file for additional reliability.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -11,3 +11,4 @@
 Paul Bethe <pbethe@google.com>
 Laurent Le Brun <laurentlb@google.com>
 Nilton Volpato <nilton@google.com>
+Justine Alexandra Roberts Tunney <jart@google.com>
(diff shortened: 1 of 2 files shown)


pmbethe09:
This seems OK, although I would like a TODO to switch back if and when bazel switches from using jGit to git directly.
Also, rules_go needs a new version tag -- would you mind holding of a day or so to pick up that new tag?

jart:
Why would you switch back to using git_repository? If Bazel switches from jgit to git, it'll be more reliable, but it will probably go even slower.

jart:
If it were up to me, honestly, I would remove git_repository() from Bazel entirely.

pmbethe09:
This morphed into an internal discussion, but the summary was: http_archive
is better, but there are reasons to keep git_repository for development.

So http_archive, SGTM.

PR #19, opened 2016-10-21
Check in parse.y.go, and add test to ensure freshness

pmbethe09:
Satisfy https://github.com/bazelbuild/buildifier/issues/16 and
https://github.com/bazelbuild/buildifier/issues/14 by checking in
parse.y.go


laurentlb:
LGTM

PR #25, opened 2016-12-12
Set buildifier:buildifier to //visibility:public

### LINKED ISSUE #21: Set //buildifier:buildifier to visibility:public for bazel
In `buildifier/BUILD`, could we set `visibility = ["//visibility:public"]` on the `go_binary(name = "buildifier", ...)` rule?

I would like to bring the `buildifier` project into my `WORKSPACE` and refer to that binary as a dependency, but the target defaults to private (at least when using `bazel`).

jwnimmer-tri:
Closes #21.


bazel-io:

pmbethe09:
LGTM.

I thought we already did this, but thanks for doing this.

PR #30, opened 2016-12-21
remove old go:generate directive which is broken in go1.8

pmbethe09:


pmbethe09:
This should unblock
https://github.com/bazelbuild/rules_go/pull/226

PR #49, opened 2017-02-12
Make buildifier and buildozer compile on Windows

asimshankar:
Without this commit, a build on Windows fails with:

```
differ\diff.go:124 undefined: syscall.Stat_t
```

I had initially tried to use two files and avoid having one file per OS:

1. `isatty_other.go` with `// +build !windows`
2. `isatty_windows.go`

But [build constraints](https://golang.org/pkg/go/build/#hdr-Build_Constraints) require that they be preceded only by line comments or blank lines, they cannot be preceded by the [general comments](https://golang.org/ref/spec#Comments) (`/* ... */`) which are used in the files in this repository.

### DIFF
--- a/differ/isatty_darwin.go
+++ b/differ/isatty_darwin.go
@@ -1,0 +1,30 @@
+/*
+Copyright 2016 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package differ
+
+import "syscall"
+
+// isatty reports whether fd is a tty.
+// Actually it reports whether fd is a character device, which is close enough.
+func isatty(fd int) bool {
+	var st syscall.Stat_t
+	err := syscall.Fstat(fd, &st)
+	if err != nil {
+		return false
+	}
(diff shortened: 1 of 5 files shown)


bazel-io:

differ/isatty_darwin.go
```diff
@@ -0,0 +1,30 @@
...
+
+package differ
+
+import "syscall"
+
+// isatty reports whether fd is a tty.
+// Actually it reports whether fd is a character device, which is close enough.
+func isatty(fd int) bool {
+	var st syscall.Stat_t
+	err := syscall.Fstat(fd, &st)
```
pmbethe09: push into if statement

differ/isatty_darwin.go
```diff
@@ -0,0 +1,30 @@
+/*
+Copyright 2016 Google Inc. All Rights Reserved.
```
pmbethe09: s/2016/2017/ here and throughout the PR

differ/isatty_linux.go
```diff
@@ -0,0 +1,30 @@
...
+
+package differ
+
+import "syscall"
+
+// isatty reports whether fd is a tty.
+// Actually it reports whether fd is a character device, which is close enough.
+func isatty(fd int) bool {
+	var st syscall.Stat_t
+	err := syscall.Fstat(fd, &st)
```
pmbethe09: push into if statement

pmbethe09:
Seems fine to me, as to the comment about build constraints -- I think it should be OK to move the copyright notice to after the 'package' and build constraint lines.  But this is also fine.

differ/isatty_linux.go
asimshankar: Done

differ/isatty_darwin.go
asimshankar: Done

differ/isatty_darwin.go
asimshankar: Done

asimshankar:
Addressed the comments, PTAL

PR #50, opened 2017-02-12
Windows: Use CRLF instead of LF when writing files out.

asimshankar:
Honestly, I'm not sure if this is the right thing but just for background:

- Git installations on Windows are typically configured to replace LF with CRLF (See [`core.autocrlf`](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration))
- Thus runs of `buildifier` on Windows without this change produce spurious differences

I suspect using `bazel`/`buildifier` on Windows is typically done in git repositories, but that may not be always true and it is still possible to configure `git` differently. Hence I'm a bit unsure of this myself.

An alternative would be to introduce a flag to switch between LF and CRLF.

### DIFF
--- a/build/print.go
+++ b/build/print.go
@@ -20,8 +20,16 @@ package build
 import (
 	"bytes"
 	"fmt"
+	"runtime"
 	"strings"
 )
+
+var breakLine = func() string {
+	if runtime.GOOS == "windows" {
+		return "\r\n"
+	}
+	return "\n"
+}()
 
 // Format returns the formatted form of the given BUILD file.
 func Format(f *File) []byte {
@@ -76,7 +84,7 @@ func (p *printer) newline() {
 		for i, com := range p.comment {
 			if i > 0 {
 				p.trim()
-				p.printf("\n%*s", p.margin, "")
+				p.printf("%s%*s", breakLine, p.margin, "")
 			}
 			p.printf("%s", strings.TrimSpace(com.Token))
 		}
@@ -84,7 +92,7 @@ func (p *printer) newline() {
 	}
 
 	p.trim()
-	p.printf("\n%*s", p.margin, "")
+	p.printf("%s%*s", breakLine, p.margin, "")
 }
 
 // breakline breaks the current line, inserting a continuation \ if needed.


bazel-io:

pmbethe09:
The problem with this, is it will create a lot of churn -- anytime I edit a BUILD file on linux that was last edited on Windows will change all of the line-endings and vice-versa.  This makes PRs appear to do a lot more than they really are doing.

asimshankar:
FWIW, without this is when I was getting a lot of churn because of how it seems that in "default" git installations on Windows, `core.autocrlf` is true.

In other words, without this change, if I ran `buildifier` on my repository on Windows, every BUILD file changed, when none should have. With this change there is no churn.

However, if `core.autocrlf` is false, then the reverse would happen.
Would you suggest introducing a flag instead?

pmbethe09:
So if git replaces LF with CRLF, what happens at code-review time?  Do all
of the changed line-endings show up as changes?

asimshankar:
I believe the idea is that on the filesystem it is CRLF but in the git index it is LF. See https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration (and look for the section on `core.autocrlf`). 

(That said, I don't have enough experience with Windows. This PR is based on documentation for `core.autocrlf` and what happened when I installed `git` on a Windows machine and ran `buildifier` on TensorFlow sources)

pmbethe09:
I also am not a Windows expert.

@damienmg @laurentlb -- can you comment on this?

laurentlb:
@dslomov @laszlocsomor

laszlocsomor:
I can't add anything that hasn't been said already: Windows uses CRLF endings, which can be a problem if Buildifier sticks to just LF. How about Buildifier detects the line ending for the file (if it's mixed, then uses the ending that's in majority), and uses that ending for every line it changes?

pmbethe09:
Jenkins, test this please.

damienmg:
test this please

(for some weird reason the testing was working but the PR wasn't getting the update)

damienmg:
@googlebot is not updating that PR. Why is the buildifier repository not inheriting from the bazelbuild org?

jchv:
My recommendation would be to autodetect the presence of CR/LF line endings in the input file, and if they exist, use CR/LF line endings. That way, it doesn't matter what Git settings the user has, or even if they are using Git at all - it just won't mess with line endings.

laurentlb:
It's also nice to use one single format for all BUILD files.

Related discussion: https://github.com/golang/go/issues/16355
gofmt outputs only LF and it seems to work reasonably well for them.

PR #51, opened 2017-02-15
Update setup instructions

bzz:
In case if `go` is already installed `go get` could be used to setup local buildifier


bazel-io:

PR #53, opened 2017-02-22
Fix the buildifier bazel build by disabling Windows on Bazel

jhseu:
Bazel support for Windows is experimental anyway. Was broken by https://github.com/bazelbuild/buildifier/pull/49

Had to redo comment styles for build constraints to work:
https://golang.org/pkg/go/build/#hdr-Build_Constraints

@asimshankar fyi


asimshankar:
So, just so I understand, without this change `bazel test //buildifier/...` works on Windows/Linux/OSX, but `bazel build //buildifier:buildifier` fails on Windows.

With this change:
- `bazel test //buildifier/...` and `bazel build //buildifier:buildifier` will no longer work on Windows, but
- `go get github.com/bazelbuild/buildifier/buildifier` will work on Windows/Linux/OSX

Correct?

(I have no concerns, just making sure I understand)

jhseu:
Yep

Issue #55, opened 2017-03-02
Make "BUILD" being the name of BUILD files configurable


swolchok:
Buck thinks that its input files should be called `BUCK`. Since buildozer also doesn't seem to have a way to accept a BUILD file on stdin (and seems to want to know the target to modify, not just the BUILD file path), this makes using buildozer with appropriately-formatted `BUCK` files difficult.

pmbethe09:
not sure I follow.  You can call:
`buildifier some/path/BUCK other/path/BUCK `
and it will run on those files.

swolchok:
> `buildifier some/path/BUCK other/path/BUCK`

Does the same hold for buildozer, not buildifier? It didn't seem to, but it's possible I was doing it wrong. `buildozer path/to/BUCK "add some_attribute True` gave me an error like `path/to/BUCK/BUILD: stat path/to/BUCK/BUILD: not a directory`, as did `buildozer path/to/BUCK :some_target_in_the_buck_file "add some_attribute True"`.

pmbethe09:
... oops ... you wrote 'buildozer' and I read 'buildifier'.

You are absolutely right, 'BUILD' is hardcoded in several places here:
https://github.com/bazelbuild/buildifier/blob/master/edit/buildozer.go

and will need to get changed to allow for 'BUILD.bazel' and 'BUCK'

swolchok:
FWIW, I think it's also hardcoded in https://github.com/bazelbuild/buildifier/blob/master/edit/edit.go, so I'm not sure how buildifier works with `BUILD.bazel` and `BUCK` either. :\

swolchok:
Is this an issue the buildifier team is interested in fixing, or should I put a PR together? If the latter, does a command-line flag seem like the way to go?

pmbethe09:
Instead of a flag, why not just look for suffixes in order: {BUILD.bazel,
BUILD, BUCK}

swolchok:
Should we close this?

PR #58, opened 2017-03-08
Add BUILD.bazel & BUCK as valid BUILD file names. (#55)

swolchok:
This is intended to fix issue #55. It's the first time I've written Go, so I'm sorry if it's not up to snuff.

### DIFF
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -673,6 +673,8 @@ func getGlobalVariables(exprs []build.Expr) (vars map[string]*build.BinaryExpr) 
 	return vars
 }
 
+var buildFileSuffixes = [...]string{"BUILD.bazel", "BUILD", "BUCK"}
+
 // rewrite parses the BUILD file for the given file, transforms the AST,
 // and write the changes back in the file (or on stdout). g4 edit is run
 // automatically if the file is not writeable.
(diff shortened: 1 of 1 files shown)


bazel-io:

swolchok:
> If you signed the CLA as a corporation, please let us know the company's name.

This contribution is being made on behalf of my employer, Facebook.

edit/buildozer.go
```diff
@@ -688,10 +690,18 @@ func rewrite(commandsForFile commandsForFile) *rewriteResult {
 			return &rewriteResult{file: name, errs: []error{err}}
 		}
 	} else {
-		// read a file
-		data, fi, err = file.ReadFile(name)
+		origName := name
+		name = strings.TrimSuffix(name, "BUILD")
```
pmbethe09: should this trim all of the possible suffixes?

pmbethe09:
LGTM with one question

edit/buildozer.go
swolchok: I don't think so. By patching here, we let the rest of the code continue to assume that BUILD files are called `BUILD`. We trim off each suffix to go to the next suffix on line 701.

If we want to avoid letting the rest of the code assume that BUILD files are called `BUILD`, that would seem to entail a substantial reworking of the flow of the tool. I tried that approach originally, but it was much more intensive and seemed to necessitate moving `ReadFile` calls earlier in the flow (so we know which suffix to use).

PR #60, opened 2017-03-10
Update copy of generated proto for 'go build' folks.

pmbethe09:
Fixes broken build: http://ci.bazel.io/job/buildifier/


laurentlb:
LGTM

PR #63, opened 2017-03-14
Shorten @foo//:foo labels to @foo

jart:
This convenient syntax been available in Bazel since November. See https://github.com/bazelbuild/bazel/commit/61affe77f9f5e5e7fcbbba7d8c2bcf8fbab776d5

Attn: @pmbethe09


build/rewrite.go
```diff
@@ -194,9 +195,9 @@ func fixLabels(f *File, info *RewriteInfo) {
 		*p = str1
 	}
 
-	// labelRE matches label strings //x/y/z:abc.
-	// $1 is //x/y/z, $2 is x/y/, $3 is z, $4 is :abc, and $5 is abc.
-	labelRE := regexp.MustCompile(`^(//(.*/)?([^:]+))(:([^:]+))?$`)
+	// labelRE matches label strings, e.g. @r//x/y/z:abc
+	// where $1 is @r//x/y/z, $2 is r, $3 is z, $4 is abc.
+	labelRE := regexp.MustCompile(`^((?:@(\w+))?//(?:.*/)?([^:]*))(?::([^:]+))?$`)
```
pmbethe09: Can you move this labelRE to a file level var?  
previous bug: MustCompile should only be used in vars...

bolinfest:
Does this honor `-buildifier_disable=label`?

build/rewrite.go
jart: Done.

laurentlb:
Thanks, LGTM

PR #64, opened 2017-03-14
Use redundant download URLs

jart:
This improves reliability and performance because GitHub goes down frequently.


pmbethe09:
Very nice, LGTM.

PR #65, opened 2017-03-14
Handle "-" as stdin/stdout mode 

rishiloyola:
Resolve  TODO with 'bazel-team' tag.


bazel-io:

buildifier/buildifier.go
```diff
@@ -136,9 +136,7 @@ func main() {
 
 	diff = differ.Find()
 
-	// TODO(bazel-team): Handle "-" as stdin/stdout mode too.
-
-	if len(args) == 0 {
+	if (len(args) == 0) || (len(args) == 1 && args[0] == "-") {
```
pmbethe09: you shouldn't need the leftmost '(' ')' as == is higher precedence than ||

rishiloyola: Yes, I am aware of that. Usually, I put conditions under the brackets.
Do you want to remove it?

pmbethe09: yes, the right most paren clarify the precedence of || and && so are needed, but the leftmost are extraneous and should be removed.

pmbethe09:
Jenkins, test this please.

buildifier/buildifier.go
rishiloyola: Done.

PR #66, opened 2017-03-15
Cmd concat

swolchok:
The `add` command assumes that its argument is an int or string and should be added as a member of the target list. I wanted to be able to concatenate an identifier that contains a list value to a target list, so I added a similar (and implementation-sharing) `concat` command. See the new tests for examples of how it works.

A point against the command as implemented is that it's not idempotent. The test on line 156 of edit_test.go illustrates this. I elected to leave it dumb as-is rather than being half-clever and refusing to concatenate the same constant if it's anywhere in the list. This is only *half*-clever because buildozer doesn't necessarily know the contents of the constant and can't fully guarantee it's not duplicating things in the list. (It's also possible that order matters in the list and therefore you don't *want* to duplicate things...)

### DIFF
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -74,17 +74,30 @@ type CmdEnvironment struct {
 
 // The cmdXXX functions implement the various commands.
 
-func cmdAdd(env CmdEnvironment) (*build.File, error) {
+func addImpl(env CmdEnvironment, concat bool) (*build.File, error) {
 	attr := env.Args[0]
 	for _, val := range env.Args[1:] {
-		if IsIntList(attr) {
-			AddValueToListAttribute(env.Rule, attr, env.Pkg, &build.LiteralExpr{Token: val}, &env.Vars)
-			continue
+		if concat {
+			expr := &build.LiteralExpr{Token: val}
+			ConcatListValueToListAttribute(env.Rule, attr, env.Pkg, expr, &env.Vars)
+		} else {
+			if IsIntList(attr) {
+				AddValueToListAttribute(env.Rule, attr, env.Pkg, &build.LiteralExpr{Token: val}, &env.Vars)
+				continue
+			}
+			expr := &build.StringExpr{Value: ShortenLabel(val, env.Pkg)}
+			AddValueToListAttribute(env.Rule, attr, env.Pkg, expr, &env.Vars)
 		}
-		strVal := &build.StringExpr{Value: ShortenLabel(val, env.Pkg)}
-		AddValueToListAttribute(env.Rule, attr, env.Pkg, strVal, &env.Vars)
 	}
 	return env.File, nil
+}
+
+func cmdConcat(env CmdEnvironment) (*build.File, error) {
+	return addImpl(env, true)
+}
+
+func cmdAdd(env CmdEnvironment) (*build.File, error) {
+	return addImpl(env, false)
 }
 
 func cmdComment(env CmdEnvironment) (*build.File, error) {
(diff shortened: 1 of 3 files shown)


bazel-io:

swolchok:
> CLA

This contribution is being made on behalf of my employer, Facebook.

pmbethe09:
Jenkins, test this please.

edit/buildozer.go
```diff
@@ -74,19 +74,32 @@ type CmdEnvironment struct {
...
-func cmdAdd(env CmdEnvironment) (*build.File, error) {
+func addImpl(env CmdEnvironment, concat bool) (*build.File, error) {
 	attr := env.Args[0]
 	for _, val := range env.Args[1:] {
-		if IsIntList(attr) {
-			AddValueToListAttribute(env.Rule, attr, env.Pkg, &build.LiteralExpr{Token: val}, &env.Vars)
-			continue
+		if concat {
+			expr := &build.LiteralExpr{Token: val}
+			ConcatListValueToListAttribute(env.Rule, attr, env.Pkg, expr, &env.Vars)
```
pmbethe09: use continue to avoid else blocks when possible.

So here, continue, then the else wrapper goes away

edit/edit.go
```diff
@@ -544,17 +544,37 @@ func AddValueToList(oldList build.Expr, pkg string, item build.Expr, sorted bool
...
 
-// AddValueToListAttribute adds the given item to the list attribute identified by name and pkg.
-func AddValueToListAttribute(r *build.Rule, name string, pkg string, item build.Expr, vars *map[string]*build.BinaryExpr) {
+// AddListValueToList adds a value (assumed to be of list type) to a
+// list.
+func addListValueToList(oldList build.Expr, pkg string, item build.Expr, sorted bool) build.Expr {
+	if oldList == nil {
+		return item
+	}
+	concat := &build.BinaryExpr{Op: "+", X: oldList, Y: item}
```
pmbethe09: just return this, avoid the variable

edit/edit.go
```diff
@@ -544,17 +544,37 @@ func AddValueToList(oldList build.Expr, pkg string, item build.Expr, sorted bool
...
+	return concat
+
```
pmbethe09: remove blankline

edit/edit_test.go
```diff
@@ -112,3 +112,68 @@ load("other loc", "symbol")`,
...
+		}
+		rule := bld.RuleAt(1)
+		AddValueToListAttribute(rule, "attr", "", &build.StringExpr{Value: "foo"}, nil)
+		got := strings.TrimSpace(string(build.Format(bld)))
+		expectedBld, err := build.Parse("BUILD", []byte(tst.expected))
+		if err != nil {
+			t.Error(err)
+			continue
+		}
+		expected := strings.TrimSpace(string(build.Format(expectedBld)))
```
pmbethe09: Go-style is "want" for the thing you want as opposed to Java's "expectedXXX".

You can also push this into the if

edit/edit_test.go
```diff
@@ -112,3 +112,68 @@ load("other loc", "symbol")`,
...
+		}
+		rule := bld.RuleAt(1)
+		ConcatListValueToListAttribute(rule, "attr", "", &build.LiteralExpr{Token: "SOME_LIST"}, nil)
+		got := strings.TrimSpace(string(build.Format(bld)))
+		expectedBld, err := build.Parse("BUILD", []byte(tst.expected))
+		if err != nil {
+			t.Error(err)
+			continue
+		}
+		expected := strings.TrimSpace(string(build.Format(expectedBld)))
```
pmbethe09: want

laurentlb:
Do you have use-cases for this? In general, we don't know which variables are defined in the same file.
I prefer to encourage a style where everything is explicitly listed. Using variables often makes code transformations more complicated (e.g. you want to remove a dependency from a rule, but the dependency comes from a variable). I wonder how often users will need this function.

swolchok:
> In general, we don't know which variables are defined in the same file.

Users can always ensure that a variable is defined by using the `new_load` command before running `concat`if necessary, can't they?

> I prefer to encourage a style where everything is explicitly listed. Using variables often makes code transformations more complicated (e.g. you want to remove a dependency from a rule, but the dependency comes from a variable)

I agree. Unfortunately, I live in a world that is more complicated and need to do automated transformations anyway. That said, I've only had a need for `concat` once so far, so if the value to the broader project isn't clear, I'm happy to sit on it for the time being. Would you like a separate PR for the `AddValueToListAttribute` tests?

laurentlb:
I think you can send a separate PR for the tests. Thank you for that!

Later, we may want to add your new command. But I think we should have a distinction between core commands (a simple set of commands to address most common cases) and other commands. Having too many commands on the same level can be distracting and look complicated for a new user.

PR #72, opened 2017-03-23
Fixed bug where buildifier swallowed up some comments.

### LINKED ISSUE #68: Buildifier removes some comments
[   # This line will be eaten by buildifier
    # but not this line.
    cc_binary(
        name = "something_%s_useful" % lang,
    )
    for lang in li]

forumulator:
Fixes #68 on github
```
[cc_binary(
    name = "translator_for_%s" % lang,
) for lang in li]

[   # This line will be eaten by buildifier
    # but not this line.
    cc_binary(
        name = "translate_%s_benchmark" % lang,
    )
    for lang in li]
```
This is due to the comment line `# This line will be eaten by buildifier` being assigned to the expression `cc_binary() for lang in li` and then never printed.

@laurentlb @pmbethe09 I thought the comment assignment a bit strange in buildifier. For example, given the input

```
[func1() for j in list2]

[  #This comment gets associated with the list above
func3()
]
```
This is due to the code [here](https://github.com/bazelbuild/buildifier/blob/master/build/lex.go#L733-L743). Thoughts and reviews?

### DIFF
--- a/build/lex.go
+++ b/build/lex.go
@@ -725,9 +725,10 @@ func (in *input) assignComments() {
 		// Do not assign suffix comments to call, list, end-of-list,
 		// whole file, or conditional expression.
 		// Instead assign them to the last argument, element, or rule.
+		// IfClause and ForClause because in single line mode, [a for b in c]
+		// the suffix comments assigned to these are swallowed up.
 		switch x.(type) {
-		case *CallExpr, *ListExpr, *End, *File, 
-			*ConditionalExpr, *IfClause, *ForClause:
+		case *CallExpr, *ListExpr, *End, *File, *ConditionalExpr, *IfClause, *ForClause:
 			continue
 		}
 
--- a/build/lex.go
+++ b/build/lex.go
@@ -726,7 +726,8 @@ func (in *input) assignComments() {
 		// whole file, or conditional expression.
 		// Instead assign them to the last argument, element, or rule.
 		switch x.(type) {
-		case *CallExpr, *ListExpr, *End, *File, *ConditionalExpr:
+		case *CallExpr, *ListExpr, *End, *File, 
+			*ConditionalExpr, *IfClause, *ForClause:
 			continue
 		}
 
--- a/build/print.go
+++ b/build/print.go
@@ -121,7 +121,16 @@ func (p *printer) file(f *File) {
 	for i, stmt := range f.Stmt {
 		switch stmt := stmt.(type) {
 		case *CommentBlock:
-			// comments already handled
+			// This is necessary so that CommentBlock
+			// doesn't swallow up suffix comments of
+			// the expression after it.
+			for _, scom := range stmt.Suffix {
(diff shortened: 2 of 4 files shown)


bazel-io:

build/lex.go
```diff
@@ -726,7 +726,8 @@ func (in *input) assignComments() {
 		// whole file, or conditional expression.
 		// Instead assign them to the last argument, element, or rule.
 		switch x.(type) {
-		case *CallExpr, *ListExpr, *End, *File, *ConditionalExpr:
+		case *CallExpr, *ListExpr, *End, *File, 
+			*ConditionalExpr, *IfClause, *ForClause:
```
pmbethe09: no need to line break -- no limit in Go
https://github.com/golang/go/wiki/CodeReviewComments#line-length

build/print.go
```diff
@@ -121,7 +121,16 @@ func (p *printer) file(f *File) {
 	for i, stmt := range f.Stmt {
 		switch stmt := stmt.(type) {
 		case *CommentBlock:
-			// comments already handled
+			// This is necessary so that CommentBlock
```
pmbethe09: This does feel too long for 1 comment line, but maybe break into 2 instead of 3?

forumulator:
@pmbethe09 Sorry for the delay, was busy with the proposal. Did the styling changes you asked for.

As for the test, what kind of general test would be good here? One thing that comes to mind is to take a big BUILD file, add comment after each line and at end of each line, and check if the comments are retained after formatting. Is this okay?

forumulator:
Sorry, please ignore my last comment. I've added the test.

pmbethe09:
Jenkins, test this please.

forumulator:
@pmbethe09 Build's failing on CI, for this as well as for #75 that you gave. It looks like the problem is the BUILD.bazel file you removed from the root(see [here](http://ci.bazel.io/job/PR-buildifier/43/BAZEL_VERSION=latest,PLATFORM_NAME=linux-x86_64/console)). Could you please take a look?

pmbethe09:
Jenkins, test this please.

forumulator:
@pmbethe09 Seems like the build is passing now, after you re-added the BUILD file. Can you merge this, please(so that I can update my master finally)?

pmbethe09:
Jenkins, test this please.

laurentlb:
I'm surprised by the test.
Why does "# Bug if this comment get eaten by buildifier" move up to the previous statement?
Why does "# Comment block should not swallow this up" move up to the previous comment?
Is there any way to keep the comments in the same instruction?

I'd expect:
```
[
    # Bug if this comment get eaten by buildifier
    # This line should be fine
    cc_binary(
        name = "translate_%s_benchmark" % lang,
    )
    for lang in li
]

```
and:

```
[
    #  Comment block should not swallow this up
    cc_binary(
        name = "%s_binary" % lang,
    )
    for lang in li
]
```

PR #77, opened 2017-04-06
Import from Google: unused_deps script for trimming deps from java_library

pmbethe09:


config/config.go
```diff
@@ -0,0 +1,25 @@
+/*
+Copyright 2017 Google Inc. All Rights Reserved.
```
damienmg: Fix indentation of the header.

unused_deps/unused_deps.go
```diff
@@ -0,0 +1,293 @@
+/*
+Copyright 2017 Google Inc. All Rights Reserved.
```
damienmg: Here too

config/config.go
pmbethe09: Hmm, I just matched the copyright in the other files -- not to say those were accurate.

Fixed?

damienmg: Better yes

unused_deps/unused_deps.go
```diff
@@ -0,0 +1,295 @@
...
+// to remove that entry from the deps attribute of the rule identified by label.
+// Returns true if at least one command was printed, or false otherwise.
+func printCommands(label string, deps map[string]bool) (anyCommandPrinted bool) {
+	buildFileName, pkg, ruleName := edit.InterpretLabel(label)
+	depsExpr := getDepsExpr(buildFileName, ruleName)
+	for _, li := range edit.AllLists(depsExpr) {
+		for _, elem := range li.List {
+			for dep := range deps {
+				str, ok := elem.(*build.StringExpr)
+				if ok && edit.ShortenLabel(str.Value, pkg) == edit.ShortenLabel(dep, pkg) {
```
shvar: Maybe instead if `==` it make sense to use`strings.Contains(edit.ShortenLabel(dep, pkg), edit.ShortenLabel(str.Value, pkg))`? Sometimes, in BUILD file people use shorter versions of labels. For example, I recently faced a case, where `str.Value = org_apache_commons_commons_lang3//jar` and `dep = @@org_apache_commons_commons_lang3//jar:jar`. 

Or maybe it make sense instead of `.ShortenLabel` use some modified version of `.LabelsEqual`?

pmbethe09: Seems reasonable -- want to take a stab at it?

side-bar: we have internal tests, but need some tests here.  I will add a basic one, but feel free to add more.

PR #83, opened 2017-04-13
modify LabelsEqual for unused_deps and some tests

shvar:
@pmbethe09, I modified a bit LabelsEqual in order to deal with `.jar-2.params` files notation.

Assuming, that I have a `maven_jar` defined in a WORKSPACE like:
```
maven_jar(
    name = "org_apache_commons_commons_lang3",
    artifact = "org.apache.commons:commons-lang3:3.1",
    sha1 = "905075e6c80f206bbe6cf1e809d2caa69f420c76",
)
```
for every direct dep on it in BUILD file I got in params file following string: `@@org_apache_commons_commons_lang3//jar:jar`

I have two problems with this
1) leading `@@`, that I think should be trimmed (and that's the reason of modifying LabelEquals)
2) `:jar` suffix, that is usually omit in BUILD files. Here, I am not sure about what to do and lean to force administratively demand writing full label name with :jar extension.

WDYT?

### DIFF
--- a/edit/edit_test.go
+++ b/edit/edit_test.go
@@ -75,7 +75,6 @@ var labelsEqualTests = []struct {
 }{
 	{"//devtools/buildozer:rule", "rule", "devtools/buildozer", true},
 	{"//devtools/buildozer:rule", "rule:jar","devtools", false},
-	{"@@io_bazel", "@io_bazel","devtools", true},
 }
 
 func TestLabelsEqual(t *testing.T) {
--- a/edit/edit_test.go
+++ b/edit/edit_test.go
@@ -68,10 +68,10 @@ func TestShortenLabel(t *testing.T) {
 }
 
 var labelsEqualTests = []struct {
-	label1 string
-	label2 string
-	pkg    string
-	result bool
+	label1   string
+	label2   string
+	pkg      string
+	expected bool
 }{
 	{"//devtools/buildozer:rule", "rule", "devtools/buildozer", true},
 	{"//devtools/buildozer:rule", "rule:jar","devtools", false},
@@ -79,10 +79,10 @@ var labelsEqualTests = []struct {
 
 func TestLabelsEqual(t *testing.T) {
 	for i, tt := range labelsEqualTests {
-		result := LabelsEqual(tt.label1, tt.label2, tt.pkg)
-		if result != tt.result {
+		got := LabelsEqual(tt.label1, tt.label2, tt.pkg)
+		if got != tt.expected {
 			t.Errorf("%d. LabelsEqual(%q, %q, %q) => %q, want %q",
-				i, tt.label1, tt.label2, tt.pkg, result, tt.result)
+				i, tt.label1, tt.label2, tt.pkg, got, tt.expected)
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


bazel-io:

shvar:
I signed it!

pmbethe09:
Jenkins, test this please.

pmbethe09:
Is @@ even legal in bazel?  Where does that come from?

shvar:
I don't think, it's a legal name for bazel label, but I see them in bazel-bin/libruleName.jar-2.params. Do you think it is actually a bug somewhere in the code, that generates that file?

pmbethe09:
That does sound like a bug -- @damienmg any ideas why the .jar-2.params would have a double @ in it?

damienmg:
Yes because @XXX is the syntax for params file for javabuilder. @@ is the way to escape that, we should de-dup the @@ when reporting in jdeps. /cc @cushon

cushon:
@damienmg I sent you a fix.

damienmg:
Wow thanks for the fast turn around!

shvar:
Wow, that was fast, thank you all!

So, @pmbethe09, I reverted logic of trimming @@ in favor of @cushon 's fix. Now it makes even more sense to use LabelsEqual in unused_deps.

edit/edit_test.go
```diff
@@ -67,6 +67,26 @@ func TestShortenLabel(t *testing.T) {
 	}
 }
 
+var labelsEqualTests = []struct {
+	label1 string
+	label2 string
+	pkg    string
+	result bool
```
pmbethe09: in our Go-style this is often called 'want' or 'expected

edit/edit_test.go
```diff
@@ -67,6 +67,26 @@ func TestShortenLabel(t *testing.T) {
...
+}{
+	{"//devtools/buildozer:rule", "rule", "devtools/buildozer", true},
+	{"//devtools/buildozer:rule", "rule:jar","devtools", false},
+}
+
+func TestLabelsEqual(t *testing.T) {
+	for i, tt := range labelsEqualTests {
+		result := LabelsEqual(tt.label1, tt.label2, tt.pkg)
```
pmbethe09: s/result/got/  (our Go-style for tests)

pmbethe09:
Jenkins, test this please.

edit/edit_test.go
shvar: sure, fixed both

unused_deps/unused_deps.go
```diff
@@ -221,11 +221,11 @@ func printCommands(label string, deps map[string]bool) (anyCommandPrinted bool)
 		for _, elem := range li.List {
 			for dep := range deps {
 				str, ok := elem.(*build.StringExpr)
-				if ok && edit.ShortenLabel(str.Value, pkg) == edit.ShortenLabel(dep, pkg) {
+				if ok && edit.LabelsEqual(str.Value, dep, pkg) {
```
pmbethe09: now that bazel is fixing the '@@' escaping I have forgotten why this change is necessary anymore?

shvar: Now it's more about cosmetics and using in buildozer exact label, as specified in BUILD file.

pmbethe09:
Jenkins, test this please.

PR #84, opened 2017-04-17
Support editing and formatting calls from structs

kleerwater:
Update buildifier and buildozer to work with calls that are composed entirely of dot accesses beginning from a literal expressions.

### DIFF
--- a/build/testdata/024.golden
+++ b/build/testdata/024.golden
@@ -4,8 +4,8 @@
 str.rewrite(".", "/")
 
 str.rewrite(
-    before = ".",
     after = "/",
+    before = ".",
 )
 
 x = (1, 2, 3)
--- a/build/testdata/024.golden
+++ b/build/testdata/024.golden
@@ -4,8 +4,8 @@
 str.rewrite(".", "/")
 
 str.rewrite(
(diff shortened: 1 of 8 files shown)


bazel-io:

kleerwater:
I signed it!

pmbethe09:
Jenkins, test this please.

build/testdata/024.golden
```diff
@@ -4,8 +4,8 @@
 str.rewrite(".", "/")
 
 str.rewrite(
-    before = ".",
     after = "/",
+    before = ".",
```
pmbethe09: why has this changed?

kleerwater: I changed it so that buildifier will sort keyword arguments on calls that are accessed via dot expressions that start with a literal expression. str.rewrite is such a call

pmbethe09: Unfortunately that breaks a bunch of internal tests, so I don't think it can be included at this time.

pmbethe09:
This passes internal tests now,

pmbethe09:
Jenkins, test this please.

PR #95, opened 2017-05-08
Update go rule version to 0.4.4

kchodorow:


damienmg:
(Ideally rules_go would tag a new release but since the previous reference was not to a tagged version, sounds good).

kchodorow:
They did. This uses http_archive, which doesn't take a tag.

Issue #99, opened 2017-05-16
Unable to set srcs to glob with buildozer.


har79:
Buildozer assumes the value for the srcs attribute is always a list, preventing the use of globs. For example:
`buildozer "set srcs glob(["*"])" $target`
produces the attribute
`srcs = ["glob([\"*\"])"]`

PR #100, opened 2017-05-16
Handle glob arguments in buildozer.

har79:
If a list attribute has a single glob argument, it will now be treated as a literal argument instead.

For example,
`> buildozer 'set srcs glob(["*.go"])' $target`
produces `srcs = glob(["*.go"])` instead of `srcs = ["glob([\"*.go\"])"]`.

To fix https://github.com/bazelbuild/buildtools/issues/99.


laurentlb:
Ideally, we'd find a more general solution.
But we can do this for now.

Issue #70, opened 2017-03-22
certain buildifier runs are not idempotent


pmbethe09:
after running buildifier, you would expect the 2nd time around to be unchanged/identical, but apparently not for some cases:

example
$ cat > /tmp/dir/BUILD
py_test(
    tags = ["manual", "local"],  # comment                                                   
)

$ cat /tmp/dir/BUILD | buildifier
py_test(
    tags = [
        "manual",
        "local",
    ],  # comment
)

$ cat /tmp/dir/BUILD | buildifier | buildifier
py_test(
    tags = [
        "manual",
        "local",  # comment
    ],
)

vladmos:
The issue has been fixed in #102

PR #102, opened 2017-06-02
Fix disappearing comments

### LINKED ISSUE #68: Buildifier removes some comments

vladmos:
Fixes #68. #70 seems to also be fixed by this.


laurentlb:
Looks good!

PR #115, opened 2017-06-21
Return non zero code, if check mode failed

davido:
Fixes: #110.

The point of check mode is to be able to verify the formatting of BUILD
files from the CI:

```
  $ buildifier -mode=check $(find . -name BUILD)
```

Choose the next free return value 4, and return it in case, reformat is
needed.


bazel-io:

pmbethe09:
Jenkins, test this please.

PR #129, opened 2017-08-24
Remove references to `g4 edit`

drigz:
gofmt has also removed some whitespace - let me know if that's not desired and I can revert it.


pmbethe09:
nice catch.

pmbethe09:
LGTM

PR #130, opened 2017-08-25
ParseLabel(): handle labels without colons

drigz:
Before: "label" was interpreted as "label:label", but it actually refers
to ":label".

https://docs.bazel.build/versions/master/build-ref.html#labels


laurentlb:
LGTM, thanks!

PR #131, opened 2017-08-25
Parse WORKSPACE for BUILD files of external repos

drigz:
This can then be used by buildozer to locate such BUILD files. It won't
work for repos defined with Skylark - for example, for [Tensorflow][1]
it could locate models.BUILD but not the BUILD files in third_party.

[1]: https://github.com/tensorflow/tensorflow/blob/master/WORKSPACE

### DIFF
--- a/wspace/workspace.go
+++ b/wspace/workspace.go
@@ -18,8 +18,11 @@ distributed under the License is distributed on an "AS IS" BASIS,
 package wspace
 
 import (
+	"io/ioutil"
 	"os"
 	"path/filepath"
+
+	"github.com/bazelbuild/buildtools/build"
 )
 
 const workspaceFile = "WORKSPACE"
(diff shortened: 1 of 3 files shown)


pmbethe09:
Why not just call bazel query instead?

drigz:
@pmbethe09 Is that possible? I can't see how I would query repository rules, and [the docs](https://docs.bazel.build/versions/master/query.html) don't say much about external repos.

pmbethe09:
Ah good point -- not so much for repository rules.

wspace/workspace.go
```diff
@@ -62,3 +65,33 @@ func Find(dir string) (string, error) {
 	}
 	return Find(filepath.Dir(dir))
 }
+
+// FindRepoBuildFiles parses the WORKSPACE to find BUILD files for non-Bazel
+// external repositories, specifically those defined by one of these rules:
+//   new_local_repository(), new_git_repository(), new_http_archive()
+func FindRepoBuildFiles(root string) (map[string]string, error) {
```
pmbethe09: seems fine, but where will this be used?

drigz: Check out my next change here: https://github.com/drigz/buildtools/commit/635278c78f52adca6bda6d34f5746cc2b4c44c88

I've been using it for running buildozer against BUILD files for non-bazel repositories.

PR #135, opened 2017-09-05
Load BUILD files for external (new_*) repos

drigz:
This adds an extra output variable to `ParseLabel()` for the repository,
to support labels of the form `@repo//pkg:rule`. Users of `ParseLabel()`
that don't want to support external repos need to change to:

`_, pkg, rule = ParseLabel(label)`


laurentlb:
LGTM

PR #136, opened 2017-09-06
Lint and warning fixes

drigz:
I haven't been able to test this as I get the following error with bazel 0.5.3:

```
ERROR: error loading package 'bazel-buildtools/external/bazel_tools/tools/jdk': Extension file not found. Unable to load package for '//tools/jdk:alias_rules.bzl': BUILD file not found on package path
```

Any idea why that is?

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -10,10 +10,17 @@ http_archive(
     ],
 )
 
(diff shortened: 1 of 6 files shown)


pmbethe09:
try a clean and rebuild?  I haven't seen that one.

Issue #139, opened 2017-09-12
bazel build //buildifier fails


cgrushko:
```
ERROR: error loading package '': Encountered error while reading extension file 'go/def.bzl': no such package '@io_bazel_rules_go//go': Error downloading [http://mirror.bazel.build/github.com/bazelbuild/rules_go/archive/2e319588571f20fdaaf83058b690abd32f596e89.tar.gz, https://github.com/bazelbuild/rules_go/archive/2e319588571f20fdaaf83058b690abd32f596e89.tar.gz] to [redacted]/external/io_bazel_rules_go/2e319588571f20fdaaf83058b690abd32f596e89.tar.gz: Checksum was 84dc11091f5209bf24f1ee710ff0c645f86016bfdbbb11fb831580e95f1975c6 but wanted 1e8e662ab93eca94beb6c690b8fd41347835e8ce0f3c4f71708af4b6673dd171.
```

I saw something similar internally, with the conclusion:

> TLDR: Don't depend on archives, depend on releases (things under "release" ending with "-dist").

From: https://github.com/bazelbuild/bazel/issues/3722 .

cgrushko:
@pmbethe09 @laurentlb

cgrushko:
Temporarily worked around this by removing `sha256 =` attributes from `http_archive`, including one inside `external/` :|

PR #142, opened 2017-09-13
update the WORKSPACE to point to rules_go 0.5.5 

pmbethe09:
also removed the sha26 which is breaking from GitHub.

Fixes https://github.com/bazelbuild/buildtools/issues/139

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,12 +1,12 @@
 workspace(name = "com_github_bazelbuild_buildtools")
 
+# 0.5.5
 http_archive(
     name = "io_bazel_rules_go",
-    sha256 = "1e8e662ab93eca94beb6c690b8fd41347835e8ce0f3c4f71708af4b6673dd171",
-    strip_prefix = "rules_go-2e319588571f20fdaaf83058b690abd32f596e89",
+    strip_prefix = "rules_go-e1c4b58c05e4a6ab67392daf28f3d57e4902f581",
     urls = [
-        "http://mirror.bazel.build/github.com/bazelbuild/rules_go/archive/2e319588571f20fdaaf83058b690abd32f596e89.tar.gz",
-        "https://github.com/bazelbuild/rules_go/archive/2e319588571f20fdaaf83058b690abd32f596e89.tar.gz",
+        "http://mirror.bazel.build/github.com/bazelbuild/rules_go/archive/e1c4b58c05e4a6ab67392daf28f3d57e4902f581.tar.gz",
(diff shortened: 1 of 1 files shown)


pmbethe09:
@cgrushko

WORKSPACE
```diff
@@ -1,12 +1,12 @@
 workspace(name = "com_github_bazelbuild_buildtools")
 
+# 0.5.5
 http_archive(
     name = "io_bazel_rules_go",
-    sha256 = "1e8e662ab93eca94beb6c690b8fd41347835e8ce0f3c4f71708af4b6673dd171",
```
cgrushko: this is required for security, I wouldn't remove it?
The problem here is that http://mirror.bazel.build/github.com/bazelbuild/rules_go/archive/2e319588571f20fdaaf83058b690abd32f596e89.tar.gz (or any URL of that form) isn't immutable and is generated on the fly, and has a different has each time.

See bazelbuild/bazel#3722

WORKSPACE
```diff
@@ -1,12 +1,12 @@
...
-    strip_prefix = "rules_go-2e319588571f20fdaaf83058b690abd32f596e89",
+    strip_prefix = "rules_go-e1c4b58c05e4a6ab67392daf28f3d57e4902f581",
     urls = [
-        "http://mirror.bazel.build/github.com/bazelbuild/rules_go/archive/2e319588571f20fdaaf83058b690abd32f596e89.tar.gz",
-        "https://github.com/bazelbuild/rules_go/archive/2e319588571f20fdaaf83058b690abd32f596e89.tar.gz",
+        "http://mirror.bazel.build/github.com/bazelbuild/rules_go/archive/e1c4b58c05e4a6ab67392daf28f3d57e4902f581.tar.gz",
```
cgrushko: why upgrade the version, actually?

pmbethe09: same reason -- issues with rules_go and sha256

https://github.com/bazelbuild/rules_go/commit/551e1f46cbeb470fe9d085e873c3bb83f075e18c

WORKSPACE
pmbethe09: Right, so how would sha256 work in concert with the mirror URL?

cgrushko: It doesn't - according to the issue I linked to, we should use Bazel releases (which are created by the Bazel team, and are immutable).

WORKSPACE
cgrushko: I see. I don't know what's the policy on hashes - they sound like a good idea, but /shrug.

pmbethe09:
Going to commit this so we can unblock other things -- circle back on the sha256 issue

PR #45, opened 2017-01-30
Introduce a -tables option.

bolinfest:
The `-tables` option makes it possible to specify a JSON file whose
properties define arguments that will be passed to the
`tables.OverrideTables()` function such that other projects can define
their own constraints on argument formatting. This is valuable to
Buck or Skylark users that define rules that need to be formatted in a
way that is different from buildifier's defaults.

Test Plan:
* Added a test: `tables/jsonParser_test.go`.
* Verified that `bazel test //...` succeeds.
* Created a JSON config for Buck and verified the output of the
following was what I expected when run from the Buck repo:

```
export BUILDIFIER_DIFF="diff -u"
find . -name BUCK | xargs -I {} ../buildifier/bazel-bin/buildifier/buildifier -buildifier_disable=label -tables=buildifier.json -d {}
```

For this test, the contents of `buildifier.json` were:

```
{
  "IsLabelArg": {
  },
  "LabelBlacklist": {
  },
  "IsSortableListArg": {
    "deps": true,
    "exported_deps": true,
    "provided_deps": true,
    "srcs": true,
    "visibility": true
  },
  "SortableBlacklist": {
    "genrule.srcs": true
  },
  "SortableWhitelist": {
  }
}
```

PR #146, opened 2017-10-04
Allow set literals in BUILD files

illicitonion:
This allows {}, {"foo"}, {"foo",}, and such to appear as attribute
values. These are canonically formatted identically to lists.

### DIFF
--- a/build/lex.go
+++ b/build/lex.go
@@ -647,6 +647,11 @@ func (in *input) order(v Expr) {
 			in.order(c)
 		}
 		in.order(&v.End)
+	case *SetExpr:
+		for _, x := range v.List {
+			in.order(x)
+		}
+		in.order(&v.End)
 	case *ForClauseWithIfClausesOpt:
 		in.order(v.For)
 		for _, c := range v.Ifs {
--- a/build/parse.y
+++ b/build/parse.y
@@ -104,7 +104,7 @@ package build
 %type	<expr>		stmt
 %type	<expr>		keyvalue
 %type	<exprs>		keyvalues
-%type	<exprs>		keyvalues_opt
+%type	<exprs>		keyvalues_no_comma
 %type	<string>	string
 %type	<strings>	strings
 
@@ -281,9 +281,19 @@ expr:
 			ForceMultiLine: $1.Line != exprStart.Line,
 		}
 	}
-|	'{' keyvalues_opt '}'
+|	'{' keyvalues '}'
 	{
 		$$ = &DictExpr{
+			Start: $1,
+			List: $2,
+			Comma: $<comma>2,
+			End: End{Pos: $3},
+			ForceMultiLine: forceMultiLine($1, $2, $3),
+		}
+	}
+|	'{' exprs_opt '}'
+	{
+		$$ = &SetExpr{
 			Start: $1,
 			List: $2,
 			Comma: $<comma>2,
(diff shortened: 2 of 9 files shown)


bazel-io:

laurentlb:
Hi,

Can you give more context for this change? This syntax is not valid in Skylark and BUILD files. Why do you want to add it here?

illicitonion:
Yeah, that was a pretty sparse description, sorry about that!

I'm looking at buildifier-formatting Pants BUILD files at Twitter to try to get in a place where we can be using and contributing to the same tools as Bazel around BUILD files. My initial inspection has shown seven major differences between Bazel and Pants BUILD files. Only three of them impact our ability to use Buildifier, and only one or two of them would need changes in Buildifier:

1. Pants allows import statements, whereas buildifier breaks them. The answer here is pretty clearly to forbid import statements.
2. Pants allows set-literal arg values, as in:
```java_library(
  name = "foo",
  srcs = ["Foo.java"],
  labels = {"testonly"},
)
```
this seems a pretty harmless thing to be adding to buildifier, even if it's unsupported by Skylark. That's this PR.
3. Pants has some extra attribute names to add to the takes-labels and sortable tables. Notably, Pants spells `deps` as `dependencies`. This is easily handled by the `-tables` option which was added in https://github.com/bazelbuild/buildtools/pull/45.
4. In Pants, // prefixes are optional on targets, and there has been talk about making it required to omit them in BUILD files. This *works*, but means that intra-package labels are sorted before inter-label packages (which is a little weird). It also breaks label shortening ("foo/bar:bar" -> "foo/bar"), which Pants does support. This is an area still to be explored - maybe we can loosen up Buildifier, maybe we can make some changes to the Pants conventions.
5. Pants does not require the name attribute if the name of the target is equal to the name of the directory in which it lives. This isn't a concern for buildifier, but will make buildozer unhappy.
6. Pants leans a lot more heavily on function-call-like attributes, a-la Bazel's FilesetEntry:
```
jvm_app(
  name = "foo",
  bundles = [
    bundle(
      rel_path = "bar",
      fileset = [
        "config.yml",
        "image.png",
      ],
    ),
  ],
)
```
This isn't a problem for buildifier, but makes buildozer a little less powerful in a Pants world. It's possible we'll want to add some better support for function-call-like attributes in the future.
7. Pants spells `glob` like `globs` and `rglobs`. Not a problem for buildifier or buildozer, but a notable difference (and really just a special case of 6).

build/lex.go
```diff
@@ -222,12 +222,12 @@ func (in *input) Lex(val *yySymType) int {
...
-			prefix := bytes.TrimSpace(in.complete[i+1:in.pos.Byte])
+			prefix := bytes.TrimSpace(in.complete[i+1 : in.pos.Byte])
 			isSuffix := true
 			if len(prefix) == 0 ||
-					prefix[len(prefix) - 1] == '[' ||
-					prefix[len(prefix) - 1] == '(' ||
-					prefix[len(prefix) - 1] == '{' {
+				prefix[len(prefix)-1] == '[' ||
+				prefix[len(prefix)-1] == '(' ||
+				prefix[len(prefix)-1] == '{' {
```
laurentlb: Were these changes made by gofmt?

laurentlb:
Thanks for the background. It would be great to collaborate and have some convergence between Bazel and Pants BUILD files. We can discuss it outside this PR.

Adding support for sets in Buildifier looks reasonable for me.

build/lex.go
illicitonion: They were. I've reverted all gofmt changes, and can send a second PR with them if you'd like.

illicitonion:
Cool, thanks!

Added some tests. All tests pass for me.

laurentlb:
LGTM, thanks!

PR #147, opened 2017-10-08
Accept either WORKSPACE or .buckconfig as indicating repo root.

### LINKED ISSUE #93: buildozer: WORKSPACE search is surprising for Buck users
Buck doesn't have an equivalent to Bazel's `WORKSPACE` files. Normally, this isn't a problem -- the search fails and buildozer assumes `$PWD` is the root directory. Because Mac OS uses a case-insensitive filesystem, this causes surprising behavior for buildozer+Buck users who happen to have a file or directory called `workspace` in their home directory (e.g., because they've used Eclipse).

Passing `-root_dir .` (and running from the Buck root) is a workaround, but it's not an obvious one; the same seems true for adding an otherwise-unnecessary `WORKSPACE` file to the Buck root. Perhaps there is something we can do in Buildozer to make using it easier for Buck users?

swolchok:
Fixes #93.

### DIFF
--- a/wspace/workspace.go
+++ b/wspace/workspace.go
@@ -27,6 +27,8 @@ import (
 
 const workspaceFile = "WORKSPACE"
 
+var repoRootFiles = [...]string{workspaceFile, ".buckconfig"}
+
 // findContextPath finds the context path inside of a WORKSPACE-rooted source tree.
 func findContextPath(rootDir string) (string, error) {
 	if rootDir == "" {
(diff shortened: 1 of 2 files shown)


bazel-io:

swolchok:
I think the CLA should be fixed now? not sure.

swolchok:
Hmm, maybe there's some latency in the CLA system? I commented pretty quick after I was told the problem had been resolved on our end.

laurentlb:
LGTM, thanks!

PR #149, opened 2017-10-12
Emphasize that Buildozer doesn't require a separate Buildifier binary

cgrushko:


cgrushko:
@laurentlb 

This came up in https://github.com/bazelbuild/BUILD_file_generator/pull/19, where we debate whether we need to install both Buildozer and Buildifier, or whether Buildozer alone is enough.

laurentlb:
LGTM

Issue #148, opened 2017-10-09
[propose/discuss] allow implicit target name


wisechengyi:
Hi,

Coming from pantsbuild/pants, we are very interested integrating buildozer into our tool chain (https://github.com/pantsbuild/pants/pull/4921). 

However, one issue we encountered is that Pants allows implicit target name (same as dirname) to encourage [1:1:1 rule](https://www.pantsbuild.org/build_files.html#target-granularity) by making  targets easier. For example:
```
$ cat foo/BUILD

java_library(
  sources=globs('*.java')
  dependencies=[...]
)
```
This means this target is //foo:foo

Currently buildozer is unable to parse the target without a name, and therefore I am proposing to have buildozer to recognize such behavior.

One may also ask how to tell target with implicit name from a function call. E.g.
```
x = a_function(my_arg=...)

java_library(
  sources=...
)
```

The heuristics we plan to take is to make only top-level symbols with no assignment eligible, so the effect would be as follows:
```
x = a_function(my_arg=...) # this is ineligible because it is an assignment

# This would be eligible
java_library(
  sources=...
  provides=(...)  # this is ineligible because it is not top level
)
```

illicitonion:
@laurentlb @pmbethe09 Would y'all accept a patch which added the described behaviour, to make buildozer more friendly to use from Pants? That behaviour being:
```
When trying to find target //foo/bar/baz:baz
if foo/bar/baz/BUILD does not contain any target with attribute name="baz":
  if foo/bar/baz/BUILD contains exactly one target with no name attribute:
    treat that target as //foo/bar/baz:baz
```
?

(This is kind of an analog to the whole //foo/bar/baz:baz == //foo/bar/baz behaviour exhibited by both bazel and pants, except at the point where targets are defined, rather than where they're referenced)

laurentlb:
Would that be temporary or permanent? Do you have plans to enforce the name attribute in Pants?

The proposal above seems reasonable because there are lots of conditions. Implicit targets sound a bit scary, especially when some targets might be generated by a macro. But if it's restricted to the case where there's only one anonymous target, that should be fine.

stuhood:
> Would that be temporary or permanent? Do you have plans to enforce the name attribute in Pants?

[...]

PR #154, opened 2017-11-09
Add Implicit Name Recognition

15Dkatz:
It would be nice to have implicit name recognition to support the smooth integration of buildozer into Pants at Twitter. This allows for target granularity by making targets easier to create. With a [1:1:1 rule](https://www.pantsbuild.org/build_files.html#target-granularity), for one target per directory, the target can implicitly use the directory name.

This code change supports this behavior:

```
When trying to find target //foo/bar/baz:baz
if foo/bar/baz/BUILD does not contain any target with attribute name="baz":
  if foo/bar/baz/BUILD contains exactly one target with no name attribute:
    treat that target as //foo/bar/baz:baz
```

Original proposal: https://github.com/bazelbuild/buildtools/issues/148

**To verify**
Setup a `tmp` directory with a BUILD file that has only one nameless target. Try and confirm:
`./buildozer 'add dependencies a/b' tmp:tmp`

@illicitonion @wisechengyi

### DIFF
--- a/edit/edit.go
+++ b/edit/edit.go
@@ -262,10 +262,10 @@ func FindRuleByName(f *build.File, name string) *build.Rule {
 
 // UseImplicitName returns the rule in the file if it meets these conditions:
 // - It is the only unnamed rule in the file.
-// - The ending directory in the file path and passed rule name match.
+// - The file path's ending directory name and the passed rule name match.
 func UseImplicitName(f *build.File, rule string) *build.Rule {
 	// We disallow empty names
-  if f.Path == "BUILD" {
+	if f.Path == "BUILD" {
 		return nil
 	}
 	ruleCount := 0
--- a/edit/edit.go
+++ b/edit/edit.go
@@ -261,10 +261,11 @@ func FindRuleByName(f *build.File, name string) *build.Rule {
 }
 
 // UseImplicitName returns the rule in the file if it meets these conditions:
-// - It is the only rule in the file.
-// - The passed package and rule names match.
+// - It is the only unnamed rule in the file.
+// - The ending directory in the file path and passed rule name match.
 func UseImplicitName(f *build.File, rule string) *build.Rule {
-	if f.Path == "BUILD" {
+	// We disallow empty names
+  if f.Path == "BUILD" {
 		return nil
 	}
 	ruleCount := 0
--- a/edit/edit.go
+++ b/edit/edit.go
@@ -18,6 +18,7 @@ package edit
 import (
 	"fmt"
 	"path"
+	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
@@ -248,7 +249,7 @@ func InsertAtEnd(stmt []build.Expr, expr build.Expr) []build.Expr {
 
 // FindRuleByName returns the rule in the file that has the given name.
 // If the name is "__pkg__", it returns the global package declaration.
-func FindRuleByName(f *build.File, pkg string, name string) *build.Rule {
+func FindRuleByName(f *build.File, name string) *build.Rule {
 	if name == "__pkg__" {
 		return PackageDeclaration(f)
 	}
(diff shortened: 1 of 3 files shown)


bazel-io:

15Dkatz:
I signed it!

edit/edit.go
```diff
@@ -248,15 +248,40 @@ func InsertAtEnd(stmt []build.Expr, expr build.Expr) []build.Expr {
...
+	var r *build.Rule
 
+	for _, stmt := range f.Stmt {
+		call, ok := stmt.(*build.CallExpr)
+		if !ok {
+			continue
+		}
+		r = &build.Rule{Call: call}
+		if r.Kind() != "" {
+			ruleCount++
```
illicitonion: Nothing is checking whether the rule has a name here. That means that this code will only work if the BUILD file has exactly one target, rather than when it has exactly one unnamed target. It would be nice to add the name check here, so we don't have that restriction :)

edit/edit.go
```diff
@@ -248,15 +248,40 @@ func InsertAtEnd(stmt []build.Expr, expr build.Expr) []build.Expr {
...
 	if i != -1 {
 		return &build.Rule{Call: f.Stmt[i].(*build.CallExpr)}
 	}
+	return UseImplicitName(f, pkg, name)
+}
+
+// UseImplicitName returns the rule in the file if it meets these conditions:
+// - It is the only rule in the file.
+// - The passed package and rule names match.
+func UseImplicitName(f *build.File, pkg string, rule string) *build.Rule {
```
illicitonion: You should be able to use filepath.Base(filepath.Dir(f.Path)) to get the last component of the package name, rather than needing to pass it in explicitly.

This also means you don't need to do the string slicing below :)

(Should probably also error if f.Path == "BUILD", because I don't believe an empty implicit name is allowed)

15Dkatz:
@illicitonion Added a test and the discussed behavior :)
* using filePath
* returning an error on `f.Path == "BUILD"`
* additional name check to ensure we use the single unnamed target in the BUILD file

edit/edit_test.go
```diff
@@ -172,3 +172,43 @@ func TestAddValueToListAttribute(t *testing.T) {
...
+	}
+
+	for _, tst := range tests {
+		path := "foo/BUILD"
+		if tst.wantRootErr {
+			path = "BUILD"
+		}
+		bld, err := build.Parse(path, []byte(tst.input))
+		if err != nil {
+			t.Error(err)
```
illicitonion: Because we're in a loop, it's useful to prefix errors with a descriptive string to say which test case this is failing for - the test runtime will only output the line number, which doesn't tell you which thing failed. Accordingly, it's fairly common to have a `description` string in your tests struct, and here do `t.Error(tst.description, err)`

You could also use this below where you're currently using tst.input for the same purpose (but two of your test cases have the same input :))

edit/edit.go
```diff
@@ -256,7 +257,37 @@ func FindRuleByName(f *build.File, name string) *build.Rule {
 	if i != -1 {
 		return &build.Rule{Call: f.Stmt[i].(*build.CallExpr)}
 	}
+	return UseImplicitName(f, name)
+}
+
+// UseImplicitName returns the rule in the file if it meets these conditions:
```
illicitonion: This comment is now a lie :)

edit/edit.go
```diff
@@ -256,7 +257,37 @@ func FindRuleByName(f *build.File, name string) *build.Rule {
...
+// - It is the only rule in the file.
+// - The passed package and rule names match.
+func UseImplicitName(f *build.File, rule string) *build.Rule {
+	if f.Path == "BUILD" {
```
illicitonion: Maybe add a comment explaining this early-return. ("We don't allow empty names.")

15Dkatz:
@illicitonion Addressed the documentation-y comments :D

illicitonion:
Great, thanks! Over to @laurentlb :)

15Dkatz:
@laurentlb Updated the `UseImplicitName` docs to mention it's supporting the Pants convention.

laurentlb:
Looks good!

drigz:
@15Dkatz This change breaks an internal user of `FindRuleByName` that uses it to check for a rule of a given name in a file. This is used before creating a new rule to avoid name collisions.

Going by the documented behaviour of `FindRuleByName`, this seems like a valid use:

```
// FindRuleByName returns the rule in the file that has the given name.
// If the name is "__pkg__", it returns the global package declaration.
```

Would you be open to restoring the old behaviour of `FindRuleByName` and adding a new function `FindRule` that has the new behaviour? Then buildozer.go could call `FindRule` and users of `FindRuleByName` won't get broken.

wisechengyi:
@illicitonion Do you see a way both could work?

wisechengyi:
@drigz could you give an example what is broken? afaict in this PR, `__pkg__` was the first thing to check for. 
https://github.com/bazelbuild/buildtools/pull/154/files#diff-d040f6ff3cbfb2eb5a54dde6fb2cebedR253

drigz:
@wisechengyi The code is something like this (but the internal code has better strings that xyz):

```
func addConfigRule(f *build.File) {
  newRuleName = "config"
  if edit.FindRuleByName(f, newRuleName) != nil {
    // A rule called "config" already exists, so use "config_xyz" as this is less likely to conflict.
    newRuleName = newRuleName + "_xyz"
  }
  ...
}
```

The pattern of using `FindRuleByName` to check if a rule exists with a given name is also used elsewhere.

The new behaviour will return the implicitly-named rule if the package is called "config" and there's an unnamed rule in the file. This is most likely to spuriously occur because of #165, so maybe fixing that would be enough. @laurentlb @illicitonion Can you think of other cases where Buildozer might mistakenly pick up an implicit rule?

PR #156, opened 2017-11-27
Allow tables.json to configure label default behaviour to strip leading //

illicitonion:
Pants has this behaviour. This is completely configurable through tables.json, so shouldn't affect anyone who doesn't want this behaviour.

Also, treat //foo:bar to be equivalent to :bar in foo/BUILD

The three commits should be usefully independently reviewable.

### DIFF
--- a/build/print_test.go
+++ b/build/print_test.go
@@ -23,7 +23,10 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"strings"
 	"testing"
+
+	"github.com/bazelbuild/buildtools/tables"
 )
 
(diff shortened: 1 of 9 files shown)


bazel-io:

illicitonion:
Ok, removed that commit, can do that as a separate PR, which maybe we can turn on after a global cleanup at Google?

I'll rebase now, as this conflicts a little with https://github.com/bazelbuild/buildtools/pull/63, assuming you're happy to accept such a patch?

illicitonion:
Pushed a rebase, PTAL @laurentlb

PR #157, opened 2017-11-28
Allow buildozer to edit arbitrary files

drigz:
This is useful for WORKSPACE files in particular, eg:

`buildozer 'print name' WORKSPACE:%http_archive`

Make sure that ParseLabel doesn't interfere with absolute paths to
support this overloading.


laurentlb:
LGTM

Issue #137, opened 2017-09-07
Building with bazel on Windows 10 fails


Xjs:
I'm [currently trying](https://github.com/bazelbuild/rules_go/issues/736) to get the rules_go and gazelle tools running under Windows 10. We have reached limited success so far, but now I have come to a point where I can't continue because buildtools won't build without at least a stub of the `isatty()` function in _differ._

The problem lies in [the `BUILD.bazel` file of differ](https://github.com/bazelbuild/buildtools/blob/master/differ/BUILD.bazel#L8). I don't know exactly how the constraints support you seem to be awaiting should look, but as far as I understand, the go compiler reads the build constraints (`// +build` comments) itself -- so there should be no harm in including both the isatty_windows.go and the isatty_other.go file in `BUILD.bazel`, should there?

I'm not sure if I'm missing something, but if not, I would appreciate very much if the `isatty_windows.go` file could be included in `differ/BUILD.bazel`.

PR #161, opened 2017-11-30
Fix differ and make buildifier build on Windows.

sgreenstein:
- Uses `FC` to diff on Windows by default.
- Adds `config_setting` to use `isatty_windows.go` on Windows
- Removes `go_repository` rule that fails on Windows

Fixes: #137
Test: Built and ran `buildifier <file>` and `buildifier -d <file>` on both Windows and Linux


bazel-io:

laurentlb:
LGTM, thanks!

PR #162, opened 2017-12-01
Make --stdout print the BUILD file even when there is no change.

laurentlb:
Also remove old commands from README.md


laurentlb:
@vladmos Can you review this?

vladmos:
//lang:tables.gen.go_checkshtest fails on all platforms, ptal.

laurentlb:
Fixed the tests

PR #163, opened 2017-12-01
Allow tables.json to configure shortening //foo:bar to :bar within foo/BUILD

illicitonion:
Also shorten //foo:foo to :foo within foo/BUILD

### DIFF
--- a/build/print_test.go
+++ b/build/print_test.go
@@ -44,6 +44,7 @@ func TestPrintGolden(t *testing.T) {
 		if strings.Contains(out, ".stripslashes.") {
 			tables.StripLabelLeadingSlashes = true
 		}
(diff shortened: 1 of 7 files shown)


bazel-io:

illicitonion:
This was originally in https://github.com/bazelbuild/buildtools/pull/156 but @laurentlb requested it not be enabled by default because it will break many files at Google.

build/print_test.go
```diff
@@ -56,13 +60,20 @@ func TestPrintRewrite(t *testing.T) {
 	for _, in := range ins {
 		prefix := in[:len(in)-len(".in")]
 		out := prefix + ".golden"
+
+		if strings.Contains(out, "/050.") {
```
laurentlb: Please add a comment (I was initially confused by the "/050.")

laurentlb:
LGTM

PR #166, opened 2017-12-04
Don't consider load() as an implicit rule

### LINKED ISSUE #165: load() statements should not be considered unnamed rules
#154 introduced the following behaviour:

```
> cat test/BUILD
load("asdf", "fdsa")
> bazel-bin/buildozer/buildozer 'print rule' //test
load("asdf", "fdsa")
```

The `load()` statement is not a rule, yet the logic in `UseImplicitName` considers it a rule.

**Fix**

We could change [the unnamed-rule condition](https://github.com/bazelbuild/buildtools/pull/154/files#diff-d040f6ff3cbfb2eb5a54dde6fb2cebedR283) to check that `kind != "load"`.

Are there any other statements that could be confused for rules?

drigz:
Fixes #165.

Also amend the documentation of FindRuleByName to explain its behaviour
in the presence of implicitly-named rules.

### DIFF
--- a/edit/edit.go
+++ b/edit/edit.go
@@ -270,6 +270,8 @@ func InsertAtEnd(stmt []build.Expr, expr build.Expr) []build.Expr {
 
 // FindRuleByName returns the rule in the file that has the given name.
 // If the name is "__pkg__", it returns the global package declaration.
+// If the name is that of the package and the file contains a single unnamed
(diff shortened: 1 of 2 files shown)


vladmos:
Load statements are now parsed as different AST nodes (`LoadStmt` instead of `CallExpr`), so this pull request is probably obsolete.

drigz:
That sounds right; I'll close it.

PR #168, opened 2017-12-05
buildozer: add command line flag to override build file names

andrewjcg:
Lots of places in the code still assume BUILD is the build file name, as such
this is mainly useful to add additional build file names.

### DIFF
--- a/buildozer/main.go
+++ b/buildozer/main.go
@@ -38,6 +38,7 @@ var (
 
 	shortenLabelsFlag  = flag.Bool("shorten_labels", true, "convert added labels to short form, e.g. //foo:bar => :bar")
 	deleteWithComments = flag.Bool("delete_with_comments", true, "If a list attribute should be deleted even if there is a comment attached to it")
+	buildFiles         = stringList("build-files", "comma-separated list of build file names")
 )
 
 func stringList(name, help string) func() []string {
@@ -72,6 +73,7 @@ func main() {
 		Quiet:             *quiet,
 		EditVariables:     *editVariables,
 		IsPrintingProto:   *isPrintingProto,
+		BuildFileNames:    buildFiles(),
 	}
 	os.Exit(edit.Buildozer(flag.Args()))
 }
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -51,10 +51,19 @@ type Options struct {
 	Quiet             bool     // suppress informational messages.
 	EditVariables     bool     // for attributes that simply assign a variable (e.g. hdrs = LIB_HDRS), edit the build variable instead of appending to the attribute.
 	IsPrintingProto   bool     // output serialized devtools.buildozer.Output protos instead of human-readable strings
+	BuildFileNames    []string // list of build file names
 }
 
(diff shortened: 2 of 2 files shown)


bazel-io:

laurentlb:
Hi, why do you need this?
I thought checking for Bazel/Buck/Pants files would be enough. Is there another scenario?

Thanks

andrewjcg:
Ah sorry, missed your comment.

We have an older repo which uses a different name for build files (`TARGETS`) and thought this might be more desirable than adding another name to the list.

We have some wrappers (supporting a narrower UI) that we use to make this work currently by temporarily copying the original to use the `BUILD` basename, so this change isn't that big of a deal.  It just means we can't run `buildozer` directly.

laurentlb:
I would prefer to avoid adding a new flag for that. Is it an option to rename your files in the old repo?

mprobst:
I have this issue too – we have packages that contain both a `BUILD` and a `BUILD.bazel` file, one for blaze and one for bazel. It'd be nice if we could select the right file when running buildozer.

benjaminp:
buildifier on master now parses `BUILD.bazel` files as `BUILD` files. In #400, I extend the heuristic to match filenames commonly used for external workspace `BUILD` files.

motiz88:
This is a few years old now, but as a data point, I've run into a case within Meta's codebase where the hardcoded list is problematic.

In a part of the repo where we use Buck, we have a particular directory with the following two files:
* `BUCK`
* `build` (some random shell script not related to the Buck build)

In a Mac environment (where the filesystem is case insensitive), buildozer errors when we ask it to edit the `BUCK` file in this directory, because it thinks `build` is a `BUILD` file (and `BUILD` always takes precedence over `BUCK` per [`BuildFileNames`](https://github.com/bazelbuild/buildtools/blob/0e2f864aac3a1b18d758b5690f273b3094031603/edit/buildozer.go#L892)). It seems like we'd have to either patch our buildozer binary to reorder `BuildFileNames`, or reorganise our repo to avoid the issue. A flag would allow us to specify the ordering dynamically when invoking buildozer.

hick209:
Any updates on this?

oreflow:
I believe this PR unfortunately lacks quite a few updates before it would be in a mergeable state.

- The changes lack tests
- The changes are based on an 8 year old revision (which might be cause for following items)
- The changes only modifies the buildozer.go file names, which is incomplete as file name is used in more locations, for instance
  - file names are referenced in https://github.com/bazelbuild/buildtools/blob/eb7356da689578567b6ba8534b900c20781a835d/buildifier/utils/utils.go#L41
  - lex.go uses the file name to determine the type of the file (since for instance bzl and BUILD files are handled differently)
https://github.com/bazelbuild/buildtools/blob/eb7356da689578567b6ba8534b900c20781a835d/build/lex.go#L125

As we are ramping up support for this repository, I think this change as a whole is quite reasonable, so if someone has time to revise this to be done on the latest revision I'm happy to review the PR. But as this is so outdated I would suggest creating a new PR rather than trying to update this one.

Created #1435 for tracking of the topic itself, but closing off this one

PR #169, opened 2017-12-07
Fix paths for build files inside packages

drigz:
Before, this would break with:
  File not found: /path/to/workspace/third_party:repo.BUILD


laurentlb:
LGTM

PR #173, opened 2017-12-13
Don't strip // prefix for targets in the root BUILD file

illicitonion:
Unless the current BUILD file is in the root directory.


bazel-io:

laurentlb:
LGTM

PR #176, opened 2017-12-18
Disable the UseImplicitName feature.

laurentlb:
This modifies the behavior of Buildozer and breaks existing users. Once
the depot is in a more stable state, we can try to reintroduce this
feature (maybe with a special syntax?).


laurentlb:
@15Dkatz fyi

Sorry, there have been some issues from our side. In particular, we should open-source the internal tests we have, and integrate Github changes more often.
This PR was causing problems, so I've temporarily disabled it.

laurentlb:
I think the issue was https://github.com/bazelbuild/buildtools/issues/165

PR #178, opened 2017-12-21
Add print_if_changed mode to buildifier

illicitonion:
This prints the newly formatted file if there are changes to apply, or
prints nothing if there are no changes to apply.

It only takes one file.

This is useful for Phabricator's linter infrastructure, because they
require a replacement text, rather than a patch.

Without this change, Phabricator users need to read BUILD files
manually, pipe them into buildifier's stdin, and manually diff the two
to check whether changes need applying. The new mode makes this process
a lot cleaner.


bazel-io:

illicitonion:
@bolinfest IIRC you put some hacks into Facebook's linter to work around this too

illicitonion:
Ping @laurentlb?

bazel-io:

laurentlb:
LGTM
Sorry for the delay!

PR #181, opened 2018-01-04
Allow overriding IsList via -tables

adamjernst:
The default implementation of `edit.types.IsList` consults `lang.tables.TypeOf`. This means that the definition of whether a given label represents a list argument is essentially hardcoded into buildifier and buildozer.

Some consumers may wish to customize this value for certain labels. Introduce a feature that allows you to override the default value using the tables feature.


laurentlb:
LGTM, thanks!

bazel-io:

PR #184, opened 2018-01-15
Mention attribute removal in documentation.

meisterT:
It was already listed in the examples below but not in the list of commands.


laurentlb:
LGTM

PR #187, opened 2018-01-16
Support tuple unpacking in comprehensions

vladmos:


build/parse.y
```diff
@@ -531,6 +531,22 @@ ident:
 	{
 		$$ = &LiteralExpr{Start: $1, Token: $<tok>1}
 	}
+| '(' idents comma_opt ')'
+  {
+		if len($2) == 1 && $3.Line == 0 {
```
pmbethe09: what is going on with the indents on this line versus after the else?

laurentlb: please always use gofmt

vladmos: It's mixed tabs and spaces (and they're mixed somewhere else in this file). What should be used here?

vladmos: Replaced with tabs as almost elsewhere in this file.

build/parse.y
```diff
@@ -531,6 +531,22 @@ ident:
 	{
 		$$ = &LiteralExpr{Start: $1, Token: $<tok>1}
 	}
+| '(' idents comma_opt ')'
```
laurentlb: This looks wrong: an `ident` cannot contain `idents`. You probably want to change `idents` instead? Or change the grammar of `for` to accept more than just idents.

In theory, lvalues can be more complicated (you don't have to do everything in the PR). See:
https://docs.bazel.build/versions/master/skylark/spec.html#grammar-reference
https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/syntax/LValue.java#L70

vladmos: Maybe the name should be changed, but looks like the syntax defined here is correct. It can be a nested structure of any level, for example:

    for a, (b, (c, d)) in e

Brackets used to be defined in the definition of a for loop, but they don't really belong there because they can exist at any nested level. Also IMO a for loop variable is not an lvalue which can be something like f(x)[1], for loop variables should only contain idents in a nested structure.

I'm not sure the grammar reference is correct in this sense: LoopVariables are defined as a sequence of PrimaryExpr, each of them can be an Operand which can be a ListComp. Is that correct?

Perhaps I can do the following: revert the definition of `ident`, introduce a `loop_var` which can be either `ident` or `'(' loop_vars ')'` and rename `idents` to `loop_vars` (`idents` is currently only used for loop variables anyway). That will be more correct.

vladmos: Apparently syntax like`for a[0] in b` is also allowed (but is also disallowed with the current formatter). Then I can define a `loop_var` as `expr` or `loop_var ',' expr`. Then it will be simple for buildifier to parse, although it won't fail on invalid expressions like `for f(x) in y`. On the other hand, it also doesn't fail on `f(x) = y` despite `f(x)` is not a correct lvalue.

laurentlb: The parser doesn't have to detect every possible error in the code. It's perfectly fine to consider that `for f(x) in y` is syntactically valid.

Just look at other Skylark grammars/parsers to find what to do. I gave some links, here's on more:
https://github.com/google/skylark/blob/master/syntax/syntax.go#L268
(when possible/easy, try to align on this Skylark interpreter for AST nodes)

laurentlb:
LGTM

pmbethe09:
LGTM

PR #189, opened 2018-01-17
Experimental formatter for bzl files with def expressions

vladmos:


vladmos:
This pull request doesn't provide a complete support of bzl files, only def-blocks are currently supported, however it already supports nested structures, so adding for-, if- and if-else-blocks is just a matter of adding a couple of new nodes.

Currently known problems (to be fixed in the follow-up pull requests):

  * Tabs are not supported properly,
  * If a function definition contains a comment it will be moved to its last argument (that's reflected in the 051.formatbzl.golden testfile)

The behavior shouldn't be changed if the `--format_bzl` flag is not passed explicitly, so these limitations shouldn't actually break anything.

build/lex.go
```diff
@@ -571,13 +632,19 @@ func (in *input) skipPython(p []byte) []byte {
...
 			}
 
-			if !isBlankOrComment(p[i:]) {
-				if !hasPythonContinuation(p[i:]) && c != ' ' && c != '\t' {
-					// Yes, stop here.
-					break
+			if tables.FormatBzlFiles {
+				// In the bzl files mode we only care about the end of the statement, we've found it.
+				break
+			} else {
```
pmbethe09: don't need 'else' when the proceeding if has a break (or return or continue)

build/lex.go
```diff
@@ -571,13 +632,19 @@ func (in *input) skipPython(p []byte) []byte {
 				rest = p[i:]
 			}
 
-			if !isBlankOrComment(p[i:]) {
-				if !hasPythonContinuation(p[i:]) && c != ' ' && c != '\t' {
-					// Yes, stop here.
-					break
+			if tables.FormatBzlFiles {
+				// In the bzl files mode we only care about the end of the statement, we've found it.
+				break
```
pmbethe09: by my read, this breaks the for loop just to 'return rest'  -- (and also check for quote).
Instead just 'return rest' here, and after the declaration of quote, you can add a 
defer func() {
  if quote != 0 {
 		in.Error("EOF scanning Python quoted string")
 	}
}()

build/lex.go
```diff
@@ -571,13 +632,19 @@ func (in *input) skipPython(p []byte) []byte {
...
+			} else {
+				// In the legacy mode we need to find where the current block ends
+				if !isBlankOrComment(p[i:]) {
+					if !hasPythonContinuation(p[i:]) && c != ' ' && c != '\t' {
+						// Yes, stop here.
+						break
```
pmbethe09: ditto, 'return rest'

vladmos:
Addressed the comments.

This `skipStmt` function existed before (although for different reasons: to skip entire Python blocks, I need 90% of its former functionality to skip statements within which I don't need to track indentation). Once I remove the old behavior it'll be easier to refactor and simplify the code, also in other files.

build/lex.go
```diff
@@ -171,52 +180,66 @@ func (in *input) Lex(val *yySymType) int {
...
-				in.startToken(val)
-				in.readRune()
-				in.endToken(val)
-				return '\n'
-			}
 			if c == '\n' {
+				in.indent = 0
+				in.cleanLine = true
+				if in.endStmt == 0 {
+					// Not in a rule. Tell parser about top-level blank line.
```
laurentlb: Replace 'rule' with 'statement'?

build/parse.y
```diff
@@ -88,6 +89,9 @@ package build
 %token	<pos>	_OR      // keyword or
 %token	<pos>	_PYTHON  // uninterpreted Python block
 %token	<pos>	_STRING  // quoted string
+%token	<pos>	_DEF     // keyword def
+%token  <pos> _INDENT  // indentation
```
laurentlb: fix alignment

build/parse.y
```diff
@@ -108,6 +112,7 @@ package build
 %type	<exprs>		keyvalues_no_comma
 %type	<string>	string
 %type	<strings>	strings
+%type <block>   block
```
laurentlb: alignment

build/parse.y
```diff
@@ -442,6 +457,19 @@ expr:
                         Else: $5,
                 }
 	}
+| _DEF _IDENT '(' exprs_opt ')' ':' block
```
laurentlb: does it support single-line function definition?
def f(): return 1

if not, add a TODO

laurentlb:
LGTM

PR #193, opened 2018-01-24
Fix end statement tracking

vladmos:
End statement tracking wasn't accurate and didn't always work well, but on the other hand it was used only for implicit `;` insertion (which is not required for Skylark syntax, removing an explicit `;` from the end of a line never changes Skylark semantics).

However proper end statement tracking is crucial for bzl files parsing: files are not just sequences of top-level statements, it's important to see where each node ends.

Statements are now considered ended before `\n` of their last line, and the `\n` is now passed to the parser explicitly.

### DIFF
--- a/build/parse.y
+++ b/build/parse.y
@@ -198,7 +198,6 @@ stmts:
 		// attach the comments to the statement.
 		if cb, ok := $<lastRule>1.(*CommentBlock); ok {
 			$$ = append($1[:len($1)-1], $2...)
-			//$$[len($1)-1] = $2
 			$2[0].Comment().Before = cb.After
 			$<lastRule>$ = $2[len($2)-1]
 			break
--- a/build/parse.y
+++ b/build/parse.y
@@ -107,13 +107,17 @@ package build
 %type	<expr>		ident
 %type	<ifs>		if_clauses_opt
 %type	<exprs>		stmts
-%type	<expr>		stmt
+%type	<exprs>		stmt        // a simple_stmt or a for/if/def block
+%type	<expr>		block_stmt  // a single for/if/def statement
+%type	<exprs>		simple_stmt // One or many small_stmts on one line, e.g. 'a = f(x); return str(a)'
+%type	<expr>		small_stmt  // A single statement, e.g. 'a = f(x)'
+%type <exprs>		small_stmts_continuation  // A sequence of `';' small_stmt`
 %type	<expr>		keyvalue
 %type	<exprs>		keyvalues
 %type	<exprs>		keyvalues_no_comma
 %type	<string>	string
 %type	<strings>	strings
-%type	<block>		block
+%type	<block>		suite
 
 // Operator precedence.
 // Operators listed lower in the table bind tighter.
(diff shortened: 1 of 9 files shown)


vladmos:
Unfortunately, when I fixed the end statements tracking a lot of heuristics became broken, for instance, the detection whether a comment should be separated from the previous line by an empty line. Therefore I had to touch different parts of the lexer in the same pull request. Also I needed to update the parser to handle `\n` lexems in the end of each statement, at the same time I made the structure more strict by making it closer to what's defined at the [spec](https://docs.bazel.build/versions/master/skylark/spec.html#grammar-reference), e.g. splitting a definition of `stmt` into `stmt`, `simple_stmt`, and `small_stmt`.

pmbethe09:
seems like a hack, but I don't have a better suggestion.

LGTM

vladmos:
What exactly seems like a hack? The functionality was already there, it just didn't catch all the cases. E.g if a line ends with a comment, it didn't catch the '\n' lexem at the end of the line, now it does because now the syntax is more formal and depends on that. And there are no hacks like the following anymore:

    // skipStmt got confused.
    // No more virtual semicolons.
    in.endStmt = -1

`endStmt == 0` used to mean that the parser is currently not inside a statement, however it was a legal situation when a statement ended at the very end of the input (not anymore because now I add `\n` at the end of every input, but I still got rid of this ambiguity). I feel like I made the code more straightforward.

pmbethe09:
adding an extra '\n', then later calling "i--" requiring a comment.

vladmos:
Adding a `\n` is required to satisfy the `simple_stmt` definition if it's the last one in the file. Even in the official Skylark spec it's described in words, not formally. If it was possible in yacc to use something like look-ahead regexps I could add another definition for `simple_stmt` that would match `... _EOF` without actually consuming `_EOF`, but I don't know if that's even possible. Another workaround would be to return `'\n'` and then `_EOF` at the end of a file, but once the input is completely read lexer is not called anymore. So if I return an extra `'\n'` at the end of the file (even only if it's necessary) I won't have a chance to return `_EOF`.

As for `i--`, I can use the following instead, does it look better?

    for i+1 < len(p) && p[i+1] != '\n' {
         i++
    }

vladmos:
However now I think that explicitly calling `i--` is more straightforward than looking ahead and stopping one symbol before `\n`. Maybe I'll refactor that later but I'd prefer to keep it as it is now.

build/parse.y
```diff
@@ -162,36 +166,47 @@ file:
...
 	{
 		// If this statement follows a comment block,
 		// attach the comments to the statement.
 		if cb, ok := $<lastRule>1.(*CommentBlock); ok {
-			$$ = $1
-			$$[len($1)-1] = $2
-			$2.Comment().Before = cb.After
-			$<lastRule>$ = $2
+			$$ = append($1[:len($1)-1], $2...)
+			//$$[len($1)-1] = $2
```
laurentlb: delete comment?

laurentlb:
lgtm

vladmos:
Jenkins, retest this please

PR #198, opened 2018-01-26
Format if-else blocks

vladmos:


vladmos:
The detection of Python blocks continuation didn't work well: the continuation was detected if it followed a block before immediately but didn't work if there was an empty line between:

    if foo:
        pass
    
    else:
        pass

Given that it's a temporary solution and will be removed anyway once --format_bzl is flipped, I removed the detection, now the legacy mode treats `if:` and `else:` blocks as two separate `PythonBlock`s. The difference is minor: now a single empty line is enforced between such blocks.

vladmos:
Also, to keep the parser simple, it doesn't disallow multiple `else:` conditions or `elif:` after `else:`. On the contrary, a single `if:` in the beginning of such sequences is mandatory.

laurentlb:
When adding new nodes to the AST, please use the same types as in https://github.com/google/skylark/blob/master/syntax/syntax.go if possible.

We'd like to eventually converge the ASTs.

PR #202, opened 2018-02-02
edit: add new substitute command for buildozer

### LINKED ISSUE #182: buildozer: add "substitute" command
### Proposal

Buildozer should support a new command, `substitute`, which would let users replace labels in a specific attribute that match a label with wildcards. This would be useful for updating build files after a package is moved.

```
substitute <attr> <from_label> <to_label>
```

### Example

Suppose we want move a large set of packages from our source tree to our test tree. We need to update  tests that depend on libraries that were moved. We could do this with `replace`, but that requires one `replace` command per moved rule, so this may be inconvenient.

```
$ git mv java/com/example/testutils javatests/com/example/testutils
$ buildozer substitute deps //java/com/example/testutils/... //javatests/com/example/testutils/... //javatests/...:%java_test
```

This command modifies the `deps` attribute of all `java_test` rules in `//javatests/...`. It replaces any label that starts with `//java/com/example/testutils` with a label that starts with `//javatests/com/example/testutils` instead.

```
java_test(
    name = "foo_test",
    ...
    deps = [
-        "//java/com/example/testutils/bar",
+        "//javatests/com/example/testutils/bar",
    ],
```

### Use case

This was originally proposed for Gazelle for use in Kuberentes (see bazelbuild/bazel-gazelle#63). However, it seems like this would be generally useful to have in Buildozer.

Kubernetes' use case is vendoring. They need to be able to copy the contents of an external repository that already has BUILD files into a subdirectory of one of their repositories. There should be an easy (automatic) way to get the BUILD files to make sense again. They could do something like this:

```
$ buildozer substitute * //... //vendor/github.com/bazelbuild/buildtools/... //vendor/github.com/bazelbuild/buildtools/...
```

jayconrod:
Substitute lets users perform substitution in string values using
regular expressions. This is a more powerful command than replace; it
should be useful for updating deps after moving a large number of
libraries.

Fixes #182

### DIFF
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -363,13 +363,13 @@ func cmdSubstitute(env CmdEnvironment) (*build.File, error) {
 	newTemplate := env.Args[2]
 	for _, key := range attrKeysForPattern(env.Rule, env.Args[0]) {
 		attr := env.Rule.Attr(key)
-		if e, ok := attr.(*build.StringExpr); ok {
-			newValue, ok := stringSubstitute(e.Value, oldRegexp, newTemplate)
-			if ok {
-				env.Rule.SetAttr(key, getAttrValueExpr(key, []string{newValue}))
-			}
-		} else {
+		e, ok := attr.(*build.StringExpr)
+		if !ok {
 			ListSubstitute(attr, oldRegexp, newTemplate)
+			continue
+		}
+		if newValue, ok := stringSubstitute(e.Value, oldRegexp, newTemplate); ok {
+			env.Rule.SetAttr(key, getAttrValueExpr(key, []string{newValue}))
 		}
 	}
 	return env.File, nil
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -355,6 +355,26 @@ func cmdReplace(env CmdEnvironment) (*build.File, error) {
 	return env.File, nil
 }
 
+func cmdSubstitute(env CmdEnvironment) (*build.File, error) {
+	oldRegexp, err := regexp.Compile(env.Args[1])
+	if err != nil {
+		return nil, err
+	}
+	newTemplate := env.Args[2]
+	for _, key := range attrKeysForPattern(env.Rule, env.Args[0]) {
+		attr := env.Rule.Attr(key)
+		if e, ok := attr.(*build.StringExpr); ok {
+			newValue, ok := stringSubstitute(e.Value, oldRegexp, newTemplate)
+			if ok {
+				env.Rule.SetAttr(key, getAttrValueExpr(key, []string{newValue}))
+			}
+		} else {
+			ListSubstitute(attr, oldRegexp, newTemplate)
+		}
+	}
+	return env.File, nil
+}
+
 func cmdSet(env CmdEnvironment) (*build.File, error) {
 	attr := env.Args[0]
 	args := env.Args[1:]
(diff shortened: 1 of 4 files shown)


edit/buildozer.go
```diff
@@ -355,6 +355,26 @@ func cmdReplace(env CmdEnvironment) (*build.File, error) {
...
 
+func cmdSubstitute(env CmdEnvironment) (*build.File, error) {
+	oldRegexp, err := regexp.Compile(env.Args[1])
+	if err != nil {
+		return nil, err
+	}
+	newTemplate := env.Args[2]
+	for _, key := range attrKeysForPattern(env.Rule, env.Args[0]) {
+		attr := env.Rule.Attr(key)
+		if e, ok := attr.(*build.StringExpr); ok {
```
pmbethe09: optional:  this one is a bit tricky with nested reuse of ok.  I would probably write the same logic as:

e, ok := attr.(*build.StringExpr)
if !ok {
  ListSubstitute(attr, oldRegexp, newTemplate)
  continue
}
if newValue, ok := stringSubstitute(e.Value, oldRegexp, newTemplate); ok {
  env.Rule.SetAttr(key, getAttrValueExpr(key, []string{newValue}))
}

edit/edit_test.go
```diff
@@ -216,3 +217,57 @@ rule()`, 1, true, false, `Error for multiple unnamed rules`},
...
+	for _, tst := range tests {
+		t.Run(tst.desc, func(t *testing.T) {
+			f, err := build.Parse("BUILD", []byte(tst.input))
+			if err != nil {
+				t.Fatal(err)
+			}
+			lst := f.Stmt[0]
+			oldRegexp := regexp.MustCompile(tst.oldPattern)
+			ListSubstitute(lst, oldRegexp, tst.newTemplate)
+			got := build.FormatString(lst)
```
pmbethe09: push into if

edit/edit_test.go
```diff
@@ -216,3 +217,57 @@ rule()`, 1, true, false, `Error for multiple unnamed rules`},
...
+			if got != tst.expected {
+				t.Fatalf("got %s ; want %s", got, tst.expected)
```
pmbethe09: Tell the user what happened (and prefer Errorf over Fatalf)

t.Errorf("build.FormatString(%v) = %q; want %q", lst, got, tst.expected)

edit/edit_test.go
```diff
@@ -216,3 +217,57 @@ rule()`, 1, true, false, `Error for multiple unnamed rules`},
 		}
 	}
 }
+
+func TestListSubstitute(t *testing.T) {
+	tests := []struct {
+		desc, input, oldPattern, newTemplate, expected string
```
pmbethe09: nit: use 'want' as more Go-y than 'expected'

edit/edit_test.go
```diff
@@ -216,3 +217,57 @@ rule()`, 1, true, false, `Error for multiple unnamed rules`},
...
+	}
+
+	for _, tst := range tests {
+		t.Run(tst.desc, func(t *testing.T) {
+			f, err := build.Parse("BUILD", []byte(tst.input))
+			if err != nil {
+				t.Fatal(err)
+			}
+			lst := f.Stmt[0]
+			oldRegexp := regexp.MustCompile(tst.oldPattern)
```
pmbethe09: Only use Must* methods as file-level.  Here use Compile and call t.Fatalf if there is an error

edit/buildozer.go
jayconrod: Done

edit/edit_test.go
jayconrod: Done

edit/edit_test.go
jayconrod: Done

edit/edit_test.go
jayconrod: Done

edit/edit_test.go
jayconrod: Done. I used `ListSubstitute` instead of `FormatString` in the new message to make it clear that's the operation that's failing. The input to `FormatString` won't be readable.

pmbethe09:
Sorry I lost this while dealing with other things -- can you deal with merge issues?

jayconrod:
Sure, merge conflict is resolved now.

pmbethe09:
LGTM

@laurentlb - ok with you?

laurentlb:
LGTM
Looks super useful

PR #208, opened 2018-02-15
All of buildozer is aware of implicit names

illicitonion:
This means that 'print' commands understand implicit names, and that
implicit names are understood by default in all commands, rather than
needing to be handled as special cases everywhere.

### DIFF
--- a/build/rule.go
+++ b/build/rule.go
@@ -18,33 +18,48 @@ distributed under the License is distributed on an "AS IS" BASIS,
 
 package build
 
-import "strings"
+import (
+	"strings"
+	"path/filepath"
+)
 
 // A Rule represents a single BUILD rule.
 type Rule struct {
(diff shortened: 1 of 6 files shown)


bazel-io:

pmbethe09:
Gazelle is now in
https://github.com/bazelbuild/bazel-gazelle

What if you switch to use that?

laurentlb:
LGTM

illicitonion:
Rebased, no longer touches WORKSPACE, PTAL @laurentlb :)

laurentlb:
Still looks good

illicitonion:
Cool, thanks :) Anything blocking a merge?

mikedanese:
HI, this change is currently breaking the kubernetes build. Can we revert this temporarily until we can roll a fix?

PR #212, opened 2018-02-19
Support augmented assignments

### LINKED ISSUE #200: Augmented assignment statements other than += are not supported
There should also be `-=`, `*=`, `/=`, `//=`, `%=`.

vladmos:
All augmented assignments except `//=` are supported here. `//=` will be supported in a follow-up commit because `//` is currenlty not supported too.

Fixes #200


laurentlb:
lgtm

PR #216, opened 2018-02-23
Remove the `Suffix bool` in the Comment structure.

laurentlb:
It turns out this information is duplicated (Comments already knows
what is suffix) and not used outside the lexer. It was also confusing:
someone modifying the AST might rely on it, when it's actually ignored
by the pretty-printer.


vladmos:
LGTM

PR #217, opened 2018-02-23
Treat 'load' as a syntax keyword

laurentlb:
After this change, 'load' cannot be used as an identifier.
AST isn't changed though (it still uses a CallExpr). Updating the AST
will require changing users of the library.


laurentlb:
cc @vladmos

pmbethe09:
LGTM

PR #219, opened 2018-02-26
Introduce "Ident" node, to distinguish between numbers and idents.

laurentlb:
Add more information to the AST, which will help tools build on top of
it. The definition of Ident matches the one from
https://github.com/google/skylark/blob/master/syntax/syntax.go

This is a breaking change, some users might need to be updated.


pmbethe09:
LGTM

PR #223, opened 2018-02-27
Add new nodes "Ident" and "LoadStmt"

laurentlb:
The nodes are not yet used, so that we can update users in a
non-breaking way.

### DIFF
--- a/build/syntax.go
+++ b/build/syntax.go
@@ -117,7 +117,18 @@ func (x *PythonBlock) Span() (start, end Position) {
 	return x.Start, x.Start.add(x.Token)
 }
 
-// A LiteralExpr represents a literal identifier or number.
+// An Ident represents an identifier.
+type Ident struct {
+	Comments
+	NamePos Position
+	Name    string
+}
+
+func (x *Ident) Span() (start, end Position) {
+	return x.NamePos, x.NamePos.add(x.Name)
+}
+
+// A LiteralExpr represents a literal number.
 type LiteralExpr struct {
 	Comments
 	Start Position
@@ -432,6 +443,27 @@ type CodeBlock struct {
 
 func (x *CodeBlock) Span() (start, end Position) {
 	return x.Start, x.End.Pos
+}
+
+// A LoadStmt loads another module and binds names from it:
+// load(Module, "x", y="foo").
+//
+// The AST is slightly unfaithful to the concrete syntax here because
+// Skylark's load statement, so that it can be implemented in Python,
+// binds some names (like y above) with an identifier and some (like x)
+// without.  For consistency we create fake identifiers for all the
+// strings.
+type LoadStmt struct {
+	Comments
+	Load   Position
+	Module *StringExpr
+	From   []*Ident // name defined in loading module
+	To     []*Ident // name in loaded module
+	Rparen Position
+}
+
+func (x *LoadStmt) Span() (start, end Position) {
+	return x.Load, x.Rparen
 }
 
 // A FuncDef represents a function definition expression: def foo(List):.


laurentlb:
@mikedanese 
@jayconrod

build/syntax.go
```diff
@@ -117,7 +117,18 @@ func (x *PythonBlock) Span() (start, end Position) {
...
+	Comments
+	NamePos Position
+	Name    string
+}
+
+func (x *Ident) Span() (start, end Position) {
+	return x.NamePos, x.NamePos.add(x.Name)
+}
+
+// A LiteralExpr represents a literal number.
```
jayconrod: From the comment here, it sounds like the plan is still to change the meaning of this type. Please avoid if you can. If `Ident` is intended to replace current uses of `LiteralExpr`, you can turn `LiteralExpr` into a type alias:

```
type LiteralExpr = Ident
```

This will make the symbols `LiteralExpr` and `Ident` equivalent. You'll need to introduce a type with a new name for literal numbers.

Go is not resilient against breaking API changes. The Go build system will build with whatever library versions are installed. It will fetch from `master` when something is missing. Authors can only control the versions of their dependencies by copying code into their repositories, which is discouraged for libraries.

This is probably Go's biggest deficiency, and we're working on a [new system](https://research.swtch.com/vgo) that will give us reproducible builds. However, even in this new system, there is an important invariant:

> If an old package and a new package have the same import path, 
> the new package must be backwards-compatible with the old package.

laurentlb: I'm a bit confused. I need a new type `Ident,` which doesn't have the same fields as the current `LiteralExpr`. So there will be a breaking change at some point.

I thought introducing the new type first would allow you to support both in your code, so I can change the parser to generate `Ident` nodes.

Regarding numbers: I'm fine with using either `LiteralExpr` or a new node (`Number`), depending on what's simpler for the migration.

jayconrod: I don't think I understand how `Ident` will be different than the current `LiteralExpr`. Will it have additional fields?

The breaking change is problematic because we can never really fix (or even know about) all the uses. I can migrate Gazelle to be compatible, but Kubernetes checks out and builds specific older versions of Gazelle against tip versions of dependencies. They're planning to vendor Gazelle to avoid problems like this in the future, but past release branches would still need to be fixed. It would be a lot of engineering effort.

laurentlb: The field names are different in Ident and LiteralExpr. I changed the names, because I hope to share the same AST type with Skylark interpreter (http://github.com/google/skylark/) in the future.

jayconrod: Ah, that makes sense. So the goal would be to migrate to `github.com/google/skylark/syntax` for the Skylark AST and parser?

It's still important to avoid changing the API of `github.com/bazelbuild/buildtools/build` though, since its old clients may never be migrated. The recommendation proposed with vgo, going forward, is to publish breaking changes in packages with distinct import paths that include the major semver number. For example, this would be `github.com/bazelbuild/build/tools/build/v2`. This lets old clients fetch the old version of the package with the old import path.

Since there won't be much left in this package without the AST and parser, maybe it's better to move `Rule`, `Format`, `Rewrite`, and the other remaining definitions into other packages?

mikedanese:
I tested this in k8s, and it seemed to work.

mikedanese:
However, I'd appreciate a ping when you decide to merge so we can verify that.

build/syntax.go
laurentlb: Yes. Ideally, I'd like to share the AST and parser with Skylark (there are a few things to figure out, in particular regarding comments). So I'm currently trying to remove differences between the two ASTs.

What if there are many changes we need to do to the AST (and they are not atomic)? I can create `v2` now, but I have more breaking changes to do. It seems weird to create a new package all the time. Can't users stop depending on head and update when they want?

I don't think there are many users of the API. So I think it's okay to do incompatible changes (and give some time to both of you to update your code to avoid a breakage).

jayconrod: > I'm currently trying to remove differences between the two ASTs.

I think it's better to migrate packages that depend on the old AST directly to the new AST. Packages like `edit` that expose AST types in their own API will need a `v2`.

> What if there are many changes we need to do to the AST (and they are not atomic)?

Probably best to make these changes on a branch and merge when the API is stable.

> I don't think there are many users of the API. So I think it's okay to do incompatible changes (and give some time to both of you to update your code to avoid a breakage).

This is more complicated than just updating projects that depend on this package to work with a new API. Kubernetes maintains several old release branches. Those branches build gazelle at specific commits, but they fetch gazelle's dependencies from `master`. So one of two things would need to happen:

* Gazelle and its dependencies would need to be vendored on all old release branches. This is not the kind of change k8s will want to do on a stable, frozen release branch.
* Compatibility changes would need to be backported to every old version of Gazelle that k8s uses. They'd need to update the release branches to point to the fixed commits.

@mikedanese Let me know if I'm missing here.

There may also be other clients of this library that we don't know about and can't migrate. There are a lot of folks working with Kubernetes in private repos, and they may have their own tools. Breaking backward compatibility here is more trouble than it's worth.

laurentlb: Creating a `v2` means copying all source files in a separate directory? I really don't want to do that each time there's a change to the API. In other Bazel repositories, we do frequent breaking changes. If you look at the history of this repository, you'll probably find a lot of incompatible changes (e.g. add an argument to a public function). My recent changes to https://github.com/google/skylark/ also changed the API, but we didn't create a new version for that.

It seems to me that users who depend on our code at head take the risk of being broken. Either they accept the risk, or they use a specific revision.

I'm okay with creating a `v2` to group all AST changes, because they can be disruptive. But I'd prefer not do that for every change in the API.

> Those branches build gazelle at specific commits, but they fetch gazelle's dependencies from `master`.

This sounds like a problem in Kubernetes.

jayconrod: > It seems to me that users who depend on our code at head take the risk of being broken. Either they accept the risk, or they use a specific revision.

That is not the norm in the Go community. Authors cannot depend on a specific commit without vendoring.

> I'm okay with creating a v2 to group all AST changes, because they can be disruptive. But I'd prefer not do that for every change in the API.

I hope it's possible to avoid too many backward incompatible changes. Most changes can be made in a backward compatible way: using type aliases, functions calling new implementations with default values for new parameters, etc.

> This sounds like a problem in Kubernetes.

Kubernetes is migrating to Bazel in order to have more reproducible builds. It's a huge effort, and they're not there yet. They're vendoring Gazelle in the next release, but they still need to maintain old releases. I'd rather not create more work for them than necessary.

laurentlb: > I hope it's possible to avoid too many backward incompatible changes

When we update Buildifier inside Google, we often have to fix existing code or update tests. Any change to the parser, the pretty-printer, the refactoring functions, etc. can break existing users. A guarantee of being 100% compatible is very challenging and makes difficult to improve the tool.

> I'd rather not create more work for them than necessary.

We can create a v2 for now, to help Kubernetes. But anyone updating to v2 should be prepared to see breaking changes at head. I don't plan to have a new version for each breaking change.

mikedanese: cc @ixdy

I wouldn't recommend creating a v2 on our account. The break yesterday was a symptom of our own tech debt. We've been meaning to vendor these libraries but haven't prioritized it.

laurentlb:
This PR is not a breaking change, so I plan to submit it.

We can continue the discussion though (here or somewhere else?).

mikedanese:
SGTM

PR #225, opened 2018-02-28
Refactor if-else nodes

vladmos:
If-elif-else nodes are represented in the AST according to the Skylark-in-Go parser (https://github.com/google/skylark/blob/master/syntax/syntax.go), as a chain of nested if-else statements.

### DIFF
--- a/build/print.go
+++ b/build/print.go
@@ -203,7 +203,7 @@ func isCodeBlock(x Expr) bool {
 		return true
 	case *ForLoop:
 		return true
-	case *IfElse:
+	case *IfStmt:
 		return true
 	default:
 		return false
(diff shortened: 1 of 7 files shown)


build/print.go
```diff
@@ -506,25 +506,39 @@ func (p *printer) expr(v Expr, outerPrec int) {
...
+			p.expr(block.Cond, precLow)
+			p.printf(":")
+			p.margin += nestedIndentation
+			p.newline()
+			p.statements(block.True)
+			p.margin -= nestedIndentation
 
-			if block.If != nil {
-				p.expr(block.If, precLow)
+			isFirst = false
```
pmbethe09: set this just after access?

PR #228, opened 2018-03-07
Refactor the parser

### LINKED ISSUE #199: Tuple assignments without brackets are not supported
The following example is not handled correctly with buildifier, however it's a valid statement in Skylark:

    a, b = b, a

Previously buildifier used to format it incorrectly as 3 statements:

    a

    b = b

    a

Now it's a syntax error.

vladmos:
Make the parser compliant with the Skylark spec (https://docs.bazel.build/versions/master/skylark/spec.html)

This change adds additional syntax primitives (such as Test, Argument, Parameter), which makes the parser stricter (disallows some invalid expressions that were previously allowed, e.g. `[a = b]`, `f(*****d)`), but on the other hand allows implicit tuples that were previosly not supported: `a, b = c, d`.

Fixes #199.

### DIFF
--- a/build/print.go
+++ b/build/print.go
@@ -244,8 +244,8 @@ const (
 	precCmp
 	precAdd
 	precMultiply
-	precSuffix
 	precUnary
+	precSuffix
 	precConcat
 )
 
@@ -453,7 +453,11 @@ func (p *printer) expr(v Expr, outerPrec int) {
 		p.seq("{}", v.List, &v.End, modeList, false, v.ForceMultiLine)
 
 	case *TupleExpr:
-		p.seq("()", v.List, &v.End, modeTuple, v.ForceCompact, v.ForceMultiLine)
+		mode := modeTuple
+		if !v.Start.IsValid() {
+			mode = modeSeq
+		}
+		p.seq("()", v.List, &v.End, mode, v.ForceCompact, v.ForceMultiLine)
 
 	case *DictExpr:
 		var list []Expr
@@ -576,7 +580,9 @@ const (
 // If multiLine is true, seq avoids the compact form even
 // for 0- and 1-element sequences.
 func (p *printer) seq(brack string, list []Expr, end *End, mode seqMode, forceCompact, forceMultiLine bool) {
-	p.printf("%s", brack[:1])
+	if mode != modeSeq {
+		p.printf("%s", brack[:1])
+	}
 	p.depth++
 
 	// If there are line comments, force multiline
@@ -647,7 +653,9 @@ func (p *printer) seq(brack string, list []Expr, end *End, mode seqMode, forceCo
 		p.newline()
 	}
 	p.depth--
-	p.printf("%s", brack[1:])
+	if mode != modeSeq {
+		p.printf("%s", brack[1:])
+	}
 }
 
 // listFor formats a ListForExpr (list comprehension).
--- a/build/syntax.go
+++ b/build/syntax.go
@@ -152,18 +152,6 @@ func (x *StringExpr) Span() (start, end Position) {
 	return x.Start, x.End
 }
 
-// A MultiExpr represents a several expressions separated by a comma.
-type MultiExpr struct {
-	Comments
-	Start       Position
-	Expressions []Expr
-	End         Position
-}
-
-func (x *MultiExpr) Span() (start, end Position) {
-	return x.Start, x.End
-}
-
 // An End represents the end of a parenthesized or bracketed expression.
 // It is a place to hang comments.
 type End struct {
--- a/build/syntax.go
+++ b/build/syntax.go
@@ -43,6 +43,11 @@ func (p Position) add(s string) Position {
 	return p
 }
 
+// IsValid reports whether the position is valid (non-zero).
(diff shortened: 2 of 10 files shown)


build/print.go
```diff
@@ -244,8 +244,8 @@ const (
 	precCmp
 	precAdd
 	precMultiply
-	precSuffix
 	precUnary
+	precSuffix
```
laurentlb: revert?

build/syntax.go
```diff
@@ -305,7 +310,7 @@ func (x *SetExpr) Span() (start, end Position) {
 // A TupleExpr represents a tuple literal: (List)
 type TupleExpr struct {
 	Comments
-	Start Position
+	Start Position // can be empty if the tuple has no brackets, e.g. `a, b = x`
```
laurentlb: I don't really like this. When modifying the AST, we might lose the positions.
Is there a simple alternative?
 * Maybe have a boolean to say if there are parens?
 * Maybe don't store the information in the AST? We decide if we have parens in the printer. For example, we can always print parens, except for the case the tuple is a l-value.

vladmos: I can implement it differently, but that's how it's implemented in the Skylark-in-Go parser, I just copied it from there.

I don't think positions will be lost, even if we need to add brackets to a bracketless tuple. There's no contract that nodes should have a non-zero `Start` attribute (not all nodes even have it), the `Span` method correctly returns the positions.

build/print.go
vladmos: I did it on purpose, otherwise `-f(x)` gets formatted as `-(f(x))`. That wasn't the case previously because it used to be parsed as `{-f}{(x)}`, and is parsed now as `-{f(x)}`, which is more correct.

laurentlb: Ah sorry, I misread. Looks good.

build/syntax.go
laurentlb: My complaint is more about the code in the printer: `if !v.Start.IsValid() {`.
Positions can be lost when a users modify the AST or insert new nodes.

vladmos: I can add a boolean attribute but it won't be compatible with the Skylark-in-Go parser. Should I add it both here and there so that the implementations are still as compatible as possible?

Regarding the second approach, bracketless tuples can occur in some other situations, for instance, `a, b = c, d` (right hand side), or `dictionary[a, b]` (not sure if we should enforce brackets here).

laurentlb: I'd suggest that the printer should decide if/when tuples are parenthesized. This improves consistency and I don't think it's very controversial. We can try and see how it goes. We can refine the rules later, e.g.

`a, b = (c, d)`
`dic[(a, b)]`

build/syntax.go
```diff
@@ -314,7 +319,12 @@ type TupleExpr struct {
 }
 
 func (x *TupleExpr) Span() (start, end Position) {
-	return x.Start, x.End.Pos.add(")")
+	if x.Start.IsValid() {
+		return x.Start, x.End.Pos.add(")")
+	}
+	start, _ = x.List[0].Span()
+	_, end = x.List[len(x.List)-1].Span()
+	return
```
pmbethe09: nit: I always prefer 'return start, end' so that you don't have to look at the signature to figure out what variables were in the return

build/print.go
```diff
@@ -453,7 +453,11 @@ func (p *printer) expr(v Expr, outerPrec int) {
 		p.seq("{}", v.List, &v.End, modeList, false, v.ForceMultiLine)
 
 	case *TupleExpr:
-		p.seq("()", v.List, &v.End, modeTuple, v.ForceCompact, v.ForceMultiLine)
+		mode := modeTuple
+		if !v.Start.IsValid() {
+			mode = modeSeq
+		}
+		p.seq("()", v.List, &v.End, mode, v.ForceCompact, v.ForceMultiLine)
```
pmbethe09: thought: consistently unwrapping v.ForceCompact, v.ForceMultiLine is a lot of boilerplate that might introduce a bug sometime if we accidentally swap and pass (..., v.ForceMultiLine, v.ForceCompact) 

todo: add methods IsForceCompact(), IsForceMultiLine() to Expr and then pass an interface declaring those methods to p. seq ?

vladmos: I had a similar thought: to get rid of these attributes and calculate the mode in the printer by looking at `.Span()`s of sequences and their elements. This will make the parser and the AST simpler and in theory replaceable with the parser of Skylark in Go.

build/syntax.go
vladmos: That would require making the printer more complicated (which is ok in general, but this pull request focuses on the parser part and is already quite complicated there). The printer currently just prints expressions recursively and doesn't know whether a value is an lvalue or not. Also `a, b = c, d` looks legit to me, I'd prefer not to enforce brackets here.

I agree that if a new tuple node is added to a tree it should have brackets by default (otherwise it can break the semantics) which is not the case now. But fixing it will break compatibility with the skylark-in-go parser. Perhaps we should address the issue after we've merged the parsers?

build/syntax.go
vladmos: Done.

PR #229, opened 2018-03-09
Tests for logic operators

vladmos:


build/testdata/052.golden
```diff
@@ -12,6 +12,9 @@ foo = bar // baz
 
 foo = bar % baz
 
+# Logic operators
+not foo < bar > baz <= foo >= bar != baz in foo == bar
```
laurentlb: Note that this is actually not allowed (https://docs.bazel.build/versions/master/skylark/spec.html#binary-operators)
But if you want to have a test - why not

PR #232, opened 2018-03-20
Change a flag for switching between build and bzl modes

vladmos:
In the future the default behavior for buildifier will be to format files as if they are bzl files (at least when they're passed through stdin, when the file names are not known).

For now the default behavior will remain the same (format files as they are BUILD files) for compatibility reasons, but users will be able to provide `--type=build` explicitly to avoid breaking in the future.

### DIFF
--- a/buildifier/buildifier.go
+++ b/buildifier/buildifier.go
@@ -52,7 +52,7 @@ var (
 	disable   = stringList("buildifier_disable", "list of buildifier rewrites to disable")
 
 	// Experimental flags
-	inputType = flag.String("type", "build", "input file type: build (default, for BUILD files) or bzl (for .bzl files).")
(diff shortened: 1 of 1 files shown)


buildifier/buildifier.go
```diff
@@ -105,7 +105,18 @@ func main() {
 	build.DisableRewrites = disable()
 	build.AllowSort = allowSort()
 
-	tables.FormatBzlFiles = *formatBzlFiles
+	// Check input type.
+	switch *inputType {
+	default:
```
pmbethe09: while this works, the convention is default is the last case and other than that looks wrong.

vladmos: I copied from the switch block below, should I fix it as well?

vladmos: Done.

PR #237, opened 2018-03-22
Add a mode to detect the formatting mode automatically based on the filename

vladmos:
The mode is probably supposed to be default in the future, but not now yet.

### DIFF
--- a/buildifier/buildifier.go
+++ b/buildifier/buildifier.go
@@ -161,7 +161,7 @@ func main() {
 	diff = differ.Find()
 
 	if len(args) == 0 || (len(args) == 1 && args[0] == "-") {
-		if *inputType  == "auto" {
+		if *inputType == "auto" {
 			fmt.Fprintf(os.Stderr, "buildifier: --type=auto can't be used with stdin\n")
 			os.Exit(2)
 		}
@@ -278,7 +278,7 @@ func processFile(filename string, data []byte, inputType string) {
 		base := filepath.Base(filename)
 		if base == "BUILD" || base == "BUILD.bazel" || base == "WORKSPACE" {
 			tables.FormattingMode = tables.BuildMode
-		} else if strings.HasSuffix(filename,".bzl") {
+		} else if strings.HasSuffix(filename, ".bzl") {
 			tables.FormattingMode = tables.DefaultMode
 		}
 	}
--- a/buildifier/buildifier.go
+++ b/buildifier/buildifier.go
@@ -263,25 +263,8 @@ func processFile(filename string, data []byte, inputType string) {
 			exitCode = 3
 		}
 	}()
-
-	defaultMode := tables.FormattingMode
-	defer func() {
-		tables.FormattingMode = defaultMode
-	}()
-
-	switch inputType {
-	case "build":
-		tables.FormattingMode = tables.BuildMode
-	case "bzl":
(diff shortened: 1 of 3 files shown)


vladmos:
I'm not quite sure how I should test the code in `buildifier.go`

buildifier/buildifier.go
```diff
@@ -253,14 +256,33 @@ var diff *differ.Differ
...
-func processFile(filename string, data []byte) {
+func processFile(filename string, data []byte, inputType string) {
 	defer func() {
 		if err := recover(); err != nil {
 			fmt.Fprintf(os.Stderr, "buildifier: %s: internal error: %v\n", filename, err)
 			exitCode = 3
 		}
 	}()
 
+	defaultMode := tables.FormattingMode
```
pmbethe09: factor out to keep this function shorter -- pass back a cleanup
func setFormattingMode(inputType string) (cleanup func())

here:
cleanup := setFormattingMode(inputType)
defer cleanup()

vladmos: Like this?

buildifier/buildifier.go
```diff
@@ -368,6 +373,28 @@ func processFile(filename string, data []byte) {
...
+	switch inputType {
+	case "build":
+		tables.FormattingMode = tables.BuildMode
+	case "bzl":
+		tables.FormattingMode = tables.DefaultMode
+	case "auto":
+		base := filepath.Base(filename)
+		if base == "BUILD" || base == "BUILD.bazel" || base == "WORKSPACE" {
+			tables.FormattingMode = tables.BuildMode
+		} else if strings.HasSuffix(filename, ".bzl") {
```
vladmos: @laurentlb should it check for the .bzl extension or should it just use the bzl mode if the filename is not equal to `BUILD` or `WORKSPACE`?

laurentlb: you can decide

vladmos: Refactored.

vladmos:
Added a test to verify that the buildifier binary can be launched and handles the filenames with the `--type=auto` flag correctly.

buildifier/integration_test.sh
```diff
@@ -0,0 +1,15 @@
...
+echo -e "$INPUT" > test/BUILD
+echo -e "$INPUT" > test/test.bzl
+
+$1 --type=auto test/*
+
+echo -e "$BUILD_OUTPUT" > test/BUILD.golden
+echo -e "$BZL_OUTPUT" > test/test.bzl.golden
+
+diff test/BUILD test/BUILD.golden
+diff test/test.bzl test/test.bzl.golden
```
laurentlb: Are you sure the test fails if the diff is incorrect?
Don't you need to return non-zero?

vladmos: Yes, diff returns a non-zero code and the whole script returns a non-zero code in that case.
If you just edit one of the variables, `bazel test //...` will crash.

laurentlb: But the return value of the previous diff is ignored?

laurentlb: I suggest using `set -e` at the top (https://stackoverflow.com/a/1379904/482260)

vladmos: You're right, added `set -e` to crash on the first error.

PR #240, opened 2018-03-23
Refactor def and for statements nodes

vladmos:
Def and For nodes are represented in the AST like in the Skylark-in-Go parser (https://github.com/google/skylark/blob/master/syntax/syntax.go).

### DIFF
--- a/build/lex.go
+++ b/build/lex.go
@@ -783,10 +783,12 @@ func (in *input) order(v Expr) {
 		in.order(v.X)
 		in.order(v.Y)
 	case *LambdaExpr:
-		for _, name := range v.Var {
-			in.order(name)
+		for _, param := range v.Params {
+			in.order(param)
 		}
-		in.order(v.Expr)
+		for _, expr := range v.Body {
+			in.order(expr)
+		}
 	case *ReturnStmt:
 		if v.Result != nil {
 			in.order(v.Result)
--- a/build/lex.go
+++ b/build/lex.go
@@ -791,19 +791,19 @@ func (in *input) order(v Expr) {
 		if v.Result != nil {
 			in.order(v.Result)
 		}
-	case *FuncDef:
-		for _, x := range v.Args {
+	case *DefStmt:
+		for _, x := range v.Params {
 			in.order(x)
 		}
-		for _, x := range v.Body.Statements {
+		for _, x := range v.Body {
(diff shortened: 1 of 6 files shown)


vladmos:
Don't think so. These nodes are not enabled by default and noone should rely on the internal structure of the AST. The formatting shouldn't be affected by this change.

laurentlb:
This shouldn't break things. These AST nodes are new and used only in bzl mode (no one should rely on it yet).

build/lex.go
```diff
@@ -844,10 +844,16 @@ func (in *input) assignComments() {
...
+		// Do not assign suffix comments to file or to block statements
 		switch x.(type) {
 		case *File:
 			continue
+		case *DefStmt:
+			continue
+		case *IfStmt:
+			continue
+		case *ForStmt:
+			continue
```
laurentlb: Merge lines: `case *File, *defStmt,` ...

build/syntax.go
```diff
@@ -504,18 +492,18 @@ func (x *ReturnStmt) Span() (start, end Position) {
...
+// A ForStmt represents a for loop block: for x in range(10):.
+type ForStmt struct {
 	Comments
-	Start    Position // position of for
-	LoopVars []Expr
-	Iterable Expr
-	Body     CodeBlock
-	End      // position of the end
+	For  Position // position of for
+	Vars []Expr
```
laurentlb: Vars should be `Expr`, not `[]Expr`

build/syntax.go
```diff
@@ -472,21 +461,20 @@ func (x *LoadStmt) Span() (start, end Position) {
 	return x.Load, x.Rparen
 }
 
-// A FuncDef represents a function definition expression: def foo(List):.
-type FuncDef struct {
+// A DefStmt represents a function definition expression: def foo(List):.
+type DefStmt struct {
```
laurentlb: This doesn't match with https://github.com/google/skylark/blob/master/syntax/syntax.go#L140

build/syntax.go
vladmos: Done.

build/lex.go
vladmos: Done.

build/syntax.go
vladmos: Done.

PR #243, opened 2018-03-23
Add experimental implementation, using the AST from the Skylark interpreter

laurentlb:


vladmos:
Seems good, but maybe add a simple smoke test that runs the new binary and checks that at least a simple file is formatted correctly? Like the one in #237.

laurentlb:
Added simple test

buildifier/buildifier2.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/bazelbuild/buildtools/build"
+	"github.com/google/skylark/syntax"
+)
+
+func main() {
+	flag.Parse()
+
+	switch len(flag.Args()) {
+	case 0:
+		fmt.Println("Argument missing")
```
pmbethe09: log.Info?

buildifier/convert_ast.go
```diff
@@ -0,0 +1,250 @@
+// This file contains functions to convert from one AST to the other.
+// Input: AST from github.com/google/skylark/syntax
+// Output: AST from github.com/bazelbuild/buildtools/build
+
+package main
```
pmbethe09: I think this file belongs in its own
package astconvert 
with some of these functions public

buildifier/convert_ast.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	switch e := e.(type) {
+	case *syntax.Literal:
+		switch e.Token {
+		case syntax.INT:
+			return &build.LiteralExpr{
+				Token:    strconv.FormatInt(e.Value.(int64), 10),
+				Comments: convComments(e.Comments())}
+		case syntax.FLOAT:
+			panic("float")
```
pmbethe09: prefer log.Fatal

PR #248, opened 2018-03-26
Move experimental buildifier2 to separate directory

laurentlb:
Having two binaries in the same directory seems to confuse 'go get'.


laurentlb:
@mattmoor

laurentlb:
The test is really simple and shared with the other binary.
We can have dedicated tests later.

PR #249, opened 2018-03-26
Buildifier2: address code review comments

laurentlb:


buildifier2/buildifier2.go
```diff
@@ -27,16 +27,14 @@ func main() {
...
 		if err != nil {
-			fmt.Printf("%+v", err)
-			os.Exit(1)
+			log.Fatalf("%+v\n", err)
 		}
-		newAst := convFile(ast)
+		newAst := convertast.ConvFile(ast)
 		fmt.Print(build.FormatString(newAst))
 
 	default:
```
laurentlb: You mentioned `log.Info; os.Exit(2)` in the code review, but `log.Fatal` seems simpler. No?

pmbethe09: log.Fatal produces a nasty stack-trace that is no fun.  So it really shouldn't be used when Exit is available in main

laurentlb: That's not what I saw when I tried it.
Looking at the doc, I see that: https://golang.org/pkg/log/#Fatal

pmbethe09: Ah.  I guess I am used to the internal version

PR #252, opened 2018-03-28
Format sequences in bzl mode flexibly

vladmos:
Sequences are printed in a compact mode if there have been originally formatted in a compact mode.


build/print.go
```diff
@@ -631,60 +631,81 @@ const (
...
+			if start.Line != previousEnd {
+				return false
+			}
+			previousEnd = end.Line
+		}
+		if end != nil && previousEnd != end.Pos.Line {
+			return false
+		}
+		return true
+	} else {
```
laurentlb: remove `else`

build/print.go
```diff
@@ -631,60 +631,81 @@ const (
...
 			}
 			p.expr(x, precLow)
 		}
-
-	default:
+		// Single-element tuple must end with comma, to mark it as a tuple.
+		if len(*list) == 1 && mode == modeTuple {
+			p.printf(",")
+		}
+	} else {
```
laurentlb: add `return`, remove `else`

build/print.go

build/print.go
vladmos: Can't return here, a closing bracket should be printed in the end of the function.

pmbethe09: you could return if you duplicate the code

p.depth--
if mode != modeSeq {
	p.printf("%s", brack[1:])
}

and you could de-dup by adding just after the p.depth++ above
defer func() {
  p.depth--
  if mode != modeSeq {
	p.printf("%s", brack[1:])
  }
}()

PR #253, opened 2018-03-28
Format def statements

vladmos:
Format def statements in a multiline mode differently:

  * Use 8 spaces instead of 4 for indentation
  * If the last argument doesn't have a comment, the closing bracket is printed on the same line
  * The comma after the last argument is never printed (it's not valid to print it after *args or **kwargs)

### DIFF
--- a/build/testdata/051.bzl.golden
+++ b/build/testdata/051.bzl.golden
@@ -26,7 +26,7 @@ def function(
     def g(
             a,
             s,
-            d,  # this is d
+            d  # this is d
     ):
         ccc
 
--- a/build/testdata/051.bzl.golden
+++ b/build/testdata/051.bzl.golden
@@ -17,14 +17,17 @@ fff
 ggg
 
 def function(
-    # This shouldn't mess with indentation counting
-    x,
-    y = z,
-):
+        # This shouldn't mess with indentation counting
+        x,
+        y = z):
     aaa
     bbb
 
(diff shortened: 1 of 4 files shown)


build/testdata/051.bzl.golden
```diff
@@ -17,14 +17,17 @@ fff
...
+        x,
+        y = z):
     aaa
     bbb
 
-    def g(a, s, d):
+    def g(
+            a,
+            s,
+            d  # this is d
```
laurentlb: I wonder if we should have `d): # this is d` instead.

vladmos: This is currently not supported yet, but what if there are comments on both of these lines? (the line with the last argument and the line with `):`)

laurentlb: On which node is the comment associated?

vladmos: In the example above, to `d`, and it's not yet supported to assign comments to block headers. But it's still possible that an input file looks like the following:

    def f(  # def beginning
        x  # my argument
    ):  # def end
        pass

In theory I can format it like this:

    def f(  # def beginning
            x):  # my argument
        # def end
        
        pass

vladmos: I've decided to go with your proposal. This formatting style is not final anyway and may be changed in the future.

PR #259, opened 2018-04-03
Support pipes as binary operators

vladmos:
Pipes are not defined in the spec, but still exist in .bzl files as legacy depset operators, therefore they should be supported by a formatting tool.


pmbethe09:
LGTM

PR #260, opened 2018-04-03
load statements are not eligible for "implicit rule names"

adamjernst:
"If a build file contains exactly one unnamed rule" currently matches the load statement in the following:

```
load("//foo", "some_rule")

some_rule(
    name = "bar",
)
```

This is clearly not the intent of the implicit rule name case, so tweak it accordingly.

### DIFF
--- a/build/rule.go
+++ b/build/rule.go
@@ -31,7 +31,7 @@ type Rule struct {
 
 func (f *File) Rule(call *CallExpr) *Rule {
 	r := &Rule{call, ""}
-	if r.AttrString("name") == "" {
+	if r.AttrString("name") == "" && r.Kind() != "load" {
 		r.ImplicitName = f.implicitRuleName()
 	}
 	return r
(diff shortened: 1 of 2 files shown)


adamjernst:
@illicitonion and @laurentlb would you mind taking a look?

laurentlb:
Thanks! This unnamed rule was indeed breaking some code at Google. Unfortunately, I haven't had time to investigate (and I'm on vacation now), but Vladmos may be able to take a look.

illicitonion:
Looks good to me. Could you throw in a couple of tests, for the "load and some named rules" case, and the "load and one unnamed rule" case? Should be trivial to add to https://github.com/bazelbuild/buildtools/blob/master/build/rule_test.go - thanks!

adamjernst:
@illicitonion done! take a look and let me know.

illicitonion:
Sorry, actually looked at this with a little more context on a computer; I think the caller at https://github.com/bazelbuild/buildtools/blob/551e0839956fce9c2154fff768428c2a0960e6a4/build/rule.go#L34 also needs a `Kind != "load"` guard, otherwise both the load and the rule will end up having the implicit name.

adamjernst:
Good to go!

illicitonion:
LGTM :)

Over to people who can actually merge stuff...!

vladmos:
Thanks!

PR #262, opened 2018-04-04
Shorten cross repo labels.

uri-canva:
This allows `@com_google_guava_guava//jar:jar` to be shortened to `@com_google_guava_guava//jar` for example.

This affects both `buildifier` and `unused_deps`: `unused_deps` isn't able to detect unused deps specified with the short hand right now.

Is "cross repo" the right nomenclature?

### DIFF
--- a/edit/edit.go
+++ b/edit/edit.go
@@ -75,7 +75,7 @@ func ShortenLabel(label string, pkg string) string {
 	if !ShortenLabelsFlag {
 		return label
 	}
-	if !strings.HasPrefix(label, "//") {
+	if !strings.Contains(label, "//") {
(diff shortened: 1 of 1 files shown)


pmbethe09:
@laurentlb -- is this safe for the merge back to internal?

laurentlb:
Looks fine to me (inside Google, there's no `@` and this PR shouldn't change the behavior).

uri-canva:
giphy

uri-canva:
It's ok, I'm not in a hurry, thanks for taking a look.
I'm happy to rebase the changes on top of the other PRs if it helps.

PR #264, opened 2018-04-04
Handle @file escapes.

uri-canva:
Bazel's `-2.params` files will escape any argument starting with `@` by prepending them with `@`. This could be to escape `@file` parameter expansion but I haven't looked into it deeply enough to confirm that.


pmbethe09:
LGTM

PR #266, opened 2018-04-09
Format non-top-level statements in the default mode

vladmos:
Non-top-level statements (i.e. statements inside def, for, if or else blocks) should be formatted in the bzl mode even if buildifier is run in the build mode. Rationale: such blocks are not allowed in BUILD files in Bazel anyway, but if that's allowed anywhere else, contents of such blocks are usually not declarations of targets but some auxiliary commands.

The tests are not affected by this change because the build mode still uses legacy PythonBlock nodes that are printed as is. Once it's removed, the tests will be fixed accordingly.

### DIFF
--- a/build/print.go
+++ b/build/print.go
@@ -56,6 +56,7 @@ type printer struct {
 	comment      []Comment // pending end-of-line comments
 	margin       int       // left margin (indent), a number of spaces
 	depth        int       // nesting depth inside ( ) [ ] { }
+	level        int       // nesting level of def-, if-else- and for-blocks
 	needsNewLine bool      // true if the next statement needs a new line before it
 }
 
@@ -150,7 +151,7 @@ func (p *printer) file(f *File) {
 		p.newline()
 	}
 
-	p.statements(f.Stmt)
+	p.statements(f.Stmt, true)
 
 	for _, com := range f.After {
 		p.printf("%s", strings.TrimSpace(com.Token))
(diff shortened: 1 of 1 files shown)


build/print.go
```diff
@@ -545,21 +557,15 @@ func (p *printer) expr(v Expr, outerPrec int) {
...
 		p.printf("for ")
 		p.expr(v.Vars, precLow)
 		p.printf(" in ")
 		p.expr(v.X, precLow)
 		p.printf(":")
-		p.margin += nestedIndentation
-		p.newline()
-		p.statements(v.Body)
-		p.margin -= nestedIndentation
+		p.statements(v.Body,false)
```
laurentlb: missing space after comma, use gofmt

build/print.go
```diff
@@ -160,7 +161,18 @@ func (p *printer) file(f *File) {
 	p.newlineIfNeeded()
 }
 
-func (p *printer) statements(stmts []Expr) {
+func (p *printer) statements(stmts []Expr, isTopLevel bool) {
```
laurentlb: How is `isTopLevel` different from `p.level == 0`?
I think you can remove the argument or rename it.

build/print.go

build/print.go
vladmos: `p.level` is changed inside this function, so it's 0 when the statements on the first level are printed.

laurentlb: I still don't understand the difference between `isTopLevel` and `p.level == 0`.
Can you make clear what is `isTopLevel`?

vladmos: `isTopLevel` is true if the top statements are being printed (i.e. directly in the File node).

If statements from a, for example, def block are being printed, `p.level` will be increased by 1, but only inside the function, so it's still 0 when the function is called (given that the def-block itself is a top-level statement).

laurentlb: Do you mind clarifying the code?

vladmos: Do you have any suggestions?

I can split the function in two, `statements(stmts []Expr)` (to print just a sequence of statements) and `nestedStatements(stmts []Expr)` (to increase the indentation and level and then call `statements`), does it look better?

laurentlb: Yes, it's probably clearer than a boolean argument

PR #267, opened 2018-04-11
Use LoadStmt instead of CallExpr for load statements

vladmos:


build/print.go
```diff
@@ -209,9 +209,12 @@ func compactStmt(s1, s2 Expr, isTopLevel bool) bool {
...
 		return false
 	}
-	if isCall(s1, "load") && isCall(s2, "load") {
-		// Load statements
+	if isLoad(s1) && isLoad(s2) {
+		// Load statements should be compact
 		return true
+	} else if isLoad(s1) || isLoad(s2) {
+		// Load statements should be separated from anything else
+		return false
```
laurentlb: Is it needed?
Top-level statements already have empty line.

vladmos: Only in the build mode, in the bzl mode empty lines are preserved by default.

PR #269, opened 2018-04-12
Use Ident instead of LiteralExpr for identifiers

vladmos:


vladmos:
That's what I was going to do, just sent it here for review.

PR #282, opened 2018-04-19
Fix the AttrLiteral method

vladmos:
Now that literals and identifiers are represented as different AST nodes, the AttrLiteral method should take that into account and check for both possibilities.

### DIFF
--- a/build/rule.go
+++ b/build/rule.go
@@ -265,11 +265,14 @@ func (r *Rule) SetAttr(key string, val Expr) {
 // If the rule has no such attribute or the attribute is not an identifier or number,
 // AttrLiteral returns "".
 func (r *Rule) AttrLiteral(key string) string {
-	lit, ok := r.Attr(key).(*Ident)
-	if !ok {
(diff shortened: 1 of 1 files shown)


build/rule.go
```diff
@@ -265,11 +265,14 @@ func (r *Rule) SetAttr(key string, val Expr) {
...
 func (r *Rule) AttrLiteral(key string) string {
-	lit, ok := r.Attr(key).(*Ident)
-	if !ok {
-		return ""
+	value := r.Attr(key)
+	if ident, ok := value.(*Ident); ok {
+		return ident.Name
 	}
-	return lit.Name
+	if literal, ok := value.(*LiteralExpr); ok {
```
laurentlb: `A LiteralExpr represents a literal number.`
I'm not sure if it makes sense to have a function that returns either an identifier, or a number stored as string.

vladmos: There are usages that expect the value to be either `1` or `True`, for instance.

PR #284, opened 2018-04-19
Update InsertLoad to respect docstring comments.

timpeut:
Loads should be added after the  """docstring comment""" at the top of the file.


laurentlb:
LGTM

PR #286, opened 2018-04-24
Use --type=auto by default

vladmos:


pmbethe09:
bold move.  I like it.

laurentlb:
In my opinion, the bold move was #278 (reformat def/if/for blocks even in build mode). :)

PR #288, opened 2018-04-25
Introduce new parse methods

vladmos:
Currently buildifier chooses the formatting mode by looking at the global variable `tables.FormattingMode`. This approach has disadvantages: it's hard to use buildifier as a library (not clear at which step the variable should be set), it makes it impossible to format files of different types in parallel.

The proposed approach is to save the information about the file to the File object during the parsing step. In the end there should be a `Parse` method that does it automatically by looking at the file name, and two explicit methods `ParseBuild` and `ParseDefault` for selecting the build or default mode for a given file.

Currently the `Parse` method is already used for BUILD files, so it's temporarily kept as an alias for `ParseBuild`. The next step will be to make all the existing users call `ParseBuild` explicitly and change the behavior of `Parse` to make it universal.

### DIFF
--- a/build/lex.go
+++ b/build/lex.go
@@ -25,12 +25,36 @@ import (
 	"sort"
 )
 
-// Parse parses the input data and returns the corresponding parse tree.
+// ParseBuild parses a file, marks it as a BUILD file and returns the corresponding parse tree.
 //
 // The filename is used only for generating error messages.
-func Parse(filename string, data []byte) (*File, error) {
+func ParseBuild(filename string, data []byte) (*File, error) {
 	in := newInput(filename, data)
-	return in.parse()
+	f, err := in.parse()
+	if err == nil {
+		f.Build = true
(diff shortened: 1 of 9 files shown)


build/lex.go
```diff
@@ -25,12 +25,36 @@ import (
...
-// Parse parses the input data and returns the corresponding parse tree.
+// ParseBuild parses a file, marks it as a BUILD file and returns the corresponding parse tree.
 //
 // The filename is used only for generating error messages.
-func Parse(filename string, data []byte) (*File, error) {
+func ParseBuild(filename string, data []byte) (*File, error) {
+	in := newInput(filename, data)
+	f, err := in.parse()
+	if err == nil {
+		f.Build = true
```
laurentlb: test if `f != nil` instead?

vladmos: Done.

PR #293, opened 2018-05-01
Omit trailing comma in calls with *args and **kwargs parameters

andyscott:
The trailing comma currently causes Skylark to fail to load formatted files.

### DIFF
--- a/build/print.go
+++ b/build/print.go
@@ -744,6 +744,7 @@ func (p *printer) seq(brack string, start *Position, list *[]Expr, end *End, mod
 		indentation = defIndentation
 	}
 	p.margin += indentation
+
 	for i, x := range *list {
 		// If we are about to break the line before the first
 		// element and there are trailing end-of-line comments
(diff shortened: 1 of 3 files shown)


build/print.go
```diff
@@ -758,10 +758,26 @@ func (p *printer) seq(brack string, start *Position, list *[]Expr, end *End, mod
 
 		p.newline()
 		p.expr(x, precLow)
-		// Don't print a comma after the last element in modeParen and in modeDef
-		if !(mode == modeDef || mode == modeParen) || i+1 < len(*list) {
-			p.printf(",")
+
+		// Situations where we don't print a final comma
```
pmbethe09: instead of this comment, make this a well-named method
func needsComma(...) bool

then
if needsComma(...) {
  p.printf(",")
}

pmbethe09:
also -- I feel like this should be for .bzl mode only as **kwargs is not legal AFAIK in BUILD mode

andyscott:
Sounds good. I'll get those changes up later today.

andyscott:
@pmbethe09 I've addressed the main comments.  I checked and `**kwargs` are forbidden in BUILD files. Since the code is invalid regardless, is it worth adding a check to disable the omitted comment for this situation?

laurentlb:
Don't worry about `**` being forbidden in BUILD files. Buildifier shouldn't care about that.

PR #294, opened 2018-05-02
Use sort.SliceStable instead of sort.Slice

vladmos:
In some cases the lexer adds comments with an empty string as a token to the tree to separate comment blocks inside a statement (e.g. if a list contains line comments inside). Such fake comments don't have real `Start.Byte` values, instead the value is the same as for the next real comment line.

Later, when the comments are sorted to restore the possibly broken order, such comments order can be broken by unstable sorting algorithms.

### DIFF
--- a/build/lex.go
+++ b/build/lex.go
@@ -815,7 +815,7 @@ func (in *input) assignLineComments() {
 		}
 		// Line comments can be sorted in a wrong order because they get assgined from different
 		// parts of the lexer and the parser. Restore the original order.
-		sort.Slice(xcom.Before, func(i, j int) bool {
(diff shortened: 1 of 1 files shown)


pmbethe09:
LGTM

vladmos:
In some cases the lexer adds comments with an empty string as a token to the tree to separate comment blocks inside a statement (e.g. if a list contains line comments inside). Such fake comments don't have real `Start.Byte` values, instead the value is the same as for the next real comment line.

Later, when the comments are sorted to restore the possibly broken order, such comments order can be broken by unstable sorting algorithms.

PR #295, opened 2018-05-03
Make build.Parse automatically choose the formatting mode by the filename

vladmos:


build/lex.go
```diff
@@ -51,10 +54,16 @@ func ParseDefault(filename string, data []byte) (*File, error) {
...
 // Parse parses the input data and returns the corresponding parse tree.
 //
-// Currently an alias for ParseBuild for compatibility reasons, in the future will delegate
-// either to ParseBuild or to ParseDefault depending on the filename.
+// Uses the filename to detect the formatting type (either build or default) and calls
+// either ParseBuild or to ParseDefault correspondingly.
 func Parse(filename string, data []byte) (*File, error) {
-	return ParseBuild(filename, data)
+	basename := filepath.Base(filename)
+	for _, name := range buildFilenames {
```
pmbethe09: Do we worry about case-insensitive file systems?  on Mac/Win what if a user creates a file 'build'?

Also -- why not a go-style set?
var buildFilenames = map[string]bool{
  "BUILD": true,
  ...
}

or as before a multi-case
switch basename{
  case "BUILD", ...:
    return ParseBuild(...
  default:
     return ParseDefault(...

vladmos: I'll convert everything to lowercase and use a map-like set.

I replaced a switch block with this because this way it can be modified, for example, if a custom API user wants to treat their own skylark files as BUILD files.

vladmos: Implemented case-insensitive check.

PR #305, opened 2018-05-15
Indented comments

vladmos:
In Skylark comments are allowed to be indented freely, not necessarily following the current indentation level. If an indented block ends with a comment, like the following:

    def f():
        pass
        # indentation level doesn't matter
    f()

it's not clear if the comment belongs to the block or to the outer level. This PR disallows indented blocks to end with a comment, so they will be treated either as standalone outer level comments or comments attached the next statement, depending on the existence of an empty line between the comment and the next statement. In theory it's possible to allow the following:

    def f():
        pass
        # the comment is attached to "pass" because of the following empty line

    f()

but for the simplicity and consistency it's better to always move trailing comments from an indented block to the level above. Otherwise the following may occur: if a user adds a line (`f()`) to the example below:

    def f():
        pass
        # the comment is attached to "pass" because of the following empty line

the comment will also be moved:

    def f():
        pass

    # the comment is attached to "pass" because of the following empty line
    f()

which is not very straightforward

### DIFF
--- a/build/parse.y
+++ b/build/parse.y
@@ -33,7 +33,7 @@ package build
 
 	// supporting information
 	comma     Position   // position of trailing comma in list, if present
-	lastRule  Expr  // most recent rule, to attach line comments to
+	lastStmt  Expr  // most recent rule, to attach line comments to
 }
 
 // These declarations set the type for a $ reference ($$, $1, $2, ...)
@@ -203,7 +203,7 @@ suite:
 			statements = append($2, $4...)
 		}
 		$$ = statements
-		$<lastRule>$ = $<lastRule>4
+		$<lastStmt>$ = $<lastStmt>4
 	}
 |	simple_stmt linebreaks_opt
 	{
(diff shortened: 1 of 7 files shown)


build/parse.y
```diff
@@ -275,29 +276,81 @@ stmts:
...
 		$$ = $1
+		$<lastStmt>$ = $1[len($1)-1]
 	}
 |	block_stmt
 	{
 		$$ = []Expr{$1}
+		$<lastStmt>$ = $1
+
+		// If the block statement ends with a comment block remove it and place
+		// after the block statement.
```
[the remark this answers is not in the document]

build/parse.y
```diff
@@ -1029,3 +1045,47 @@ func forceMultiLine(start Position, list []Expr, end Position) bool {
...
+	var body *[]Expr
+	switch block := stmt.(type) {
+	case *DefStmt:
+		body = &block.Body
+	case *ForStmt:
+		body = &block.Body
+	case *IfStmt:
+		// find the innermost node in the if-elif-else chain
+		inner := block
+		for {
```
pmbethe09: opt: I tend to avoid these sorts of conditionless loops w/ break in favor of standalone funcs that use return.  In this case you could use recursion to walk the chain.

PR #306, opened 2018-05-16
Indent contents of docstrings when they're reindented

vladmos:


laurentlb:
All tests have a `def` inside a `if`.
Can we have tests with just a `def` (which is the case we care about)?

vladmos:
Added more tests.

build/rewrite.go
```diff
@@ -888,3 +896,70 @@ func (args loadArgs) Less(i, j int) bool {
...
+					break
+				}
+			}
+		}
+		if i != len(lines)-1 && isAllSpaces(line) {
+			// Replace a line that contains only spaces with an empty line
+			// unless it's the last line that's responsible for the indentation of the closing """
+			line = ""
+		}
+		lines[i] = line
```
laurentlb: As we're processing the string, we could remove trailing whitespace at the same time.

build/rewrite.go
```diff
@@ -888,3 +896,70 @@ func (args loadArgs) Less(i, j int) bool {
...
 }
+
+// indentDocstrings modifies the value of docstrings if their indentation level has been changed
+func indentDocstrings(f *File, info *RewriteInfo) {
+	Walk(f, func(v Expr, stk []Expr) {
+		def, ok := v.(*DefStmt)
+		if !ok {
+			return
+		}
+		if len(def.Body) == 0 {
```
laurentlb: Merge both `if`

vladmos: Done.

build/rewrite.go
vladmos: Done. If I do that, should I also modify docstrings that are not being indented to remove trailing spaces from their lines? `"indentdocstrings"` will need to be renamed to `"formatdocstrings"` in that case.

vladmos: Remove trailing space from all docstrings for consitency.

PR #313, opened 2018-05-17
[buildozer] Add ability to print separators in output

nataliejameson:
Sometimes we want to print out multiple fields in print commands. If any of those fields are a list, it causes the number of columns to be variable, and it's kind of hard to parse things out. This also applies to when arrays are flattened. If there is a space in any of the elements, we can't tell one element from multiple elements. This patch allows us to specify how things should be printed out. 

Also added a simple integration test that passed with bazel test edit/... . Lemme know if there's a different place to write those integration style tests, as I didn't see anything.

### DIFF
--- a/buildozer/main.go
+++ b/buildozer/main.go
@@ -47,6 +47,9 @@ var (
 
 	shortenLabelsFlag  = flag.Bool("shorten_labels", true, "convert added labels to short form, e.g. //foo:bar => :bar")
 	deleteWithComments = flag.Bool("delete_with_comments", true, "If a list attribute should be deleted even if there is a comment attached to it")
+	printListSep       = flag.String("print_list_separator", " ", "when using the print command, characters that should separate list elements")
+	printFieldSep      = flag.String("print_field_separator", " ", "when using the print command, characters that should separate fields")
+	printLineSep       = flag.String("print_line_separator", "\n", "when using the print command, characters that should separate lines")
 )
 
 func stringList(name, help string) func() []string {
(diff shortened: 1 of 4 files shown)


philipjameson:
I should be in the fb CLA org now.

philipjameson:
@googlebot One more once

laurentlb:
Thanks!

(sorry, I'll be on vacation for a few days)

philipjameson:
I'd lean towards json if that's what we'd want to do as it's a little more universally parsable. I was actually toying with just taking the normal protobuf objects and marshaling them with the json marshaler instead, but it seems like maybe the protobuf object has some internal details and that like that get a bit messy (e.g. I think there was a 'bazel_quoted_string' object or something like that at one point).

philipjameson:
Holy cow I let this sit for a long time. @laurentlb ?

laurentlb:
@vladmos, what do you think?

I think it would be more valuable to have a JSON output. The separators flags look like a partial workaround and might not be enough (separators might be ambiguous unless we also have proper escaping of the content).

Possibly a command `buildozer print_json` or a flag to select json output

philipjameson:
Yeah, I think something like --json should work. The initial concern was that for large changes, you can't really read iteratively (I'm using this in a tool that outputs a fair chunk of data), but that may be negligible. Should this just be the protobuf format, or something different?

philipjameson:
Added in a --output_format field (can rename if desired), added a few more tests, and removed the extra flags I had before. This now just prints out the json proto format.

hick209:
Any updates on this? I see this here got merged https://github.com/bazelbuild/buildtools/pull/612, should this PR be closed? Else what's the hold on this PR?

oreflow:
+1 on https://github.com/bazelbuild/buildtools/pull/313#issuecomment-3715328622 Looks like this was resolved to a large extent already, so closing it. feel free to reopen if it's still relevant

Issue #316, opened 2018-05-30
Allow builifier and buildozer to be run via a rule


ash2k:
Would be great to be able to use buildifier and buildozer via a rule that downloads what's needed and builds these tools automatically. `go get ...` is not the best user experience.
Gazelle is very convenient to use https://github.com/bazelbuild/bazel-gazelle#running-gazelle-with-bazel

ash2k:
Related https://github.com/bazelbuild/buildtools/issues/170

pmbethe09:
What do you mean by 'rule'?  
In a bazel world if you include `com_github_bazelbuild_buildtools` in your WORKSPACE
then you can build/run/use buildifier/buildozer via
@com_github_bazelbuild_buildtools//buildifier
@com_github_bazelbuild_buildtools//buildozer

pmbethe09:
ah, I see you want to operate on files in the current workspace.

https://blog.bazel.build/2018/04/11/bazel-0.12.html
there is a new flag --direct_run plus environment variables in that post for doing as you want

jacobstr:
I wanted to explore the suggestion of including `com_github_bazelbuild_buildtools` in my WORKSPACE. There are some ergonomic issues.

The (abridged) setup. Running bazel 0.11.1.

```
# WORKSPACE
go_repository(
    name = "com_github_bazelbuild_buildtools",
    importpath = "github.com/bazelbuild/buildtools",
    commit = "331d283a59329fee6ca033fbaf8f7560b2be3718",
)
```

```
# BUILD
alias(
    name = "buildozer",
    actual = "@com_github_bazelbuild_buildtools_two//buildozer"
)
```

The result:

```
bazel run //:buildozer
ERROR: /private/var/tmp/_bazel_jacob/356a8a15beef06fa16abc554ad147a1a/external/com_github_bazelbuild_buildtools/buildozer/BUILD.bazel:10:1: @com_github_bazelbuild_buildtools//buildozer:buildozer: no such attribute 'library' in 'go_binary' rule
```

I recognize the error as relating to a change from `importpath` to `embed` in rules_go. It looks like something else is already including `com_github_bazelbuild_buildtools` and the `go_repository` rule in  my workspace is being shadowed by it (the contents of the `BUILD.bazel` do not match the contents of the file at the commit I want, which does indeed use `embed`).

> Incidentally, if someone has a protip on how track what's pushing this external dependency into my workspace, I'm very curious.

[...]

PR #318, opened 2018-06-01
Rule to conveniently use buildifier via Bazel

ash2k:
Updates #316.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,5 +1,8 @@
 workspace(name = "com_github_bazelbuild_buildtools")
 
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
 git_repository(
     name = "io_bazel_rules_go",
     # 0.11.1
@@ -13,18 +16,24 @@ http_archive(
     url = "https://github.com/bazelbuild/bazel-gazelle/releases/download/0.11.0/bazel-gazelle-0.11.0.tar.gz",
 )
 
-load(
-    "@io_bazel_rules_go//go:def.bzl",
-    "go_rules_dependencies",
-    "go_register_toolchains",
+http_archive(
+    name = "bazel_skylib",
+    sha256 = "57e8737fbfa2eaee76b86dd8c1184251720c840cd9abe5c3f1566d331cdf7d65",
+    strip_prefix = "bazel-skylib-0.4.0",
+    url = "https://github.com/bazelbuild/bazel-skylib/archive/0.4.0.tar.gz",
 )
+
+load("@io_bazel_rules_go//go:def.bzl", "go_register_toolchains", "go_rules_dependencies")
 load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")
+load("@com_github_bazelbuild_buildtools//buildifier:deps.bzl", "buildifier_dependencies")
 
 gazelle_dependencies()
 
 go_rules_dependencies()
 
 go_register_toolchains()
+
+buildifier_dependencies()
 
 # used for build.proto
 http_archive(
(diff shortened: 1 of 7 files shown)


buildifier/def.bzl
```diff
@@ -0,0 +1,62 @@
+load("@bazel_skylib//:lib.bzl", "shell")
+
+def _buildifier_impl(ctx):
```
ash2k: This is written looking at how Gazelle does it.

buildifier/runner.bash.template
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+BUILDIFIER_SHORT_PATH=@@BUILDIFIER_SHORT_PATH@@
+ARGS=@@ARGS@@
+
+buildifier_short_path=$(readlink "$BUILDIFIER_SHORT_PATH")
+cd "$BUILD_WORKSPACE_DIRECTORY"
```
ash2k: This is using new Bazel 0.12 env variable. See https://blog.bazel.build/2018/04/11/bazel-0.12.html

buildifier/def.bzl
```diff
@@ -0,0 +1,62 @@
...
+    attrs = {
+        "verbose": attr.bool(
+            doc = "Print verbose information on standard error",
+        ),
+        "mode": attr.string(
+            default = "fix",
+            doc = "Formatting mode",
+            values = ["check", "diff", "fix"],
+        ),
+        "exclude_path_patterns": attr.string_list(
```
ash2k: I'm not happy with this attribute name. Any suggestions?

laurentlb: exclude_paths?
exclude_patterns?

Attributes can specify documentation (see 'doc' in https://docs.bazel.build/versions/master/skylark/lib/attr.html#string_list). Maybe mention here how it should look like.

buildifier/def.bzl
```diff
@@ -0,0 +1,62 @@
...
+            allow_empty = True,
+            default = [],
```
laurentlb: No need to specify the default value (it's already the empty list).

ash2k:
@laurentlb Thanks for review! I've addressed the comments. Also added "showlog" attribute.

ash2k:
@laurentlb I'm a bazel newbie and I had these questions while working on this change:
- Is there a way to tell bazel to capture stdout from a `ctx.actions.run()` and put it into an output file? Creating a wrapper bash script just to do that is a bit annoying. Had this issue in another rule.
- Would be useful to be able to "attach" command line arguments to an executable action. Right now the only way to pass information is runfiles and/or a templated wrapper script (like in this PR). Is there a way to do this already? I've just copied what Gazelle does and was not able to find a better way. I.e. I'd like to do this:
```bzl
    return [DefaultInfo(
        files = depset([out_file]),
        runfiles = runfiles,
        executable = out_file,
        args = args,      # <------ THIS
    )]
```

WORKSPACE
```diff
@@ -1,31 +1,38 @@
 workspace(name = "com_github_bazelbuild_buildtools")
 
-git_repository(
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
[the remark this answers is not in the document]
laurentlb: Thanks for this change!

laurentlb:
Short answer is no, we don't provide a better way at the moment.

Capturing stdout a common limitation, we should do something for this.
Attaching command-line arguments is a good idea. I wonder if it should be built-in in Bazel, or if an implementation in a bzl file is enough.

Apparently tests are failing. Any idea what's going wrong?

ash2k:
> Apparently tests are failing. Any idea what's going wrong?

I have rolled back rules_go and gazelle update and everything is fine now. I don't know what is wrong, something with protobuf generation if I'm understanding it correctly (I might be not). I'll leave it to someone who understands the problem to update them later.

laurentlb:
Thanks!

ash2k:
@laurentlb FYI I've filled https://github.com/bazelbuild/bazel/issues/5511

PR #319, opened 2018-06-04
Fix indented comments

vladmos:
This PR fixes the same thing as #305, but this time correclty handles the situation if an indented block has multiple comments in the end (all of them are to be not indented, not just the last one).

### DIFF
--- a/build/parse.y
+++ b/build/parse.y
@@ -301,9 +301,9 @@ stmt:
 	{
 		$$ = []Expr{$1}
 		$<lastStmt>$ = $1
-		if cb := extractTrailingComment($1); cb != nil {
-			$$ = append($$, cb)
-			$<lastStmt>$ = cb
+		if cbs := extractTrailingComments($1); len(cbs) > 0 {
+			$$ = append($$, cbs...)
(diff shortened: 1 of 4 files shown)


build/parse.y
```diff
@@ -1062,25 +1062,36 @@ func forceMultiLineComprehension(start Position, expr Expr, clauses []Expr, end
...
+	var comments []Expr
 	if body != nil && len(*body) > 0 {
-		lastStmt := (*body)[len(*body)-1]
-		if cb, ok := lastStmt.(*CommentBlock); ok {
-			// Remove the comment block
-			*body = (*body)[:len(*body)-1]
-			return cb
+		// Detach and return all trailing comment blocks
+		for i := len(*body)-1; i >= 0; i-- {
+			if cb, ok := (*body)[i].(*CommentBlock); ok {
```
laurentlb: reverse the condition, remove the continue

vladmos: If I reverse the condition I'll need to assign `cb` to a variable that's defined outside of this block. Will it be more readable?

laurentlb: I think so.

PR #320, opened 2018-06-05
Ensure nested calls aren't treated as rules.

katre:
Updates #317.

### DIFF
--- a/build/rule.go
+++ b/build/rule.go
@@ -51,8 +51,7 @@ func (f *File) Rules(kind string) []*Rule {
 
 			// Skip nested calls.
 			for _, frame := range stk {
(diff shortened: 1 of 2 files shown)


build/rule.go
```diff
@@ -48,12 +48,22 @@ func (f *File) Rules(kind string) []*Rule {
 			if !ok {
 				return
 			}
+
+			// Skip nested calls.
+			for _, frame := range stk {
+				_, ok := frame.(*CallExpr)
```
vladmos: Nit:

    if _, ok := frame.(*CallExpr); ok {
        return
    }

katre: Done.

PR #328, opened 2018-06-11
Allow postfix traversal of the AST.

laurentlb:


laurentlb:
I'm not happy with the name `Walk2`.
Ideally, I'd like rename `Walk` to `WalkPrefix` and `Walk2` to `Walk`. Downside is that it's an API change.

build/walk.go
```diff
@@ -23,14 +23,28 @@ package build
...
+		}
+		return nil
+	})
+}
+
+// Walk2 is similar to Walk1, except that the function is also called with a nil
+// argument after we've visited the node and all its children. This is useful to
+// run code in postorder traversal (the last item of stk contains the current
+// node).
+func Walk2(v Expr, f func(x Expr, stk []Expr)) {
```
pmbethe09: how about 
WalkIncludingNil

build/walk_test.go
```diff
@@ -0,0 +1,43 @@
+package build
+
+import "testing"
+
+func nodeToString(e Expr) string {
+	if bin, ok := e.(*BinaryExpr); ok {
+		return bin.Op
+	} else if lit, ok := e.(*LiteralExpr); ok {
```
pmbethe09: can drop else

build/walk_test.go
```diff
@@ -0,0 +1,43 @@
...
+		return lit.Token
+	} else {
```
pmbethe09: can drop else, return in previous branch

build/walk_test.go
laurentlb: done

build/walk_test.go
laurentlb: done

build/walk.go
laurentlb: WalkIncludingNil doesn't look readable to me.
Renamed to `WalkWithPostfix`. What do you think?

build/walk.go
```diff
@@ -23,14 +23,28 @@ package build
...
+	var stack []Expr
+	walk1(&v, &stack, func(x Expr, stk []Expr) Expr {
+		if x != nil {
+			f(x, stk)
+		}
+		return nil
+	})
+}
+
+// Walk2 is similar to Walk1, except that the function is also called with a nil
```
pmbethe09: comment is out of date

build/walk.go
```diff
@@ -42,7 +56,7 @@ func Edit(v Expr, f func(x Expr, stk []Expr) Expr) Expr {
 	return walk1(&v, &stack, f)
 }
 
-// walk1 is the actual implementation of Walk and WalkAndUpdate.
+// walk1 is the actual implementation of Walk, Walk2, and Edit.
```
pmbethe09: s/Walk2/WalkWithPostfix/

laurentlb:
Please take another look. I've found that the function was not flexible enough (e.g. user cannot decide to stop the recursion). 
I've added `WalkOnce`, a simpler function that does just one thing. It seems much more powerful, because the caller has full control.

(cc @alandonovan because I'd suggest the same function in https://github.com/google/skylark)

PR #335, opened 2018-07-02
Fix the pipe operator precedence

### LINKED ISSUE #280: shift/reduce conflicts
As of recent grammar changes for skylark we now get

`conflicts: 58 shift/reduce`

when running: go tool yacc

Worth investigating.

vladmos:
`a | b | c` was erroneously formatted as `a | (b | c)` because the tree was parsed right-to-left.

Partially fixes #280.

### DIFF
--- a/build/parse.y
+++ b/build/parse.y
@@ -164,7 +164,7 @@ package build
 %left  _AND
 %left  '<' '>' _EQ _NE _LE _GE _NOT _IN
 %left  '+' '-'
-%left  '*' '/' '%',
+%left  '*' '/' '%'
 %left  '|'
 %left  '.' '[' '('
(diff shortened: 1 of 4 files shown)


build/parse.y
```diff
@@ -164,7 +164,8 @@ package build
 %left  _AND
 %left  '<' '>' _EQ _NE _LE _GE _NOT _IN
 %left  '+' '-'
-%left  '*' '/' '%'
+%left  '*' '/' '%',
```
laurentlb: what is the trailing comma?

vladmos: A typo, although it doesn't affect the behavior (i.e. the generated parse.y.go is not affected). Removed.

PR #336, opened 2018-07-03
Add support for "-2.params" files that contain --direct_dependencies.

brown:
Newer versions of Bazel write "-2.params" files with a different format.
This patch handles the change.

### DIFF
--- a/unused_deps/unused_deps.go
+++ b/unused_deps/unused_deps.go
@@ -18,6 +18,7 @@ limitations under the License.
 package main
 
 import (
+	"archive/zip"
 	"bufio"
 	"bytes"
 	"errors"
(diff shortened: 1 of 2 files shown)


brown:
I signed it!

unused_deps/unused_deps.go
```diff
@@ -116,9 +117,54 @@ func inputFileName(blazeBin, pkg, ruleName, extension string) string {
 	return fmt.Sprintf("%s/%s/%s.%s", blazeBin, pkg, ruleName, extension) // *_{binary,test}
 }
 
-// directDepParams returns --direct_dependency entries from paramsFileName (a jar-2.params file)
-// as a map from jar files to labels.
-func directDepParams(paramsFileNames ...string) (depsByJar map[string]string) {
+// targetLabel reads the manifest of the jar specified by jarFileName.  If the
+// manifest has a "Target-Label" property, targetLabel returns its value.
+func targetLabel(jarFileName string) (string, error) {
+	const BAZEL_TARGET_KEY = "Target-Label:"
```
pmbethe09: up to you: make this a file-level const

style: still use lowerCamelCase for variables, even constants

unused_deps/unused_deps.go
```diff
@@ -116,9 +117,54 @@ func inputFileName(blazeBin, pkg, ruleName, extension string) string {
...
+			return "", err
+		}
+		defer rc.Close()
+
+		contents, err := ioutil.ReadAll(rc)
+		if err != nil {
+			return "", err
+		}
+		key := []byte(BAZEL_TARGET_KEY)
+		found := bytes.Index(contents, key)
```
pmbethe09: I guess the idea with using bytes.Index is to avoid converting all of the bytes from the rc file to string?
It just seems a little obfuscated and you've already read all of the bytes into memory

option 1) change to 'b, err := ioutil.ReadAll(rc) ...' then 'contents := string(b)'
and do all of the rest on the string

option 2) use a bufio.Scanner on rc to avoid having everything in memory at the same time.  Skip until start, then read/copy until \n

brown: Sorry about the obfuscation.  I haven't written much Go code recently ... forgot that strings are not encoded.  Thanks very much for reviewing.

It turns out that my code for parsing the value of "Target-Label" did not take into consideration continuation lines, so it truncated long values.  I have fixed that problem in the revised commit I uploaded.

brown:
I have pushed a new version of the commit.  It should address your review comments and also handles jar file manifest value continuation lines.  The spec for jar manifest key/value pairs is here:  https://docs.oracle.com/javase/8/docs/technotes/guides/jar/jar.html#Manifest_Specification

unused_deps/unused_deps.go
```diff
@@ -116,9 +117,80 @@ func inputFileName(blazeBin, pkg, ruleName, extension string) string {
...
+// newline and the index of the byte following the newline.
+func manifestIndexNewline(manifest string) (int, int, error) {
+	// A newline in a jar manifest is denoted with CR LF, CR, or LF.
+	n := strings.IndexByte(manifest, '\n')
+	r := strings.IndexByte(manifest, '\r')
+	if n < 0 && r < 0 {
+		return -1, -1, fmt.Errorf("no newline in '%s'", manifest)
+	}
+
+	if n >= 0 {
```
pmbethe09: invert to reduce indent?

if n < 0 {
  return r, r + 1, nil
}
if n < r {
  ...

brown:
Reduced indentation.

pmbethe09:
LGTM

petroseskinder:
woo!

Issue #34, opened 2017-01-12
Make indent and quote style configurable.


bolinfest:
To get this to follow Buck's conventions, I had to make some small changes:

https://github.com/bolinfest/buildifier/commit/68b6d08b1e2dd2887ae1d786e20a33abf3bf98c9

It would be nice if these things were configurable, as we would prefer not to maintain a fork.

pmbethe09:
I saw your indent flag PR: https://github.com/bazelbuild/buildifier/pull/35
That seems fine since the default configuration can be unchanged.

If you want to do the same for those other features, that sounds fine to me, as long as the defaults are preserved as is.  Too many BUILD files inside Google would be broken otherwise.

bolinfest:
Yep, now that I got legal signoff to contribute, I'll put up a PR for this!

pmbethe09:
And as with the other PR, give existing defaults in the core package for library-based usage.  Thanks.

laurentlb:
Hi,

My preference is to avoid this kind of settings, and have everyone use the same style. Tools based on Buildifier (e.g. Buildozer) are going to modify files using the standard formatting and ignore your options. Would it be an option for you to adopt Buildifier default style?

bolinfest:
This is the style we have been using at Facebook. As the person who
introduced the style, I'd rather keep what we have and avoid updating
hundreds of our build files just to flip the quotes and lengthen the lines.

bolinfest:
Obviously we could also fork, but I would prefer not to do that because
then it's unlikely that we'll contribute anything else back going forward.

bolinfest:
Incidentally, Buildozer does not appear to be a public tool. Is it going to be? If so, when? And on what timeline?

bolinfest:
Without knowing the details of what Buildozer does, if using it were really important to us, presumably we could use `buildifier` to reformat our files in the Google format, run Buildozer to do whatever it does, and then run `buildifier` again with our settings to convert them back.

pmbethe09:
buildozer was open sourced in this repo a month ago
https://github.com/bazelbuild/buildifier/commit/91f1a2a9ca5ace0b06d86153d7521492348e54bd
with a bug-fix Dec 15
https://github.com/bazelbuild/buildifier/commit/736c577fa275bbc4176f19d0c654cb3f93b12062

bolinfest:
[...]

PR #339, opened 2018-07-10
Add support for BUCK syntax

rmaz:
There are some small differences in BUCK syntax that required changes to the buildifier:

- Buck does not support the shortened name syntax of Bazel
- Buck does not handle escaped unicode in its BUCK files (unclear to me why Latin-1 is a requirement here, the skylark docs seem to mention unicode strings are fine)

This PR adds a type attribute instead of a bool to know which output type we are dealing with, and gates the `label` rewrite function as well as skips unicode escaping. Its missing some tests, but I wanted to get the PR up to get the conversation started on if this is upstreamable / acceptable style and so on. 

Thoughts etc?

### DIFF
--- a/build/lex.go
+++ b/build/lex.go
@@ -27,48 +27,34 @@ import (
 )
 
 // BuildFilenames is a collection of filenames in lowercase that are treated as BUILD files.
-var BuildFilenames = map[string]bool{
-	"build": true,
-	"build.bazel": true,
-	"workspace": true,
-	"workspace.bazel": true,
-	"stdin": true,
-}
-
-// ParseBuild parses a file, marks it as a BUILD file and returns the corresponding parse tree.
-//
-// The filename is used only for generating error messages.
-func ParseBuild(filename string, data []byte) (*File, error) {
-	in := newInput(filename, data)
-	f, err := in.parse()
-	if f != nil {
-		f.Build = true
-	}
-	return f, err
-}
-
-// ParseDefault parses a file, marks it as not a BUILD file (e.g. bzl file) and returns the corresponding parse tree.
-//
-// The filename is used only for generating error messages.
-func ParseDefault(filename string, data []byte) (*File, error) {
-	in := newInput(filename, data)
-	f, err := in.parse()
-	if f != nil {
-		f.Build = false
-	}
-	return f, err
+var BuildFilenames = map[string]FileType{
+	"buck":            BUCK,
+	"build":           BUILD,
+	"build.bazel":     BUILD,
+	"workspace":       BUILD,
+	"workspace.bazel": BUILD,
+	"stdin":           BUILD,
 }
 
 // Parse parses the input data and returns the corresponding parse tree.
-//
-// Uses the filename to detect the formatting type (either build or default) and calls
-// either ParseBuild or to ParseDefault correspondingly.
-func Parse(filename string, data []byte) (*File, error) {
-	basename := filepath.Base(filename)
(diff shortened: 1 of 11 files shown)


rmaz:
I signed it!

laurentlb:
@vladmos Can you take a look at this?

Ideally, we should try to remove differences between Bazel and Buck.

rmaz:
> Ideally, we should try to remove differences between Bazel and Buck

I definitely agree with this. We could add the shortened target name logic to buck instead, but for the unicode change I would be less keen. I don't understand the requirement for escaping it in this way, do you have context on this? These string values are inserted into Xcode project configs directly and read out by other tools, so we would have to add the unescaping logic in several places, which would be less than ideal.

ttsugriy:
@rmaz, buildifier works very well for Buck - it's used internally at Facebook for all of our repos without any issues.

> * Buck does not support the shortened name syntax of Bazel
It's true but buildifier has an argument to not shorten target names, but shortening can be disabled by passing `-buildifier_disable=label` to buildifier. The decision not to shorten was also [intentional](https://github.com/bazelbuild/buildtools/issues/34).
> * Buck does not handle escaped unicode in its BUCK files (unclear to me why Latin-1 is a requirement here, the skylark docs seem to mention unicode strings are fine)
There was a bug in Buck that has been fixed months ago and since then Buck handles unicode just fine.

burakcoskun:
great, then problem solved for short labels, but what about unicode escaping logic in buildifier? buildifier escapes unicode (non-latin 1) characters and BUCK doesn't understand that 

> Buck does not handle escaped unicode in its BUCK files (unclear to me why Latin-1 is a requirement here, the skylark docs seem to mention unicode strings are fine)

build/lex.go
```diff
@@ -27,50 +27,36 @@ import (
...
-// ParseBuild parses a file, marks it as a BUILD file and returns the corresponding parse tree.
-//
-// The filename is used only for generating error messages.
-func ParseBuild(filename string, data []byte) (*File, error) {
-	in := newInput(filename, data)
-	f, err := in.parse()
-	if f != nil {
-		f.Build = true
+// Parse parses the input data and returns the corresponding parse tree.
+func Parse(filename string, fileType FileType, data []byte) (*File, error) {
```
vladmos: Could you please not change the signature (at least for now)? There are existing projects that use buildifier as a library, and they have calls to `build.Parse(filename string, data []byte)`, such change will break all of them at once.

If you want to change the signature, consider adding alternative API (for example, a function called `ParseFile` with a new set of parameters) while keeping the existing functions for compatibility.

build/parse_test.go
```diff
@@ -32,7 +32,7 @@ func TestParse(t *testing.T) {
 		if tt.out != nil {
 			name = tt.out.Path
 		}
-		p, err := Parse(name, []byte(tt.in))
+		p, err := Parse(name, 0, []byte(tt.in))
```
vladmos: As I've mentioned, `build.Parse` is called from other projects, passing a magic number reduces readability. Perhaps it's better to create another method `ParseAuto` or a constant that's equal to 0.

rmaz:
I'll close this since it seems the label shortening is more easily disabled. 

> There was a bug in Buck that has been fixed months ago and since then Buck handles unicode just fine.

The issue we have here is not that buck doesn't handle unicode, its that non ascii characters are being escaped in such a way that external tools (eg Xcode) cannot read them any more. I don't see a solution for this right now, but will put up a separate PR to disable this in a similar manner to the label shortening.

PR #344, opened 2018-07-30
More portable bash invocation

amckague:
This helps with nixos where bash is installed elsewhere by default. (after fixing the cla of course...)


amckague:
I signed it!

PR #347, opened 2018-08-02
test case for shortenlabel

borkaehw:
# Issue
We got a problem back in April that `buildozer "remove deps @com_typesafe_play_twirl_parser_2_11//jar:jar"` removes `@com_typesafe_play_twirl_parser_2_11//jar:jar` but not `@com_typesafe_play_twirl_parser_2_11//jar`. We have our own solution on our fork but it seems to be fixed on upstream. It would be good to have test cases for this specific deps format.


pmbethe09:
LGTM

PR #350, opened 2018-08-09
Fix shortenlabel repo equals rule case

borkaehw:
# Issue
We are failing to shorten label when repo equals to rule, to be specific
`@foo//:foo` should be shortened to `@foo`
Also, we add test cases for this issue.


edit/edit_test.go
```diff
@@ -62,11 +62,14 @@ var shortenLabelTests = []struct {
...
 	{"//base:base", "devtools", "//base"},
 	{"//base", "base", ":base"},
+	{"//devtools/buildozer:buildozer", "", "//devtools/buildozer"},
 	{"@r//devtools/buildozer:buildozer", "devtools/buildozer", "@r//devtools/buildozer"},
 	{"@r//devtools/buildozer", "devtools/buildozer", "@r//devtools/buildozer"},
 	{"@r//devtools", "devtools", "@r//devtools"},
 	{"@r:rule", "", "@r:rule"},
 	{"@r", "", "@r"},
+	{"@foo//:foo", "", "@foo"},
+	{"@foo//devtools:foo", "", "@foo"},
```
pmbethe09: this looks wrong.  while the previous should shorten, here it should remain as @foo//devtools:foo and the package should be devtools

edit/edit.go
```diff
@@ -90,6 +90,9 @@ func ShortenLabel(label string, pkg string) string {
 		}
 		return "@" + repo + "//" + labelPkg
 	}
+	if strings.HasPrefix(label, "@") && repo == rule {
```
pmbethe09: based on your test, this looks incomplete

borkaehw:
Thanks for pointing the problem out, hopefully this makes more sense. I also added more test cases.

edit/edit.go
borkaehw: I've made some changes.

pmbethe09: it seems to me this simplifies as just
if  strings.HasPrefix(label, "@") && repo == rule && labelPkg == "" {
￼   return "@" + repo
￼ }

edit/edit_test.go
```diff
@@ -58,15 +58,21 @@ var shortenLabelTests = []struct {
...
 	{"//base:base", "devtools", "//base"},
 	{"//base", "base", ":base"},
+	{"//devtools/buildozer:buildozer", "", "//devtools/buildozer"},
 	{"@r//devtools/buildozer:buildozer", "devtools/buildozer", "@r//devtools/buildozer"},
 	{"@r//devtools/buildozer", "devtools/buildozer", "@r//devtools/buildozer"},
 	{"@r//devtools", "devtools", "@r//devtools"},
 	{"@r:rule", "", "@r:rule"},
 	{"@r", "", "@r"},
+	{"@foo//:foo", "", "@foo"},
+	{"@foo//devtools:foo", "", "@foo//devtools:foo"},
```
pmbethe09: this looks out of place, same LHS as the next case, but empty 'pkg' here looks wrong.

borkaehw: As far as I know, we are referring to external repo while we are in the current workspace root. For the next case, we are also referring to external repo but we want to make sure even though we have the same package name, we shouldn't shorten it anyway.

Could you elaborate more what looks wrong to you? Thanks.

edit/edit.go
borkaehw: This makes sense, thanks for feedback again.

edit/edit_test.go
pmbethe09: only one of
{"@foo//devtools:foo", "", "@foo//devtools:foo"},
{"@foo//devtools:foo", "devtools", "@foo//devtools:foo"},

should work, as with the same input you expect the same output -- these are 2 tests with the same input but 2 different outputs.

borkaehw:
@pmbethe09 I would like to know if there is other improvement we should make to get this merged?

pmbethe09:
LGTM (i was away).  Also hecking with @vladmos

PR #355, opened 2018-08-17
Fix PackageDeclaration() - package() is added after load() statements.

laurentlb:
https://docs.bazel.build/versions/master/skylark/build-style.html#file-structure

### DIFF
--- a/edit/edit.go
+++ b/edit/edit.go
@@ -183,8 +183,8 @@ func ExistingPackageDeclaration(f *build.File) *build.Rule {
 }
 
 // PackageDeclaration returns the package declaration. If it doesn't
-// exist, it is created at the top of the BUILD file, after leading
-// comments / load statements.
+// exist, it is created at the top of the BUILD file, after optional
+// docstring, comments, and load statements.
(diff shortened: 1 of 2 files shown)


edit/edit.go
```diff
@@ -184,20 +184,23 @@ func ExistingPackageDeclaration(f *build.File) *build.Rule {
...
 	added := false
 	call := &build.CallExpr{X: &build.Ident{Name: "package"}}
-	// Skip CommentBlocks and find a place to insert the package declaration.
 	for _, stmt := range f.Stmt {
-		_, ok := stmt.(*build.CommentBlock)
-		if !ok && !added {
-			all = append(all, call)
-			added = true
+		switch stmt.(type) {
+		case *build.CommentBlock, *build.LoadStmt:
```
vladmos: Perhaps it should also be added after a docstring if it exists? Not sure what's the formal definition (e.g. whether it must be the first statement in the file or can follow several comment blocks).

laurentlb: Done (a string literal at top-level is a good approximation)

Issue #349, opened 2018-08-07
-types java_test,cc_test,sh_test,py_test not enforced 


uriklagnes:
I'm trying to add `tags = ["block-network"]` to every ***_test** rule in a bazel build file.  
My invocation is `buildozer -types java_test,cc_test,sh_test,py_test "set tags block-network" //pkg_name:*`  
  
But what I see happening is that the 'package' rule is also getting this tag added.  
  
Before:
```
package(default_visibility = ["//visibility:public"])
  
py_binary(
    name = "compile",
    srcs = ["compile.py"],
)

py_test(
    name = "compile_test",
    srcs = ["compile_test.py"],
    deps = [
        ":compile",
    ],
    tags = ["block-network"],
)
```

After:
```
package(
    default_visibility = ["//visibility:public"],
    tags = ["block-network"],
)
  
py_binary(
    name = "compile",
    srcs = ["compile.py"],
)

py_test(
    name = "compile_test",
    srcs = ["compile_test.py"],
    deps = [
        ":compile",
    ],
    tags = ["block-network"],
)
```

**VERSIONS:**
buildozer:  Downloaded latest on master branch today
bazel: Build label: 0.16.0

pmbethe09:
I think that may be from
https://github.com/bazelbuild/buildtools/blob/master/edit/buildozer.go#L542

I am not sure why it is there, but it has been there since the open-sourcing.

uriklagnes:
Should I create a pull request to delete that?

pmbethe09:
sure.  create a PR.  meanwhile I am checking with other authors if they know why that is there.

PR #356, opened 2018-08-17
Remove targetting package rules implicitly

uriklagnes:
See issue #349


uriklagnes:
I signed it!

pmbethe09:
LGTM, but it appears there is still a CLA problem.

PR #359, opened 2018-08-26
Disable label rewrite if shortenLabelsFlag is false

burakcoskun:
shortenLabelsFlas is being used in edit ShortenLabels function
but not used in rewrite ShortenLabels, which causes all the labels to be
shorten though this flag is set to false.


burakcoskun:
I signed it!

pmbethe09:
LGTM

PR #366, opened 2018-08-29
Also detect LiteralExprs as UsedSymbols.

varomodt:
The unusedLoads fixer doesn't work for some BZL files b/c it removes used
symbols which contain a constant (e.g. my_import.CONSTANT) since UsedSymbols
only detects *build.Ident. This CL adds detection of potential imported
symbols from LiteralExprs to UsedSymbols.

Internal bug: b/110991860

### DIFF
--- a/edit/edit.go
+++ b/edit/edit.go
@@ -741,11 +741,7 @@ func EditFunction(v build.Expr, name string, f func(x *build.CallExpr, stk []bui
 func UsedSymbols(f *build.File) map[string]bool {
 	symbols := make(map[string]bool)
 	build.Walk(f, func(expr build.Expr, stack []build.Expr) {
-		literal, ok := expr.(*build.Ident)
-		if !ok {
-			return
-		}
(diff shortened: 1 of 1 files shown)


laurentlb:
`my_import.CONSTANT` should be parsed as a `DotExpr` (https://github.com/bazelbuild/buildtools/blob/master/build/syntax.go#L189). `LiteralExprs` are not supposed to include dots.

varomodt:
Well, I see them as LiteralExprs experimentally: Were it a DotExpr, we would recurse in and see an Ident I think. Maybe that should be a separate issue?

laurentlb:
Parser looks correct to me: https://github.com/bazelbuild/buildtools/blob/master/build/parse.y#L433

Either there's a bug, or something else modifies the AST. Can you double-check?

varomodt:
Retracting this PR, as it was a result of an error in internal code.

PR #371, opened 2018-09-10
Fix spans for various nodes

vladmos:
Span() is support to return a semi-inclusive interval, although for some nodes it return the position of the last byte instead of the next byte.


vladmos:
I'll probably add the tests (although if we switch to another parser soon that won't matter).

PR #372, opened 2018-09-10
Add manual tag to the constructed buildifier target

ash2k:
Today if the buildifier rule user runs `bazel test //...` on their project, buildifier-wrapping bash script will be built along with buildifier itself and everything it depends upon (go sdk, etc).
This change wraps the rule invocation in a macro to add the "manual" tag to it so that it is not matched by wildcard target patterns.
This is the same technic that is used in gazelle https://github.com/bazelbuild/bazel-gazelle/blob/82d15888a19178195d56d9a288a2653d16436b07/def.bzl#L116

### DIFF
--- a/buildifier/def.bzl
+++ b/buildifier/def.bzl
@@ -33,7 +33,7 @@ def _buildifier_impl(ctx):
         executable = out_file,
     )]
 
-buildifier = rule(
+_buildifier = rule(
     implementation = _buildifier_impl,
     attrs = {
         "verbose": attr.bool(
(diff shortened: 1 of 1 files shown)


buildifier/def.bzl
```diff
@@ -52,15 +52,23 @@ buildifier = rule(
             doc = "A list of glob patterns passed to the find command. E.g. './vendor/*' to exclude the Go vendor directory",
         ),
         "_buildifier": attr.label(
-            default = Label("@com_github_bazelbuild_buildtools//buildifier"),
```
ash2k: `attr.label()` converts strings to labels by itself.

buildifier/def.bzl
```diff
@@ -52,15 +52,23 @@ buildifier = rule(
...
+            default = "@com_github_bazelbuild_buildtools//buildifier",
             cfg = "host",
-            allow_single_file = True,
```
ash2k: Private attribute and the value is a label. No need to allow to specify a file.

ash2k:
@vladmos maybe you could have a look at this PR? (pinging the most active contributor)

ash2k:
Thanks!

Issue #126, opened 2017-08-14
Buildozer: Add "key: value" to dict


MarkusTeufelberger:
I couldn't find any way to add a key/value pair to a dictionary attribute. Am I missing something or is this functionality simply not there?

MarkusTeufelberger:
Which keyword would you prefer?

Maybe something like "dict_add"/"dict_replace"/"dict_move" similar to the list functionality?

pmbethe09:
@laurentlb -- thoughts?

laurentlb:
This can be useful, indeed.
Can describe more precisely the syntax you suggest?

MarkusTeufelberger:
The current functionality of `add`:
> `add <attr> <value(s)>`: Adds value(s) to a list attribute of a rule. If a value is already present in the list, it is not added.

How I would expect `dict_add` to work:

> `dict_add <attr> <key> <value>`: Adds a value at a certain key to a dict attribute of a rule. If a value is already present at this key in the dict, it is not added.

Currently I'm creating dict attributes by just using `set` and entering a dict in there, which works but isn't pretty. It also makes it awkward to edit dicts.

laurentlb:
Great, looks good!

PaulSonOfLars:
Any news on this?

MarkusTeufelberger:
Not from my side.

PaulSonOfLars:
Great, I'll open a PR for this in a bit then! `dict_add`, `dict_set`, and `dict_remove` should cover all dict essentials?

bpfoley:
Shouldn't this issue be closed now? It was implemented by PR#421.

alexeagle:
@vladmos ☝🏻

PR #376, opened 2018-09-17
Add dict manipulation tools.

PaulSonOfLars:
This address issue #126, adding dictionary manipulation tools.

Specifically, this adds the following to buildozer:

`dict_add <attr> <(key:value)(s)>` Which does not overwrite existing values.
`dict_set <attr> <(key:value)(s)>` Which overwrites existing values.
`dict_remove <attr> <key(s)>` Which removes keys from the dictionary, removing the final dictionary if it is now empty.

### DIFF
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -454,6 +454,78 @@ func cmdCopyNoOverwrite(opts *Options, env CmdEnvironment) (*build.File, error) 
 	return copyAttributeBetweenRules(env, attrName, from)
 }
 
+// cmdDictAdd adds a key to a dict, if that key does _not_ exit already.
+func cmdDictAdd(opts *Options, env CmdEnvironment) (*build.File, error) {
+	attr := env.Args[0]
+	args := env.Args[1:]
+
+	dict := &build.DictExpr{}
+	currDict, ok := env.Rule.Attr(attr).(*build.DictExpr)
+	if ok {
+		dict = currDict
+	}
+
+	for _, x := range args {
+		kv := strings.Split(x, ":")
+		expr := build.StringExpr{
+			Value:       kv[1],
+			TripleQuote: false,
+			Token:       "",
+		}
+
+		prev := DictionaryGet(dict, kv[0])
+		if prev == nil {
+			// Only set the value if the value is currently unset.
+			DictionarySet(dict, kv[0], &expr)
+		}
+		env.Rule.SetAttr(attr, dict)
+	}
+	return env.File, nil
+}
+
+// cmdDictSet adds a key to a dict, overwriting any previous values.
+func cmdDictSet(opts *Options, env CmdEnvironment) (*build.File, error) {
+	attr := env.Args[0]
+	args := env.Args[1:]
+
+	dict := &build.DictExpr{}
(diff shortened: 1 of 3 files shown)


petemounce:
cc @devnev re: block-network?

edit/edit.go
```diff
@@ -682,6 +682,18 @@ func DictionarySet(dict *build.DictExpr, key string, value build.Expr) build.Exp
 	return nil
 }
 
+// DictionaryGet looks for the key in the dictionary expression, and returns the
+// current value. If it is unset, it returns nil.
+func DictionaryGet(dict *build.DictExpr, key string) build.Expr {
+	for _, e := range dict.List {
+		kv, _ := e.(*build.KeyValueExpr)
```
pmbethe09: while it should be a KeyValueExpr, might as well inspect and continue if not

kv, ok := e.(*build.KeyValueExpr)
if !ok {
  continue
}

edit/buildozer.go
```diff
@@ -454,6 +455,85 @@ func cmdCopyNoOverwrite(opts *Options, env CmdEnvironment) (*build.File, error)
...
+}
+
+// cmdDictRemove removes a key from a dict.
+func cmdDictRemove(opts *Options, env CmdEnvironment) (*build.File, error) {
+	attr := env.Args[0]
+	args := env.Args[1:]
+
+	thing := env.Rule.Attr(attr)
+	dictAttr, ok := thing.(*build.DictExpr)
+	if ok {
```
pmbethe09: if !ok, then this isn't a dict, so probably cleaner to invert-return
if !ok {
  return env.File, nil
}
for _, x := range args {

edit/buildozer.go
```diff
@@ -454,6 +455,85 @@ func cmdCopyNoOverwrite(opts *Options, env CmdEnvironment) (*build.File, error)
...
+			// should errors here be flagged?
+			DictionaryDelete(dictAttr, x)
+			env.Rule.SetAttr(attr, dictAttr)
+		}
+	}
+
+	// If the removal results in the dict having no contents, delete the attribute (stay clean!)
+	if dictAttr == nil || len(dictAttr.List) == 0 {
+		if env.Rule.DelAttr(attr) != nil {
+			return env.File, nil
```
pmbethe09: this seems like clutter
if dictAttr == nil || len(dictAttr.List) == 0 {
  env.Rule.DelAttr(attr)
}
return env.File, nil

PaulSonOfLars:
We need a way of calling function/methods from inside dicts; would it be acceptable to use a `dict_add_func` method for adding function calls to dictionaries? or is there some syntactic sugar already present to denote "this is a function" vs "this is a string"?

petemounce:
Hi there - @PaulSonOfLars has moved on from improbable.io. I still work there. I notice the CLA check fails, but I'm pretty sure we're covered by a company-wide CLA. Is there anything I can do to unblock and get this merged?

laurentlb:
@petemounce Can you create a new PR?

petemounce:
@laurentlb - sure, done in #421.

PaulSonOfLars:
Closing, since this was merged with #421. Big thanks to @petemounce taking over the CLA for me!

PR #378, opened 2018-09-24
Correctly interpret call context with DotExprs.

varomodt:
This upstreams an internal change which corrects the Buildifier's
rewrite behavior when the call site is a DotExpr as opposed to an Ident.

### DIFF
--- a/build/rewrite.go
+++ b/build/rewrite.go
@@ -23,7 +23,7 @@ import (
 	"sort"
 	"strings"
 
-	"google3/devtools/buildifier/tables"
+	"github.com/bazelbuild/buildtools/tables"
 )
 
 // For debugging: flag to disable certain rewrites.
--- a/build/rewrite.go
+++ b/build/rewrite.go
@@ -340,6 +340,7 @@ func callName(call *CallExpr) string {
 		case *DotExpr:
 			elems = append([]string{fun.(*DotExpr).Name}, elems...)
 		default:
+			return ""
 		}
 	}
 }
--- a/build/rewrite.go
+++ b/build/rewrite.go
@@ -23,7 +23,7 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/bazelbuild/buildtools/tables"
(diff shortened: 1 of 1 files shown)


build/rewrite.go
```diff
@@ -23,7 +23,7 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/bazelbuild/buildtools/tables"
```
laurentlb: please revert this change

build/rewrite.go
```diff
@@ -460,12 +467,21 @@ func sortStringLists(f *File, info *RewriteInfo) {
...
+				// this context is in the blacklist.
+				if disabled("unsafesort") || !f.Build || tables.SortableBlacklist[context] {
 					continue
 				}
-				if disabled("unsafesort") && !tables.SortableWhitelist[context] && !allowedSort(context) {
+
+				// Drop out unless this field name is whitelisted, we're explicitly
+				// allowed to sort it for debugging reasons, or its field name is
+				// explicitly allowed.
+				if !tables.SortableWhitelist[context] && !allowedSort(context) && !tables.IsSortableListArg[key.Name] {
```
laurentlb: For clarification: is there any behavior change here?

build/rewrite.go
varomodt: Done. Sorry about that -- I'm not used to github.

build/rewrite.go
varomodt: There are some minor behavior changes here:

- If "unsafesort" is disabled, then this makes this rewrite _never_ apply, whereas at head it will still apply if allowedSort returns true or the field is in SortableWhitelist, which seemed inconsistent with the semantics implied by the "disabled" method. Note that every other case here already drops out immediately if unsafesort is disabled.

- This makes SortableWhitelist and allowedSort allow a field to be sorted even if the given key.Name is not in IsSortableListArg: the current behavior seems to violate the implicit semantics of an explicit SortableWhitelist.

PR #390, opened 2018-10-08
Fix the anchors

vladmos:
Github-flavored markup only supports html anchors


vladmos:
@laurentlb please merge the PR, because I don't have the permissions now.

PR #391, opened 2018-10-09
Enable linter functionality for buildifier

vladmos:


warn/bzlenv.go
```diff
@@ -0,0 +1,218 @@
+// Package bzlenv provides function to create and update a static environment.
+package bzlenv
```
jmillikin-stripe: FYI: adding a second package to the `warn/` directory prevents the basic Go toolchain from building buildifier.

vladmos: Thanks for pointing out, I'll move it into the `warn` package.

PR #393, opened 2018-10-11
Move bzlenv to a separate directory

vladmos:
As suggested at https://github.com/bazelbuild/buildtools/pull/391#discussion_r224268280, refactoring buildtools so that every package is located in a separate directory.


pmbethe09:
LGTM

PR #399, opened 2018-10-22
Don't print warnigns for fixed findings

vladmos:


vladmos:
Isn't that a bug that a warning without a suggested fix is actionable?

laurentlb:
Normally, non-actionable is used to show a message to the user that is non-actionable (user doesn't need to change anything).
I think all warnings should be marked as actionable (the feature was added to handle a special case inside Google). I don't know if we still need that.

vladmos:
The flag is used by buildifier in the `--lint=fix` mode to show whether a warning has been fixed automatically, and currently it shows "fixed" for all warnings, which is not necessarily true. Should I add another flag (`Fixed bool`) and use it instead?

laurentlb:
Do we need to print what was fixed?

vladmos:
Reverted, changed to not print warnings for fixed findings.

PR #400, opened 2018-10-24
Expand the set of file names considered to be build files.

benjaminp:
When using Bazel external workspaces, it's common to have files named like something.BUILD or BUILD.something, which become the BUILD file for the `@something` repository. This CL modifies build.ParseDefault so that it considers any filename with an extension or base called "BUILD" to be a BUILD file. While workspace file naming in the wild is less creative, I didn't see a reason not to extend the same rule to workspace files.

The build.BuildFilenames variable is public, so I preserved its behavior as an override. If it's not actually considered a public API, I can delete it.

### DIFF
--- a/build/BUILD.bazel
+++ b/build/BUILD.bazel
@@ -29,6 +29,7 @@ go_test(
     size = "small",
     srcs = [
         "checkfile_test.go",
+        "lex_test.go",
         "parse_test.go",
         "print_test.go",
         "quote_test.go",
@@ -44,6 +45,7 @@ go_test(
     ],
     embed = [":go_default_library"],
     deps = [
+        "//tables:go_default_library",
         "//testutils",
     ],
 )
--- a/build/lex.go
+++ b/build/lex.go
@@ -28,11 +28,7 @@ import (
 
 // BuildFilenames is a collection of filenames in lowercase that are treated as BUILD files.
 var BuildFilenames = map[string]bool{
-	"build":           true,
-	"build.bazel":     true,
-	"workspace":       true,
-	"workspace.bazel": true,
-	"stdin":           true,
+	"stdin": true,
 }
 
 // ParseBuild parses a file, marks it as a BUILD file and returns the corresponding parse tree.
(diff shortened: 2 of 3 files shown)


pmbethe09:
you appear to need to rebase

build/lex.go
```diff
@@ -59,13 +56,23 @@ func ParseDefault(filename string, data []byte) (*File, error) {
...
 }
 
+var buildRegexp = regexp.MustCompile(`(?:^|\.)(?:build|workspace)(?:$|\.)`)
+
+func isBuildFilename(basename string) bool {
+	basename = strings.ToLower(basename)
+	if isBuild, ok := BuildFilenames[basename]; ok {
+		return isBuild
+	}
+	return buildRegexp.MatchString(basename)
```
vladmos: This will return true for `build.bzl` and `workspace.bzl` files. For example:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/workspace.bzl
https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/docker/build.bzl

benjaminp: That's a good point.

Is there a reason this matching is case-insensitive? I actually do have a bunch of BUILD-like files named `BUILD.bzl`, which I wanted to be picked up.

vladmos: It's case-insensitive because of case-insensitive file systems (e.g. on Windows or Mac).

build/BUILD.bazel
```diff
@@ -44,6 +45,7 @@ go_test(
     ],
     embed = [":go_default_library"],
     deps = [
+        "//tables:go_default_library",
```
vladmos: Is it required? I don't see any usages of `tables` in the pull reqest.

build/lex.go
benjaminp: Those are case-insensitive but case-preserving, right? So, I would expect to be getting `BUILD.whatever` (in that casing) back from the filesystem.

build/BUILD.bazel
benjaminp: This is just from running gazel. I can manually remove it if it's a problem.

build/lex.go
benjaminp: I've updated the PR to explicitly exclude `.bzl` and .sky` extensions from being treated as BUILD files.

vladmos:
Thanks!

keith:
A quick follow up to this https://github.com/bazelbuild/buildtools/pull/437

PR #401, opened 2018-10-26
A warning a fix for load statements not on top

vladmos:


warn/warn.go
```diff
@@ -258,6 +258,41 @@ func packageOnTopWarning(f *build.File, fix bool) []*Finding {
...
+	findings := []*Finding{}
+	firstStmtIndex := -1 // index of the first seen non-load statement
+	for i := 0; i < len(f.Stmt); i++ {
+		stmt := f.Stmt[i]
+		_, isString := stmt.(*build.StringExpr) // typically a docstring
+		_, isComment := stmt.(*build.CommentBlock)
+		if isString || isComment {
+			continue
+		}
+		if load, ok := stmt.(*build.LoadStmt); ok {
```
pmbethe09: I was too slow.  as always I suggest invert/continue to reduce indent.  So here
load, ok := stmt.(*build.LoadStmt)
if !ok {
  if firstStmtIndex == -1 {
    firstStmtIndex = i
  }
  continue
}
if firstStmtIndex == -1 {
  continue
}
if !fix {
  start, end := load.Span()
  findings = append(findings, makeFinding(f, start, end, "load-on-top",
				"Load statements should be at the top of the file.", true, nil))
 continue
}
... finally the main code ...

vladmos: Thanks, I'll update the code in a follow-up pull request.

PR #411, opened 2018-11-05
Update console output for linter messages

vladmos:
* `--lint=fix` now sorts warnings by line number for every file
  * `--lint=warn` doesn't show unfixed warnings (but shows their count)


vladmos:
This is the same as #410 which can't be submitted because of a stalled CLA check.

PR #413, opened 2018-11-07
Implement type inference

vladmos:


vladmos:
This is required to implement warnings and fixes that are specific to certain data types (e.g. not iterable strings and depsets, deprecated `+` for dicts). That can also be applied to bazel-specific data structures such as [args](https://github.com/bazelbuild/rules_go/issues/511): I'll need to detect that `ctx.actions.args()` return an `Args` object for which some usages of the method `.add` are deprecated.

warn/types.go
```diff
@@ -0,0 +1,146 @@
...
+				// preserve the type
+				if t, ok := result[node.X]; ok {
+					nodeType = t
+				} else if t, ok := result[node.Y]; ok {
+					nodeType = t
+				}
+			}
+		}
+
+		if nodeType != Unknown {
```
pmbethe09: opt: make this a defer right after the declaration, then in the switch, use return instead of break

warn/types_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	"github.com/bazelbuild/buildtools/build"
+	"github.com/bazelbuild/buildtools/testutils"
+)
+
+func checkTypes(t *testing.T, input, output string) {
+	input = strings.TrimLeft(input, "\n")
+	f, err := build.Parse("test.bzl", []byte(input))
+	if err != nil {
+		panic(fmt.Sprintf("%v", err))
```
pmbethe09: don't panic

t.Fatalf

warn/types_test.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	types := detectTypes(f)
+
+	var edit func(expr build.Expr, stack []build.Expr) build.Expr
+	edit = func(expr build.Expr, stack []build.Expr) build.Expr {
+		if t, ok := types[expr]; ok {
```
pmbethe09: invert
if !ok {
  return nil
}
build.EditChildren(...

build/parse.y
```diff
@@ -549,13 +549,13 @@ primary_expr:
 			ForceMultiLine: forceMultiLine($1, $2, $3),
 		}
 	}
-|	'{' tests_opt '}'  // TODO: remove, not supported
+|	'{' tests comma_opt '}'  // TODO: remove, not supported
```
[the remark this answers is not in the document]
vladmos: This PR doesn't make it supported, it just fixes a bug when empty curly braces were parsed as a set.

Sets are still not supported in Bazel, but may exist in other Skylark implementations, I'm not sure if we should keep them here or not.

warn/types.go
vladmos: Done.

warn/types_test.go
vladmos: Done.

warn/types_test.go
vladmos: Done.

PR #416, opened 2018-11-09
Implement a warning for the deprecated string iteration

vladmos:


warn/warn_test.go
```diff
@@ -413,3 +413,34 @@ d += foo + bar
...
+
+max(s)
+min(s)
+all(s)
+any(s)
+reversed(s)
+zip(s, a, b)
+zip(a, s)
+
+[foo(x) for x in s]
```
laurentlb: nit. add also a test that doesn't show a warning (e.g. argument is a list)

vladmos:
@laurentlb Please approve again, approvals on Github expire after every additional commit.

Issue #396, opened 2018-10-15
unused_deps not working


mkilpatrick:
unused_deps stopped working for us about 6 months ago. When we run it, it tries to open files that don't exist. 

```
mkilpatrick-mpro:alpha mkilpatrick$ unused_deps src/com/yext/bulk/common/...
2018/10/15 11:43:37 running: bazel query --tool_tag=unused_deps --keep_going --color=yes --curses=yes kind('(java|android)_*', src/com/yext/bulk/common/...)
2018/10/15 11:43:37 running: bazel build --tool_tag=unused_deps --keep_going --color=yes --curses=yes src/com/yext/bulk/common/...
Using remote cache (va). Disable by setting BAZEL_NO_CACHE=1
WARNING: Option 'experimental_remote_spawn_cache' is deprecated
INFO: Analysed 57 targets (0 packages loaded).
INFO: Found 57 targets...
INFO: Elapsed time: 0.780s, Critical Path: 0.01s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action

2018/10/15 11:43:39 open /private/var/tmp/_bazel_mkilpatrick/ecc7778653efb77408ccae0a11894b62/execroot/yext/bazel-out/darwin-fastbuild/bin/src/com/yext/bulk/common/fieldnames/fieldidentifier/canonicalizer/canonicalizer.jar-2.params: no such file or directory
2018/10/15 11:43:39 open /private/var/tmp/_bazel_mkilpatrick/ecc7778653efb77408ccae0a11894b62/execroot/yext/bazel-out/darwin-fastbuild/bin/src/com/yext/bulk/common/fieldnames/fieldidentifier/uuid/uuid.jar-2.params: no such file or directory
2018/10/15 11:43:39 open /private/var/tmp/_bazel_mkilpatrick/ecc7778653efb77408ccae0a11894b62/execroot/yext/bazel-out/darwin-fastbuild/bin/src/com/yext/bulk/common/fieldnames/fielddefinition/profilefieldnode/profilefieldnode.jar-2.params: no such file or directory
2018/10/15 11:43:39 no deps attribute for protobuf in /Users/mkilpatrick/alpha/src/com/yext/bulk/common/fieldnames/fieldidentifier/BUILD
[...]

PR #420, opened 2018-11-09
Fix unused_deps' params file handling

cushon:
This works around the following changes to params files:
https://groups.google.com/d/topic/bazel-discuss/oMguVWL-Ckg/discussion

The fix uses an apsect to extra the argument list of JavaBuilder
actions, instead of relying on params files being available at
well-known locations.

See #396


pmbethe09:
LGTM -- but needs a merge from master

PR #421, opened 2018-11-10
Add dictionary handling

petemounce:
(This supercedes #376 to deal with CLA; all credit to @PaulSonOfLars)

This address issue #126, adding dictionary manipulation tools.

Specifically, this adds the following to buildozer:

`dict_add <attr> <(key:value)(s)>` Which does not overwrite existing values.
`dict_set <attr> <(key:value)(s)>` Which overwrites existing values.
`dict_remove <attr> <key(s)>` Which removes keys from the dictionary, removing the final dictionary if it is now empty.


vladmos:
Thanks!

PR #427, opened 2018-11-13
Fix for the order of function arguments

vladmos:


warn/warn.go
```diff
@@ -640,6 +640,52 @@ func depsetIterationWarning(f *build.File, fix bool) []*Finding {
 	return findings
 }
 
+func argumentsOrderWarning(f *build.File, fix bool) []*Finding {
+	argumentType := func(expr build.Expr) int {
```
pmbethe09: this doesn't appear to need 'f'f or 'fix', so why an inline function rather than standalone?

warn/warn.go
```diff
@@ -640,6 +640,52 @@ func depsetIterationWarning(f *build.File, fix bool) []*Finding {
...
+			}
+		case *build.BinaryExpr:
+			if expr.Op == "=" {
+				return 2
+			}
+		}
+		return 1
+	}
+
+	getComparator := func(args []build.Expr) func(i, j int) bool {
```
pmbethe09: same

warn/warn.go
```diff
@@ -640,6 +640,52 @@ func depsetIterationWarning(f *build.File, fix bool) []*Finding {
...
+			sort.SliceStable(call.List, comparator)
+			return
+		}
+		if sort.SliceIsSorted(call.List, comparator) {
+			return
+		}
+		start, end := expr.Span()
+		findings = append(findings,
+			makeFinding(f, start, end, "args-order",
+				"Function call arguments should be in the following order: positional, keyword, *args, **kwargs.", true, nil))
```
pmbethe09: these indents are distracting.  consider making the error message a file-constant so that all of this cleanly fits on one line)

warn/warn.go
vladmos: This file as well as `warn_bazel.go` has multiple independent warn/fix functions, IMO it looks better when the utility functions that are needed just for one of them are defined inside them, otherwise it's not clear which function belongs to which fix.

warn/warn.go
vladmos: I'm not happy about this code either, in each warning function I essentially need to copy the following boilerplate:

    findings := []*Finding{}
    ...
    start, end := ....Span()
    findings = append(findings, makeFinding(f, start, end, "<category>", "<warning text>", true, nil)).

I may make the last two parameters different for some warnings in the future, but right now only the category name and the warning message are different for each function, and the category name can be taken from `FileWarningMap`. Any idea how I can get rid of the boilerplate code? Ideally I'd like to just call something like `newWarning(expr, "<warning text>")`, should I pass such function to each warning function as a callback?

PR #428, opened 2018-11-14
Refactor the code

vladmos:
This CL addresses the code review comments.

### DIFF
--- a/warn/warn_bazel.go
+++ b/warn/warn_bazel.go
@@ -1,6 +1,7 @@
 package warn
 
 import (
+	"fmt"
 	"github.com/bazelbuild/buildtools/build"
 	"github.com/bazelbuild/buildtools/bzlenv"
 	"github.com/bazelbuild/buildtools/edit"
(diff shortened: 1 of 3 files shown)


warn/warn_bazel.go
```diff
@@ -87,7 +88,7 @@ func globalVariableUsageCheck(f *build.File, category, global, alternative strin
 		start, end := ident.Span()
 		findings = append(findings,
 			makeFinding(f, start, end, category,
-				"Global variable \""+global+"\" is deprecated in favor of \""+alternative+"\". Please rename it.", true, nil))
+				fmt.Sprintf(`Global variable "%s" is deprecated in favor of "%s". Please rename it.`, global, alternative), true, nil))
```
pmbethe09: alternate option -- allow makeFinding to take a format-string and a var-args list, do the fmt.Sprintf in there.

vladmos: Thanks for the idea, I'll consider doing this when I refactor the warnings, currently I need to pass too many arguments to `makeFinding`, it may be confusing that some of them are for the string formatting and some are not.

PR #429, opened 2018-11-14
[buildifier] Warn and fix loads from the same file.

ttsugriy:
Loads like
```
load(":f.bzl", "s1")
load(":f.bzl", "s2")
```
can be simplified as
```
load(":f.bzl", "s1", "s2")
```
which is more concise and less error prone.

### DIFF
--- a/WARNINGS.md
+++ b/WARNINGS.md
@@ -141,6 +141,34 @@ or at the beginning of a rule.
 
 --------------------------------------------------------------------------------
 
+## <a name="same-origin-load"></a>Same label is used for multiple loads
+
+  * Category_name: `same-origin-load`
+  * Automatic fix: yes
+
+### Background
+
+[load](https://docs.bazel.build/versions/master/skylark/concepts.html#loading-an-extension)
+is used to import definitions in a BUILD file. If the same label is used for loading
+symbols more the ones, all such loads can be merged into a single one.
+
+### How to fix it
+
+Merge all loads into a single one. For example,
+
+```
+load(":f.bzl", "s1")
+load(":f.bzl", "s2")
+```
+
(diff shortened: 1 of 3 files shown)


warn/warn.go
```diff
@@ -663,6 +693,7 @@ var FileWarningMap = map[string]func(f *build.File, fix bool) []*Finding{
 	"http-archive":       nativeHTTPArchiveWarning,
 	"integer-division":   integerDivisionWarning,
 	"load":               unusedLoadWarning,
+	"same-origin-load":   sameOriginLoadWarning,
```
pmbethe09: preserve alphabetical here -- so push down to the natural spot

vladmos:
Please also add the warning description to the WARNINGS.md file.

ttsugriy:
> Please also add the warning description to the WARNINGS.md file.

@vladmos, I actually had a warning description initially, but I removed it because I somehow got under impression that those were for buildozer fix instead of buildifier, since there is only one reference to buildifier and a lot of reference including command examples for buildozer :) Now that you've removed the confusion for me, I readded the warning description.

WARNINGS.md
```diff
@@ -141,6 +141,34 @@ or at the beginning of a rule.
 
 --------------------------------------------------------------------------------
 
+## <a name="load"></a>Same label is used for multiple loads
```
vladmos: Please rename to `same_origin_load`

ttsugriy: sorry about that and thanks for the catch!

vladmos:
Thanks!

ttsugriy:
thank you for the supersonic review, @vladmos!

PR #430, opened 2018-11-14
suppress output of goyacc by writing to /dev/null

pmbethe09:


pmbethe09:
addresses 
https://github.com/bazelbuild/buildtools/issues/280#issuecomment-436705613
while I work on actually fixing the grammar

PR #433, opened 2018-11-16
[buildifier] Preserve symbol semantics when removing unused loads.

ttsugriy:
When the same symbol is loaded multiple times, during evalution the
last symbol wins and all previous loads are ignored, but `buildifier`
would preserve the first instance and remove all other usages
which breaks semantics unless all loads are from the same file.

In other words

```
load(":defs1.bzl", "s1")
load(":defs2.bzl", "s1")
```

is equivalent to

```
load(":defs2.bzl", "s1")
```

but `buildifier` would produce

```
load(":defs1.bzl", "s1")
```

To address this, all statements are analyzed in a reverse order, so the very
last (in terms of line numbers) is preserved and all preceding loads are
removed. This also slightly simplifies `load` removal logic, since we no longer
need to adjust index.

### DIFF
--- a/warn/warn.go
+++ b/warn/warn.go
@@ -120,7 +120,10 @@ func unusedLoadWarning(f *build.File, fix bool) []*Finding {
 	loaded := make(map[string]bool)
 
 	symbols := edit.UsedSymbols(f)
-	for stmtIndex := 0; stmtIndex < len(f.Stmt); stmtIndex++ {
+	// statements are considered in reserve order to make sure that the last
+	// load wins, just like what happens during evaluation, which is very
+	// important if same symbol is loaded from different files.
+	for stmtIndex := len(f.Stmt) - 1; stmtIndex >= 0; stmtIndex-- {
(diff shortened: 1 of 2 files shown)


laurentlb:
The new behavior is not perfect either (what if the symbol is used between the two loads?).

But the interpreter should be stricter in the future: we'll forbid symbol redefinition in that case (as in bzl files, you shouldn't be able define twice the same symbol at top-level).

vladmos:
Please also add a more complex test case, where both `load`s don't need to be removed entirely, e.g.:

    load("foo", a, b, c)
    load("bar", a, d, e)

    a + b + d

Should be fixed as

    load("foo", b)
    load("bar", a, d)

    a + b + d

ttsugriy:
@laurentlb, you're right that this behavior is not 100% correct too, but fortunately it plays well with another lint rule - `load-on-top` - https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#load-must-appear-at-top-of-file which should make sure that no symbol is used until all loads happen :)

PR #440, opened 2018-11-20
[buildifier] Warn about and fix non-lexicographic load order.

ttsugriy:
This makes it easier to developers to locate loads, but more
importantly reduces changes for large-scale automated refactorings
to result in merge conflicts, since they usually add loads at the
end or some other fixed position.


ttsugriy:
That is an awesome point, @vladmos! I have special-cased `@` and applied your great stylistic suggestions.

PR #444, opened 2018-11-21
Only print lint fixes with verbose mode

keith:
Previously every file that was processed by the buildifier -lint=fix
command would produce output, even if there weren't any fixes. To make
this more like the buildifier fix formatting mode, this wraps that print
in the `-v` flag.


keith:
Updated!

vladmos:
Thanks!

PR #459, opened 2018-11-26
Fix the out-of-order-load check

### LINKED ISSUE #457: buildifier lint infinite fix loop
Currently if I run buildifier's lint through bazel with this rule:

```python
buildifier(
    name = "buildifier",
    lint_mode = "fix",
    mode = "fix",
)
```

and `bazel run //:buildifier`

It fixes this starting file:

```python
load(
    "//bazel:api.bzl",
    "static_library",
    "swiftpm_test",
)
load(
    "@build_bazel_rules_swift//swift:swift.bzl",
    "swift_binary",
)

... more stuff
```

To this:

```python
load(
    "@build_bazel_rules_swift//swift:swift.bzl",
    "swift_binary",
)
load(
    "//bazel:api.bzl",
    "static_library",
    "swiftpm_test",
)

... more stuff
```

Then when I run it again, it reverts this change and puts them back in the original order.

vladmos:
This commit fixes the out-of-order-load check ensuring that its fix is idempotent.

On top of that a an idempotency test is enabled for all the fixes.

Fixes #457

### DIFF
--- a/warn/warn.go
+++ b/warn/warn.go
@@ -334,6 +334,20 @@ func loadOnTopWarning(f *build.File, fix bool) []*Finding {
 }
 
 func outOfOrderLoadWarning(f *build.File, fix bool) []*Finding {
+	// compareLoadLabels compares two module names
+	compareLoadLabels := func(load1Label, load2Label string) bool {
+		// handle absolute labels with explicit repositories separately to
+		// make sure they preceed absolute and relative labels without repos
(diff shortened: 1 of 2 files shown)


laurentlb:
@ttsugriy

keith:
Verified for our case and it works as expected! Thanks for the quick turnaround

PR #461, opened 2018-11-27
[buildifier] Warn about and fix unsorted dictionary items.

ttsugriy:
Dictionary items should be sorted lexicagraphically by their keys. This makes
it easier to find the item of interest and reduces reduces chances of conflicts
when performing large-scale automated refactoring.


ttsugriy:
I cannot think of a reason as to why would someone want to leave keys unsorted, but if anyone does please let me know :) Originally I've used edit.HasComments to on a `DictExpr` to do this, but turns out that it's very tricky to add a comment to the `DictExpr` itself, so I just gave up :(

ttsugriy:
@vladmos, can you please take a look when you have a chance?

WARNINGS.md
```diff
@@ -420,6 +420,17 @@ of a symbol load and its usage can change resulting in runtime error.
 
 --------------------------------------------------------------------------------
 
+## <a name="unsorted-dict-items"></a>Dictionary items should be ordered by their keys.
+
+  * Category_name: `unsorted-dict-items`
+  * Automatic fix: yes
+
+Dictionary items should be sorted lexicagraphically by their keys. This makes
+it easier to find the item of interest and reduces reduces chances of conflicts
```
vladmos: Duplicated "reduces"

WARNINGS.md
```diff
@@ -420,6 +420,17 @@ of a symbol load and its usage can change resulting in runtime error.
 
 --------------------------------------------------------------------------------
 
+## <a name="unsorted-dict-items"></a>Dictionary items should be ordered by their keys.
```
vladmos: Please also add it to the table of contents at the top of the file.

ttsugriy:
thank you for the excellent points, @vladmos! In fact one of your points made me realize that sorting numbers and identifiers probably also makes sense :)

vladmos:
I'm not sure this check should be enabled for all dicts (or by default). I can imagine dicts that shouldn't be forced to be sorted, e.g.:

    transformation = {
        "before": ...,
        "after": ...,
    }

or

    configurations = {
        "development": ...,
        "staging": ...,
        "production": ...,
    }

In some situations this warning indeed makes sense, not sure what can be the formal criteria.

ttsugriy:
@vladmos, I can totally see that there could be exceptions, although I would imagine they are fairly rare, which is why I was actually trying to add support for a comment like `@do-not-sort` to disable sorting of items of marked with this comment dictionaries, but unfortunately the comment seems to get associated with binary operations and not dictionaries. Do you happen to have a suggestion as to how I may be able to disable sorting for a specific dictionary?

vladmos:
@ttsugriy The first thing that comes to my mind is to look at the last element of `stack` and if it's a `BinaryExpr` with `Op ==  "="`, also check whether it has this comment. This will cover both variable assignments and keyword arguments. I think it's also safe to assume that a dict literal can't be on the left side of an assignment expression.

Another similar, but not so common, situation is when a dict is nested to another dict, so you may want to also check if the parent node is a `KeyValueExpr`.

ttsugriy:
thanks for the suggestion, @vladmos. I was thinking about doing this, but thought it was a bit hacky, but since there are no better alternatives, I'll go ahead and implement this one for now :)

ttsugriy:
I've decided to be even more conservative than you've mentioned - I'm walking up the stack and in case I see any expression that as a comment `@unsorted-dict-items`, I ignore the dict altogether. Does this make sense?

warn/warn.go
```diff
@@ -413,6 +413,70 @@ func integerDivisionWarning(f *build.File, fix bool) []*Finding {
...
+	}
+
+	build.Walk(f, func(expr build.Expr, stack []build.Expr) {
+		dict, ok := expr.(*build.DictExpr)
+		if !ok {
+			return
+		}
+		// do not process dictionaries nested within expressions that do not
+		// want dict items to be sorted
+		for i := len(stack) - 1; i >= 0; i-- {
```
vladmos: What if the dict itself has a comment? I think that can still be the case, e.g.

    d.update(
        # comment 1
        {a: b},  # comment 2
    )

  If I understand correctly, `stack` doesn't contain the current node.

warn/warn_test.go
```diff
@@ -510,6 +510,118 @@ load(":b.bzl", "b")
...
+
+	checkFindingsAndFix(t, "unsorted-dict-items", `
+# @unsorted-dict-items
+d = {
+	"key" : {
+		"b": "b value",
+		"a": "a value",
+	}
+}`, `
+# @unsorted-dict-items
```
vladmos: Tip: you can just use `checkFindings` if the input and the output are the same. It also ensures that the code is not modified because fixes are not expected.

warn/warn.go
ttsugriy: interesting, I could not find a way to associate a comment with a dictionary, which is why I had to use `stack` in the first place, but the examples you've provided actually manage to achieve this :D Added tests and code to make sure they are handled properly. Thanks for your help!

thomasvl:
I was actually talking about the literal passed as `rule(..., attrs = { ... }, ...` - https://docs.bazel.build/versions/master/skylark/lib/globals.html#rule

That's a literal dict in most sources, so would this mean it has to be sorted meaning, a rule author can't front load the "important" things to help in discovery when reading the bzl?

thomasvl:
Would these examples trip things up:

https://github.com/bazelbuild/examples/blob/64ffe9b0ca1f7335f2bef37f09f9e4c3acf0adef/rules/attributes/printer.bzl#L20-L24

https://github.com/bazelbuild/examples/blob/master/rules/depsets/foo.bzl#L26-L29

https://github.com/bazelbuild/examples/blob/master/rules/depsets/foo.bzl#L47-L56

ttsugriy:
@thomasvl, for the files you've mentioned the latest version of buildifier produces the following patch:
```
diff --git a/rules/attributes/printer.bzl b/rules/attributes/printer.bzl
index 1a7abf1..facd739 100644
--- a/rules/attributes/printer.bzl
+++ b/rules/attributes/printer.bzl
@@ -18,8 +18,8 @@ def _impl(ctx):
 printer = rule(
     implementation = _impl,
     attrs = {
+        "deps": attr.label_list(allow_files = True),
         # Do not declare "name": It is added automatically.
         "number": attr.int(default = 1),
-        "deps": attr.label_list(allow_files = True),
     },
 )
diff --git a/rules/depsets/foo.bzl b/rules/depsets/foo.bzl
index 5923a64..01bede4 100644
--- a/rules/depsets/foo.bzl
+++ b/rules/depsets/foo.bzl
@@ -24,8 +24,8 @@ def _foo_library_impl(ctx):
 foo_library = rule(
     implementation = _foo_library_impl,
     attrs = {
-        "srcs": attr.label_list(allow_files = True),
         "deps": attr.label_list(),
+        "srcs": attr.label_list(allow_files = True),
     },
 )

@@ -45,14 +45,14 @@ def _foo_binary_impl(ctx):
 foo_binary = rule(
     implementation = _foo_binary_impl,
     attrs = {
-        "srcs": attr.label_list(allow_files = True),
-        "deps": attr.label_list(),
         "_foocc": attr.label(
             default = Label("//depsets:foocc"),
             allow_files = True,
             executable = True,
             cfg = "host",
         ),
+        "deps": attr.label_list(),
+        "srcs": attr.label_list(allow_files = True),
     },
     outputs = {"out": "%{name}.out"},
 )
```

[gist](https://gist.github.com/ttsugriy/0d06af12e63ce75bbc48f759a8c9c00c)

Does this look reasonable?

thomasvl:
That's part of my concern, if you format a BUILD file that uses this rule, won't it move `deps` after `srcs` because the order when the rules are used isn't purely alphabetical.  There is some priority given to the order of the rule attributes and deps is considered "lower"?

thomasvl:
ps - the `_foocc` private attribute really should go at the end, so there likely needs to be some special casing around private attributes for rules at a minimum.

ttsugriy:
AFAIK the order of attributes in the rule definition does not have to match the one used in a build file, since [positional-args](https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#positional-args) warning should prevent users from passing arguments to rules without keywords.

As for special-casing private fields, I can totally create a PR to do this :)

thomasvl:
I know they don't work positionally, but my question was about being able to order them the same sorta order one would expect when using the rule.

It seems odd that we will force the order in usage to be different than the order in declaration.

ttsugriy:
oh, I see what you're saying @thomasvl. I'll see how hard it would be to reuse the same logic, although I'm curious as to the scope in which priorities should apply - is it just for `attrs` or all dictionary literals?

ttsugriy:
@thomasvl, #469

PR #466, opened 2018-11-28
[buildifier] Fix load order comparison warning.

ttsugriy:
Currently the warning would be printed in case two identical labels
are compared, which cannot be fixed, since reordering the same load
would not change the comparison result.

This change makes the warning flag only instances where `i`th label
is "less than" the `i-1`th label.


ttsugriy:
cc @vladmos

vladmos:
Thanks!

PR #469, opened 2018-12-04
[buildifier] Take name priority into account when ordering dictionary items.

ttsugriy:
Since it's common to use dictionaries for passing `**kwargs` to rules, their
ordering should match the ordering used rule attributes in build files.

Build files rely on `NamePriority` `tables` dictionary to adjust priorities
according to common or project-specific guidelines and now the same rules
apply to ordering dictionary items.

### DIFF
--- a/warn/warn.go
+++ b/warn/warn.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/bazelbuild/buildtools/build"
 	"github.com/bazelbuild/buildtools/edit"
+	"github.com/bazelbuild/buildtools/tables"
 )
 
 // A Finding is a warning reported by the analyzer. It may contain an optional suggested fix.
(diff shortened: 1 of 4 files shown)


warn/warn.go
```diff
@@ -8,6 +8,8 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/bazelbuild/buildtools/tables"
```
vladmos: Nit: please move after `"github.com/bazelbuild/buildtools/edit"`

warn/warn_test.go
```diff
@@ -601,6 +601,18 @@ d = {
...
+	checkFindingsAndFix(t, "unsorted-dict-items", `
+d = {
+	"deps": [],
+	"srcs": ["foo.go"],
+}`, `
+d = {
+	"srcs": ["foo.go"],
+	"deps": [],
+}`,
+		[]string{"3: Dictionary items are out of their lexicographical order."},
```
vladmos: I'm not sure if this message is clear in case priorities from `NamePriority` are used, but can't suggest anything better.

PR #476, opened 2018-12-10
[buildifier] Fail when fixing with warnings

keith:
Since some warnings cannot be automatically fixed, this change makes
buildifier exit with a non zero status code in that case. That way if
you're running this on CI, you know you there are still warnings that
need to be resolved.

### DIFF
--- a/buildifier/buildifier.go
+++ b/buildifier/buildifier.go
@@ -373,7 +373,10 @@ func processFile(filename string, data []byte, inputType, lint string, warningsL
 			exitCode = 4
 		}
 	case "fix":
-		warn.FixWarnings(f, pkg, warningsList, *vflag)
+		hasWarnings := warn.FixWarnings(f, pkg, warningsList, *vflag)
+		if hasWarnings == true {
+			exitCode = 4
+		}
 	}
 
 	if *filePath != "" {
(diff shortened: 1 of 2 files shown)


keith:
I think it might be nice to print the warnings that made this fail as well, thoughts?

vladmos:
It's a bit tricky with the warnings: they contain locations (at least line numbers are printed), and these locations may be incorrect due to other fixes applied or because the file hasn't been formatted properly (and `--lint=fix` reformats it).

laurentlb:
I think `fix` shouldn't fail in this case. If you want to know if there are warnings, use `--lint=warn`.

keith:
Any other thoughts on this one or should I drop this change?

keith:
thoughts @vladmos ?

vladmos:
I'm still not sure. Maybe it should be controlled with a flag or an environment variable?

In your case failing makes sense, but there can be other situations when you want to cleanup automatically as much as possible (e.g. as a commit hook), throwing an error may break such usages.

You can always run `buildifier --lint=fix && buildifier --lint=warn` to fix everything that's possible and throw an error in case anything is left. May be an overkill, but in my experience buildifier is quite fast so it shouldn't be a bottleneck.

keith:
I'll drop this for now

PR #481, opened 2018-12-14
[buildifier] Make sure default select condition is the last one.

ttsugriy:


tables/tables.go
```diff
@@ -192,13 +192,14 @@ var NamePriority = map[string]int{
...
-	"implements":     6,
-	"alwayslink":     7,
+	"destdir":              1,
+	"exports":              2,
+	"runtime_deps":         3,
+	"deps":                 4,
+	"implementation":       5,
+	"implements":           6,
+	"alwayslink":           7,
+	"//conditions:default": 50, // default condition is the last one by convention
```
laurentlb: The other lines are mostly for keyword-arguments when calling a rule. Since this last line can be surprising I recommend updating the comment.

It should mention it's for the dictionary in select.

thomasvl:
Thanks for doing this.

@laurentlb can this make the next release?

vladmos:
Thanks!

Issue #220, opened 2018-02-26
Diff mode doesn't include the file that needs updating


thomasvl:
Atleast in the last released version (0.6.0), diff mode doesn't say what file the diff is for, that isn't very useful when it is checking multiple files at once.

thomasvl:
Still a problem in 0.11.1.

pmbethe09:
Will this break any existing tooling?

thomasvl:
Nothing I've written. Since there are no file markers, if you pass multiple files, you have no way to know what the diffs apply to, so I can't really see how someone could have automated it with multiple files.

The one case they could automate is single file mode, because then you know what the diffs apply to.  So if you don't want to break any tooling that does that, don't print the file header if invoked on a single file.

thomasvl:
Still and issue in 0.19.2

thomasvl:
0.22.0 does seem to be fixed

PR #486, opened 2018-12-18
Display the filenames for --mode=diff

Profpatsch:
Previously, only the lines were output, which makes tracking down the
changes a bit hard.

I’m assuming the `--mode=diff` output is not intended to be parse-able by e.g. `patch`, but I might be wrong.

### DIFF
--- a/buildifier/buildifier.go
+++ b/buildifier/buildifier.go
@@ -376,7 +376,9 @@ func processFile(filename string, data []byte, inputType, lint string, warningsL
 			return
 		}
 		infile := filename
+		displayFilename := ""
 		if filename == "" {
+			displayFilename = "<stdin>"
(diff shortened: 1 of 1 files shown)


thomasvl:
Likely fixing #220

buildifier/buildifier.go
```diff
@@ -376,7 +376,9 @@ func processFile(filename string, data []byte, inputType, lint string, warningsL
 			return
 		}
 		infile := filename
+		displayFilename := ""
```
vladmos: Nit: the code will look a bit simpler if you replace this line with `displayFilename = filename` and remove the else-clause.

Profpatsch: Those are two different things, didn’t want to confuse them.

PR #488, opened 2018-12-20
Remove skylib and io_bazel dependencies from WORKSPACE

mjduijn:
Seeing as these are already imported via buildifier_dependencies they are redundant.

In response to a [comment](https://github.com/bazelbuild/buildtools/pull/478#issuecomment-446767227) on an earlier PR, and I figured this would make updating versions easier. 

There is a bit of a nasty catch however, in our setup we noticed that the loading of the dependencies isn't entirely deterministic when using the _maybe rules (different target run calls may load different versions of the same dependency).
So if you intended it this way, feel free to disregard the PR.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -17,13 +17,6 @@ http_archive(
     url = "https://github.com/bazelbuild/bazel-gazelle/archive/44ce230b3399a5d4472198740358fcd825b0c3c9.tar.gz",  # 2018-12-10
 )
 
-http_archive(
-    name = "bazel_skylib",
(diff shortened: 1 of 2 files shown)


pmbethe09:
Can you also change deps.bzl to the same version of bazel skylib?
The one you are deleting is c00ef493869e2966d47508e8625aae723a4a3054 from Dec 6, the one in deps.bzl is older, d7c5518fa061ae18a20d00b14082705d3d2d885d from Nov 20.

mjduijn:
Ah missed that, fixed now

PR #491, opened 2018-12-21
Update io_bazel to 0.21.0

keith:
This fixes some incompatible issues

### DIFF
--- a/buildifier/deps.bzl
+++ b/buildifier/deps.bzl
@@ -11,11 +11,10 @@ def buildifier_dependencies():
     _maybe(
         http_archive,
         name = "io_bazel",
-        sha256 = "f59608e56b0b68fe9b18661ae3d10f6a61aaa5f70ed11f2db52e7bc6db516454",
-        strip_prefix = "bazel-0.20.0",
+        sha256 = "71b4294c18429fb730e223d1ceca135245fecda43191998d6afef39e1ba48e78",
+        strip_prefix = "bazel-0.21.0",
         urls = [
-            "http://mirror.bazel.build/github.com/bazelbuild/bazel/archive/0.20.0.tar.gz",
-            "https://github.com/bazelbuild/bazel/archive/0.20.0.tar.gz",
+            "https://github.com/bazelbuild/bazel/archive/0.21.0.tar.gz",
         ],
     )
 


buildifier/deps.bzl
```diff
@@ -11,11 +11,10 @@ def buildifier_dependencies():
     _maybe(
         http_archive,
         name = "io_bazel",
-        sha256 = "f59608e56b0b68fe9b18661ae3d10f6a61aaa5f70ed11f2db52e7bc6db516454",
-        strip_prefix = "bazel-0.20.0",
+        sha256 = "71b4294c18429fb730e223d1ceca135245fecda43191998d6afef39e1ba48e78",
+        strip_prefix = "bazel-0.21.0",
         urls = [
-            "http://mirror.bazel.build/github.com/bazelbuild/bazel/archive/0.20.0.tar.gz",
```
keith: This URL doesn't seem to exist for 0.21.0

philwo: There's no automatic mirroring. Someone from the Bazel team has to upload the file there.

vladmos: 0.20.0 is not there either, so let's just remove the line for now.

philwo: Can you instead depend on the -dist.zip? That would be available here:

https://releases.bazel.build/0.21.0/release/bazel-0.21.0-dist.zip
https://github.com/bazelbuild/bazel/releases/download/0.21.0/bazel-0.21.0-dist.zip

Two advantages:
1) The file is automatically pushed to the releases.bazel.build host as part of the release.
2) It will always have the same hash sum as opposed to the above used URL which occasionally changes (because GitHub upgrades some tools on their backend, etc.).

keith: Updated

buildifier/deps.bzl
```diff
@@ -11,11 +11,10 @@ def buildifier_dependencies():
     _maybe(
         http_archive,
         name = "io_bazel",
-        sha256 = "f59608e56b0b68fe9b18661ae3d10f6a61aaa5f70ed11f2db52e7bc6db516454",
-        strip_prefix = "bazel-0.20.0",
```
keith: These new zips don't have a prefix like the others did

Issue #367, opened 2018-08-30
buildifier using bazel with release binaries


hanwen:
I like the convenience of using bazel to version & download buildifier, but I'd rather not have our build depend on the Go toolchain. 

Would you be able to provide a rule that can be setup like 

  https://github.com/bazelbuild/buildtools/tree/master/buildifier#setup-and-usage-via-bazel

but downloads a binary release and runs that?

pmbethe09:
So this is a
macro
buildifier_binary(version="0.15.0")
or something like that?  Which just needs a base buildtools in the WORKSPACE?

hanwen:
yes; it should also automatically download for the host platform.

aiuto:
Do you normally want specific version pinning, or is the more common case to simply to pin the buildifier binary to the matching Bazel release?

If the most common use case is the latter, we could solve this another way - simply have a prebuilt buildifier as part of the Bazel distribution. This goes for buildozer and any other Bazel command line tools. Because I fundamentally agree with you. When I get Bazel, I expect to have binaries for all the tools I need without having to build them (especially with toolchains I may not have).

Not that we should just bloat the distribution. We could package Bazel as two components.
1. bazel-core which is just bazel
2. bazel which depends on bazel-core but includes pre-built for everything under buildtools

oferb:
Here's an example of how to bundle these tools with a repo:
https://github.com/google/startup-os/tree/master/tools/buildtools_wrappers
https://github.com/google/startup-os/blob/master/WORKSPACE#L67-L107

You can then call them the same way on both Linux and Mac:
https://github.com/google/startup-os/blob/master/commands.txt#L1-L8

aiuto:
My latest thinking on this is that there will be two kinds of Bazel releases: "bazel" and "bazel-full". Alongside that, there will be binary releases for Buildifier.  At each Bazel release we will test against a specific buildifier release. The bazel packaging will stand alone. The bazel-full package will depend on the tested buildifier release.  If you install bazel-full, you get the tools at the needed versions, but you can still do an apt-get to get a newer buildifier release when it is available.  This still needs to be fleshed out for multiple platforms before it is a real plan, but the basic idea is that a large class of users should be able to just install bazel-full and not hunt for anything else.

PR #495, opened 2019-01-02
Add command line arguments -diff and -multi_diff.

brown:
The current technique buildifier uses to select a diff program and set the multidiff flag is
obscure.  It relies on undocumented environment variables, involves grepping for the
string "tkdiff" in a command name, etc.  Buildifier should take command line arguments
instead.  As a first step toward this goal, let's add flags for the diff program and whether
it supports tkdiff-style multi diff.  Later, we can deprecate the undocumented environment
variables, etc.

### DIFF
--- a/buildifier/buildifier.go
+++ b/buildifier/buildifier.go
@@ -45,6 +45,8 @@ var (
 	vflag         = flag.Bool("v", false, "print verbose information on standard error")
 	dflag         = flag.Bool("d", false, "alias for -mode=diff")
 	mode          = flag.String("mode", "", "formatting mode: check, diff, or fix (default fix)")
+	diffProgram   = flag.String("diff_command", "", "command to run when the formatting mode is diff (default uses the BUILDIFIER_DIFF, BUILDIFIER_MULTIDIFF, and DISPLAY environment variables to create the diff command)")
+	multiDiff     = flag.Bool("multi_diff", false, "the command specified by the -diff_command flag can diff multiple files in the style of tkdiff (default false)")
 	lint          = flag.String("lint", "", "lint mode: off, warn, or fix (default off)")
 	warnings      = flag.String("warnings", "", "comma-separated warnings used in the lint mode or \"all\"")
 	filePath      = flag.String("path", "", "assume BUILD file has this path relative to the workspace directory")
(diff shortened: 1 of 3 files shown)


buildifier/buildifier.go
```diff
@@ -66,12 +68,13 @@ func stringList(name, help string) func() []string {
 }
 
 func usage() {
-	fmt.Fprintf(os.Stderr, `usage: buildifier [-d] [-v] [-mode=mode] [-lint=lint_mode] [-path=path] [files...]
+	fmt.Fprintf(os.Stderr, `usage: buildifier [-d] [-v] [-diff=diff_program] [-mode=mode] [-lint=lint_mode] [-path=path] [files...]
```
gertvdijk: Please add the new `-multi_diff` flag as well.

```suggestion
	fmt.Fprintf(os.Stderr, `usage: buildifier [-d] [-v] [-diff=diff_program] [-multi_diff] [-mode=mode] [-lint=lint_mode] [-path=path] [files...]
```

buildifier/buildifier.go
```diff
@@ -200,6 +203,10 @@ func main() {
 	}
 
 	diff = differ.Find()
+	if *diffProgram != "" {
+		diff.Cmd = *diffProgram
```
gertvdijk: Unfortunately, it seems like it does not allow me to specify arguments with the command. E.g.:

    BUILDIFIER_DIFF="diff -u" buildifier [...]

works, but

    buildifier -diff="diff -u" [...]

gives me

    ./BUILD:
    --: diff -u: command not found
    exit status 127

Or am I doing something wrong?

(I could not find an obvious reason why it works via the environment; I'm not too skilled in Go.)

gertvdijk:
Nice. I was looking for this. Would you mind to consider adding support for this when buildifier is run as Bazel rule?

See https://github.com/gertvdijk/buildtools/commit/af089f9c1a58def66a4e29dba7ba82d9629ff5a1 for a starting point, which I created on top of your commit.

buildifier/buildifier.go
brown: I uploaded a new version of the commit with this change.  I also changed the flag name to
"diff_command" since some people will want to set it to "diff -u" ... which seems better to
refer to as the diff command to run.

I'm not now changing the default behavior.  I can remove the environment variables in a
second commit ... maybe after some time of warning that they are deprecated.

buildifier/buildifier.go
brown: I think you must be doing something wrong.  Using -diff_command="diff -u" works for me.

gertvdijk:
@brown 

> I don't understand why there's a buildifier Bazel rule at all. Bazel is for compiling code. I can run
> buildifier from the shell without Bazel.

It allows for **running** buildifier using Bazel within your project, e.g.:

    bazel run //:buildifier

See also: [**Setup and usage via Bazel**](https://github.com/bazelbuild/buildtools/tree/fbc50a990dcf0dec0c3e947c2067c9656c62bfb1/buildifier#setup-and-usage-via-bazel)

The main advantages: versioning the Buildifier used within the project/branch along with the settings, developers don't have to install it themselves.

----

My approach in the aforementioned commit was to patch the wrapper shell script, but your approach using code is much more elegant. However, exposing the settings in the Bazel rule would make is actually useful for those who run buildifier with Bazel.

brown:
I've updated the commit to include changes to def.bzl.
I've verified that "diff -u" works as the diff command both from the command line and using the buildifier rule in def.bzl, so I think the commit is ready to merge.

I still think the buildifier Bazel rule is weird.  Why not just download buildifier and run it?
The rule is over 100 lines of code just to implement the functionality of a 5 line shell script.
Wouldn't it make more sense to implement a "-R" argument to buildifier that reformats an
entire directory tree?

gertvdijk:
> I still think the buildifier Bazel rule is weird. Why not just download buildifier and run it?

Apart from what I already mentioned in my previous comment... Not everyone likes the idea of having to download/manage all kinds of tooling/linters/checkers on their system. Versions to be used are heavily depending on the project/branch in my situation for example. It's a way better developer-experience if your code formatting is actually shared across your team / codebase. Bazel let's you do that, and we love it.

Related discussion maybe: #367.

> The rule is over 100 lines of code just to implement the functionality of a 5 line shell script.
Wouldn't it make more sense to implement a "-R" argument to buildifier that reformats an
entire directory tree?

Yeah, that makes a lot of sense. Surprisingly, I couldn't find a feature request for that yet, so I filed one: #516.

buildifier/buildifier.go
gertvdijk: Apparently I was doing something wrong indeed. It's working fine with your commit and

```
buildifier(
    name = "buildifier_diff",
    diff_command = "diff -u",
    mode = "diff",
    verbose = True,
)
```

:+1:

brown:
Do I need to do anything else to get this commit applied to master?  
gertvdijk approved the change, but I don't see it yet on the master branch.

vladmos:
Thanks!

PR #503, opened 2019-01-08
Provide a flexible way of enabling disabling warnings

### LINKED ISSUE #493: FR: Disabling warnings with `--warnings`
Currently `--warnings=` requires the completely list of warning one wants to enable, but as more warning are added, that means someone using this to skip a few tests ends up freezing the set of validations and doesn't pick up new ones.

It would be nice if there was a way to exclude warnings instead, say similar to how bazel has some targets that can take `-` prefixed targets/tags in some case to subtrack out things.  So say something like `--warnings=-out-of-order-load,-unsorted-dict-items` to get the default (`all`), except for those two warnings.

vladmos:
* Mark certain warnings as non-default (currently `out-of-order-load` and `unsorted-dict-items`), they won't be applied unless explicitly added (or with `--warnings=all`)
  * Allow users to provide `+` and `-` modifiers to the warnings list. E.g `--warnings=-foo,+bar` means all default warnings except "foo" but with "bar" (which is assumed to be non-default).

Fixes #493

### DIFF
--- a/buildifier/buildifier.go
+++ b/buildifier/buildifier.go
@@ -46,7 +46,7 @@ var (
 	dflag         = flag.Bool("d", false, "alias for -mode=diff")
 	mode          = flag.String("mode", "", "formatting mode: check, diff, or fix (default fix)")
 	lint          = flag.String("lint", "", "lint mode: off, warn, or fix (default off)")
-	warnings      = flag.String("warnings", "all", "comma-separated warnings used in the lint mode or \"all\" (default all)")
+	warnings      = flag.String("warnings", "", "comma-separated warnings used in the lint mode or \"all\"")
 	filePath      = flag.String("path", "", "assume BUILD file has this path relative to the workspace directory")
 	tablesPath    = flag.String("tables", "", "path to JSON file with custom table definitions which will replace the built-in tables")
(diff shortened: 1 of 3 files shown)


vladmos:
@ttsugriy FYI: these two warnings are becoming not enabled by default, but still accessible with `--warnings=all`

ttsugriy:
thanks for the heads up, @vladmos. I'll update our internal config to enable all warnings.

buildifier/buildifier.go
```diff
@@ -172,10 +172,38 @@ func main() {
...
 	switch *warnings {
-	case "", "all":
+	case "", "default":
+		warningsList = warn.DefaultWarnings
+	case "all":
 		warningsList = warn.AllWarnings
 	default:
-		warningsList = strings.Split(*warnings, ",")
+		// Either all or no warning categories should start with "+" or "-".
+		// If all of them start, the semantics is "default set of warnings + something - something".
```
thomasvl: "// If all of them start, "

Doesn't seem like a complete thought, was this "If all fo them start with '+'" or something like that>?

vladmos: Thanks for catching. The corrrect sentence is "if all of them start with `+` or `-`.

PR #504, opened 2019-01-08
[warn] Sort loads by package and file name separately.

### LINKED ISSUE #501: Confirming the buildifier behavior of out-of-order-load checks.
The behavior here might just be what we're stuck with since it likely matches what has been done with BUILD files for labels on other attributes (`deps`, etc.), but when looking at load statements, a _pure_ lexicographical ordering seems odd.

Sending:
```
"""Module comment."""

load("//foo:xyz.bzl", "xyz")
load("//foo/bar:mno.bzl", "mno")

def something():
    xyz.do()
    mno.do()
    return True
```

Through `buildifier --lint=fix` reorder the `load()` calls:

```
"""Module comment."""

load("//foo/bar:mno.bzl", "mno")
load("//foo:xyz.bzl", "xyz")

def something():
    xyz.do()
    mno.do()
    return True
```

While ":" > "/", this feels sorta wrong.  It seems like all the `load()`s from _//foo_ should get listed before the ones from _//foo/bar_ since this is down a level from a package pov.

Also, given where ":", "/", and the numbers (_0-9_) fall in ASCII, it also means:

```
"""Module comment."""

load("//foo:xyz.bzl", "xyz")
load("//foo2:mno.bzl", "mno")

def something():
    xyz.do()
    mno.do()
    return True
```

re-sorts to:

```
"""Module comment."""

load("//foo2:mno.bzl", "mno")
load("//foo:xyz.bzl", "xyz")

def something():
    xyz.do()
    mno.do()
    return True
```

So rather than a pure lexicographical sort of the string, should the string be split into the package and name, and that tuple be sorted for the order of the load statements?

ttsugriy:
Previously, the sorting was performed based on the entire module string
lexicographically, which would result in somewhat suboptimal ordering of
things like
```
load("//foo2:mno.bzl", "mno")
load("//foo:xyz.bzl", "xyz")
```
and
```
load("//foo/bar:mno.bzl", "mno")
load("//foo:xyz.bzl", "xyz")
```

This change makes it so the ordering logic first tries to compare loads
by package alone and break ties using file names.

Fixes #501

### DIFF
--- a/warn/warn.go
+++ b/warn/warn.go
@@ -343,8 +343,26 @@ func outOfOrderLoadWarning(f *build.File, fix bool) []*Finding {
 		isExplicitRepo1 := strings.HasPrefix(load1Label, "@")
 		isExplicitRepo2 := strings.HasPrefix(load2Label, "@")
 		if isExplicitRepo1 == isExplicitRepo2 {
-			// Either both labels have explicit repository names or both don't, compare lexicographically
-			return load1Label < load2Label
+			// Either both labels have explicit repository names or both don't, compare their packages
+			// and break ties using file names if necessary
+
+			module1Parts := strings.Split(strings.TrimLeft(load1Label, "@"), ":")
+			package1, filename1 := module1Parts[0], module1Parts[1]
+			module2Parts := strings.Split(strings.TrimLeft(load2Label, "@"), ":")
+			package2, filename2 := module2Parts[0], module2Parts[1]
+
+			// in case both packages are the same, use file names to break ties
+			if package1 == package2 {
+				return filename1 < filename2
+			}
+
+			// in case one of the packages is empty, the empty one goes first
+			if len(package1) == 0 || len(package2) == 0 {
+				return len(package1) > 0
+			}
+
+			// both packages are non-empty and not equal, so compare them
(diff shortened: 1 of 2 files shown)


warn/warn.go
```diff
@@ -344,7 +344,31 @@ func outOfOrderLoadWarning(f *build.File, fix bool) []*Finding {
...
-			return load1Label < load2Label
+			canonicalModule1 := strings.TrimLeft(load1Label, "@")
+			canonicalModule2 := strings.TrimLeft(load2Label, "@")
+
+			module1Parts := strings.Split(canonicalModule1, ":")
+			package1, filename1 := module1Parts[0], module1Parts[1]
+			module2Parts := strings.Split(canonicalModule2, ":")
+			package2, filename2 := module2Parts[0], module2Parts[1]
+
+			// in case both loads lack packages, compare by file name
```
thomasvl: Do you need these two checks or can you go straight to the package1 != package2 case? (delete 355 - 358)

If both packages are identical (including both empty), it would go to the `return filename1 < filename2`; but if only one package is empty, wouldn't the `return package1 < package2` still do the right thing?

PR #510, opened 2019-01-10
Safer duplicated load fix

vladmos:
If a symbol is loaded twice from the same label, it's safe to delete the second time it's loaded. It may be not safe to delete the first time because it may be used in between.
If a symbol is loaded twice from different labels, it's not safe to automatically delete it, instead just a warning should be shown. For instance, the following code should be fixed manually by renaming the loaded symbols and eliminating the confusion:

    load("foo", "a")

    a()  # from foo

    load("bar", "a")

    a()  # from bar

### DIFF
--- a/build/lex.go
+++ b/build/lex.go
@@ -42,11 +42,15 @@ const (
 )
 
 func (t FileType) String() string {
-	return [...]string{
-		".bzl",
-		"BUILD",
-		"WORKSPACE",
-	}[t]
+	switch t {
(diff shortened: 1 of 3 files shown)


laurentlb:
You may want to add a test with aliases?

vladmos:
Added a test, also improved the logic to also take the original variable name into account, e.g. the following loaded symbols are now considered different:

    load("foo", "a")
    load("foo", a = "b")

PR #515, opened 2019-01-14
Add a check for missing return values

vladmos:
Warn if a function explicitly returns values somewhere, but not everywhere where it can terminate.

### DIFF
--- a/warn/warn.go
+++ b/warn/warn.go
@@ -960,6 +960,79 @@ func nativePackageWarning(f *build.File, fix bool) []*Finding {
 	return findings
 }
 
+// findEmptyReturns searches for return statements without a value and returns whether
+// the current list of statements terminates (either by a return or fail() statements on the current
+// level in all subranches.
+func findEmptyReturns (stmts []build.Expr, callback func(build.Expr)) bool {
+	if len(stmts) == 0 {
+		// May occur in empty else-clauses
+		return false
+	}
+	terminated := false
+	for _, stmt := range stmts {
+		switch stmt := stmt.(type) {
+		case *build.ReturnStmt:
+			if stmt.Result == nil {
+				callback(stmt)
+			}
+			terminated = true
+		case *build.CallExpr:
+			ident, ok := stmt.X.(*build.Ident)
+			if ok && ident.Name == "fail" {
+				terminated = true
+			}
+		case *build.IfStmt:
+			// Save to separate values to avoid short circuit evaluation
+			term1 := findEmptyReturns(stmt.True, callback)
+			term2 := findEmptyReturns(stmt.False, callback)
+			if term1 && term2 {
+				terminated = true
+			}
+		}
+	}
+	return terminated
+}
+
+func missingReturnValueWarning(f *build.File, fix bool) []*Finding {
+	findings := []*Finding{}
+
+	for _, stmt := range f.Stmt {
+  	function, ok := stmt.(*build.DefStmt)
+  	if !ok {
+			continue
+		}
+
+  	var hasNonEmptyReturns bool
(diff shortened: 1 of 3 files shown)


vladmos:
I've just realized I don't take for-loops into account, i.e. don't search for empty return statements there. I'll fix it tomorrow.

vladmos:
Can be reviewed now.

warn/warn_test.go
```diff
@@ -1107,3 +1107,132 @@ native.cc_library(name = "lib")
 		`:1: "native.package()" shouldn't be used in .bzl files.`,
 	}, scopeBzl)
 }
+
+func TestMissingReturnValueWarning(t *testing.T) {
```
laurentlb: Did you check the tests from Skylint?

warn/warn.go
```diff
@@ -960,6 +960,83 @@ func nativePackageWarning(f *build.File, fix bool) []*Finding {
 	return findings
 }
 
+// findEmptyReturns searches for return statements without a value and returns whether
+// the current list of statements terminates (either by a return or fail() statements on the current
+// level in all subranches.
```
laurentlb: I found it hard to guess what the function does based on its name. Do you have an idea for a better name?

The comment should say what `callback` is for.

warn/warn.go
```diff
@@ -960,6 +960,83 @@ func nativePackageWarning(f *build.File, fix bool) []*Finding {
...
+			terminated = true
+		case *build.CallExpr:
+			ident, ok := stmt.X.(*build.Ident)
+			if ok && ident.Name == "fail" {
+				terminated = true
+			}
+		case *build.ForStmt:
+			// Even if a for-loop is guaranteed to terminate in each iteration, buildifier still can't
+			// check whether the loop is not empty
+			findEmptyReturns(stmt.Body, callback)
```
laurentlb: I don't understand the comment and why you call the function (but ignores its return value).

warn/warn_test.go
vladmos: Yes, they are more or less similar to mine.

warn/warn.go
vladmos: Renamed to `findReturnsWithoutValue`, updated the comment.

warn/warn.go
vladmos: The function tracks possible execution paths to determine whether it can reach the end where the implicit `return` is. The following example is considered as a termination point:

    if foo:
        return foo
    else:
        return bar
    # no implicit return here

however the following is not:

    for x in y:
        return x
    # implicit return here if the for loop above is empty

warn/warn.go
```diff
@@ -960,6 +960,85 @@ func nativePackageWarning(f *build.File, fix bool) []*Finding {
...
+			continue
+		}
+
+		var hasNonEmptyReturns bool
+		build.Walk(function, func(expr build.Expr, stack []build.Expr) {
+			ret, ok := expr.(*build.ReturnStmt)
+			if !ok {
+				return
+			}
+			if ret.Result != nil {
```
laurentlb: `if ret, ok := expr.(*build.ReturnStmt); ok && ret.Result != nil {`

warn/warn.go
```diff
@@ -960,6 +960,85 @@ func nativePackageWarning(f *build.File, fix bool) []*Finding {
...
+			term2 := findReturnsWithoutValue(stmt.False, callback)
+			if term1 && term2 {
+				terminated = true
+			}
+		}
+	}
+	return terminated
+}
+
+func missingReturnValueWarning(f *build.File, fix bool) []*Finding {
```
laurentlb: Add short comment

PR #517, opened 2019-01-15
Warnings for unreachable statements

vladmos:
Warn about statements that are not reachable by the execution.


warn/warn_test.go
```diff
@@ -1235,3 +1235,68 @@ def foobar():
...
+def bar():
+  for x in y:
+    if x:
+      break
+    elif y:
+      continue
+    else:
+      return x
+
+    foo()
```
laurentlb: Add a comment on unreachable lines (it's not obvious which one is line 19 otherwise)

Issue #500, opened 2019-01-07
Enable "args-order" by default


laurentlb:
https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#function-call-arguments-should-be-in-the-following-order

I think Buildifier should fix this by default (when formatting), as it will be an error in Starlark. The change should affect very few users, as far as I know.

vladmos:
Fixed with #521.

PR #519, opened 2019-01-16
Refactor warn.go

vladmos:
Currently `warn.go` contains almost all warning functions that are defined in buildifier (except a few bazel API-specific functions defined in `warn_bazel.go`, which makes it hard to navigate in the file.

This commit splits the file into the following submodules:

`warn_execution.go`: warnings related to the execution of a file:
  * missing return values
  * unreachable statements 
  * statements with no effect
  * unused variables
  * unused loads
  * redefined variables

`warn_operation.go`: deprecated operations
  * dictionary concatenation
  * depset union
  * string iteration
  * depset iteration
  * integer division with `/`

`warn_syntax.go`: deprecated syntax
  * wrong parameters order (e.g. **kwargs before positional)

`warn_appearance.go`: warnings about appearance of certain elements in the file
  * `load` on top
  * `package` on top
  * multiple `load`s from the same origin
  * unsorted dict keys
  * unsorted `load` statements

`warn_api.go`: former `warn_bazel.go` plus the following:
  * old-style providers
  * rules with duplicated names
  * rules with positional arguments

`warn_misc.go`: warnings that don't fall into the categories above
  * constant globs
  * `native` in BUILD files
  * `native.package` in .bzl files


laurentlb:
I'd prefer to have Bazel-specific warnings in a separate file (anything related to depset, rules, globs, providers and `package`).

Rename `execution` to `control_flow`?
Rename `appearance` to `cosmetic`?

We can have a file for language constructs that should be forbidden:
 - dictionary concatenation
 - string iteration
 - integer division with /
 - wrong parameters order
 - multiple loads from the same origin

vladmos:
> wrong parameters order

I'm going to move this check to `build/rewrite.go` (but not in this PR).

vladmos:
>I'd prefer to have Bazel-specific warnings in a separate file (anything related to depset, rules, globs, providers and package).

`warn_bazel.go` is already too large and may require further splitting. I've renamed `warn_api.go` back to `warn_bazel.go`, moved depset-related functions from  `warn_syntax.go` to `warn_bazel_syntax.go`, and renamed `warn_misc.go` to `warn_bazel_misc.go`.

> Rename execution to control_flow?
> Rename appearance to cosmetic?

Done.

> We can have a file for language constructs that should be forbidden:

> dictionary concatenation
> string iteration
> integer division with /

That's a single file now (`warn_operation.go`)

> wrong parameters order

That's the only function in `warn_syntax.go`, will be converted to a rewrite (#500)

> multiple loads from the same origin

That's rather a cosmetic change.

laurentlb:
Overall, it looks good, except for the distinction between `warn_bazel.go` and `warn_bazel_misc.go`. Can you think of a better name or a better split?

vladmos:
What about: everything about deprecated API (e.g. the `cfg = "data"` configuration, global `git_repository`, `attr.license()` is moved to `warn_bazel_api.go`, everything else from `warn_bazel.go` and `warn_bazel_misc.go` is moved to `warn_bazel.go` (e.g. constant globs, positional rule arguments, duplicated rule names)? So the split is "deprecated API" vs "general common mistakes / best practices".

laurentlb:
Yes, that's good

PR #522, opened 2019-01-18
Enable tests on Windows

meteorcloudy:


laurentlb:
This is great, except that the test is failing.
I think we should fix it before merging this PR.

meteorcloudy:
Definitely we should fix the tests. Can you look into `//build:go_default_test`, I'm not sure what's happening, but it has a clear error message? 

I can check `//extra_actions_base_proto:extra_actions_base.gen.pb.go_checkshtest` because it's passing on my local machine.

laurentlb:
@vlasmos

laurentlb:
@vladmos

meteorcloudy:
@vladmos Do you have time to look into the failure of `//build:go_default_test`?

vladmos:
I'm looking at the moment. It seems like buildifier uses os-specific methods of `path` to perform operations on labels (which are not os-specific), therefore they're behaving differently on different platforms. Will send a fix shortly, after which `go_default_test` should pass here.

meteorcloudy:
@vladmos Thank you!

meteorcloudy:
Tests passed! @laurentlb Can you LGTM this?

PR #523, opened 2019-01-18
Warn about local variables that may not have been initialized

vladmos:


vladmos:
Add a check for terminated execution.

PR #524, opened 2019-01-21
Create a separate node for pass, break, and continue statements

vladmos:
Currently the `Ident` node is used to represent `pass`, `break`, and `continue` statements, which made the following code valid:

    if break:
        pass + continue

### DIFF
--- a/build/lex.go
+++ b/build/lex.go
@@ -560,6 +560,14 @@ func (in *input) Lex(val *yySymType) int {
 	if k := keywordToken[val.tok]; k != 0 {
 		return k
 	}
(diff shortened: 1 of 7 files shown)


build/lex.go
```diff
@@ -642,6 +650,8 @@ func (in *input) order(v Expr) {
 		// nothing
 	case *Ident:
 		// nothing
+	case *BranchStmt:
```
pmbethe09: future clean-up: can use , to combine 'nothing' cases

case *Ident, *BranchStmt, *StringExpr:
   // nothing

PR #528, opened 2019-01-25
Add support for shortened label names in buildozer 'set', 'set if abs…

pdoyle12:
…ent', and 'replace' commands. Support already exists in the 'add' command.


pmbethe09:
From https://github.com/bazelbuild/buildtools/issues/532 -- can this also be done for 'remove'?

pdoyle12:
Re: tests -- currently it looks like there are no tests for any of the buildozer commands themselves, just tests for their helper methods.  Are there integration tests anywhere, or something else I could extend?

PR #533, opened 2019-01-29
Traverse subnodes of LoadStmt

### LINKED ISSUE #527: build: WalkOnce doesn't visit LoadStmt children
`build.WalkOnce` is supposed to call its callback argument for each child of its `Expr` argument. It doesn't seem to have a case for `LoadStmt`. This prevents tools from modifying module names and symbols in load statements using `build.Edit`, which calls `WalkOnce` internally. One of the tools in the Gazelle repo needed a special case to work around this.

vladmos:
`WalkOnce` should visit `LoadStmt` children.

Fixes #527


build/walk.go
```diff
@@ -145,6 +145,13 @@ func WalkOnce(v Expr, f func(x *Expr)) {
 		f(&v.Then)
 		f(&v.Test)
 		f(&v.Else)
+	case *LoadStmt:
+		for i := range v.From {
```
jayconrod: Also need to visit `v.Module`. That's actually the main use case for me: I have a program that needs to perform a substitution on all labels in a file. It works by calling `build.Edit` and matching `*StringExpr`.

vladmos: Yes, you're right, will fix that.

build/walk.go
```diff
@@ -145,6 +145,13 @@ func WalkOnce(v Expr, f func(x *Expr)) {
...
+			from := (Expr)(v.From[i])
+			f(&from)
+			to := (Expr)(v.To[i])
+			f(&to)
```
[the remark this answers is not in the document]
vladmos: Done.

build/walk.go
vladmos: Done.

PR #534, opened 2019-01-30
Create CODEOWNERS

dslomov:


CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# Admins
+* @vladmos @laurentlb @ @pmbethe09
```
vladmos: Is the single `@` a typo?

pmbethe09:
LGTM

CODEOWNERS
dslomov: It is, thanks for catching this!

PR #536, opened 2019-02-06
Implement docstring warnings

vladmos:
For modules, ensure that a docstring exists.

For public functions that are long enough (at least 5 statements) ensure that a docstring exists and documents input and output parameters.


warn/warn_docstring.go
```diff
@@ -0,0 +1,250 @@
...
+			spaces++
+		} else {
+			break
+		}
+	}
+	return spaces
+}
+
+var argHeaderRegex = regexp.MustCompile("^ *(\\*?\\*?\\w+)( *\\([\\w\\ ,]+\\))?:")
+var argRegex = regexp.MustCompile("[\\*\\w]+")
```
laurentlb: Use raw strings instead, with \`

warn/warn_docstring.go
```diff
@@ -0,0 +1,250 @@
...
+
+// parseFunctionDocstring parses a function docstring and returns a docstringInfo object containing
+// the parsed information about the function, its arguments and its return value
```
laurentlb:
```suggestion
// the parsed information about the function, its arguments and its return value.
```

warn/warn_docstring.go
```diff
@@ -0,0 +1,250 @@
...
+	if len(blocks) > 0 && len(blocks[0].lines) == 1 {
+		// Exactly one line in the first block
+		info.hasHeader = true
+	}
+
+	// Iterate over the blocks, extract data
+	for _, block := range blocks {
+		switch block.lines[0] {
+		case "Args:", "Arguments:":
+			if block.lines[0] == "Arguments:" {
```
laurentlb: Add comment, to state why you look at "Arguments"

warn/warn_docstring.go
```diff
@@ -0,0 +1,250 @@
...
+				info.argumentsLine = block.startLineNo
+			}
+
+			argIndentation := -1
+			for _, line := range block.lines[1:] {
```
laurentlb: Add comment - what is this doing?

warn/warn_docstring.go
```diff
@@ -0,0 +1,250 @@
...
+			continue
+		}
+		if strings.HasPrefix(def.Name, "_") {
+			// Private functions aren't required to have docstrings
+			continue
+		}
+		if stmtsCount(def.Body) < FunctionLengthDocstringThreshold {
+			// The function is too short
+			continue
+		}
```
laurentlb: Optional, can be done later:
For private and short functions, we might still want to make sure that the docstring (if any) is correct.

warn/warn_docstring.go
```diff
@@ -0,0 +1,250 @@
...
+				}
+			}
+
+			if hasReturnValues(def) && !info.returns {
+				findings = append(findings, makeFinding(f, start, end, "function-docstring",
+					fmt.Sprintf(`Return value of "%s" is not documented.`, def.Name), true, nil))
+
+			}
+
+		} else {
```
laurentlb: swap conditions
Add continue to avoid the `else`

warn/warn_docstring.go
```diff
@@ -0,0 +1,250 @@
...
+	build.Walk(def, func(expr build.Expr, stack []build.Expr) {
+		ret, ok := expr.(*build.ReturnStmt)
+		if ok && ret.Result != nil {
+			result = true
+		}
+	})
+	return result
+}
+
+func functionDocstringWarning(f *build.File, fix bool) []*Finding {
```
laurentlb: opt. The function is quite long. It may be useful to split it.

warn/warn_docstring.go
```diff
@@ -0,0 +1,250 @@
...
+				argumentsEnd.LineRune += len("Arguments:")
+				findings = append(findings, makeFinding(f, argumentsStart, argumentsEnd, "function-docstring",
+					`Prefer 'Args:' to 'Arguments:' when documenting function arguments.`, true, nil))
+			}
+
+			for _, param := range def.Params {
+				name := getParamName(param)
+				if !info.args[name] {
+					findings = append(findings, makeFinding(f, start, end, "function-docstring",
+						fmt.Sprintf(`Argument "%s" is not documented.`, name), true, nil))
```
laurentlb: Do we have the opposite warnning: an argument is documented, but doesn't exist

warn/warn_docstring_test.go
```diff
@@ -0,0 +1,238 @@
+package warn
```
laurentlb: We've had lots of complains inside Google about Skylint that the docstring parser was too strict or had unclear error messages, e.g. when there was an indentation problem.
Can you check that we don't have the same problems again?

warn/warn_docstring.go
vladmos: Done.

warn/warn_docstring.go
vladmos: Done.

warn/warn_docstring.go
vladmos: Done.

warn/warn_docstring.go
vladmos: The method almost only contains the emission of warnings, all the parsing logic has already been moved out. I'm not sure how I can split it further.

warn/warn_docstring.go
vladmos: Added.

warn/warn_docstring.go
vladmos: Done.

warn/warn_docstring.go
vladmos: Done.

warn/warn_docstring.go
```diff
@@ -136,18 +136,31 @@ func parseFunctionDocstring(doc *build.StringExpr) docstringInfo {
...
-				if newIndentation <= argIndentation || argIndentation == -1 {
+				if newIndentation <= argIndentation {
+					// Extract the header, e.g. "  my_arg (optional, deprecated):"
 					header := argHeaderRegex.FindString(line)
 					if len(header) > 0 {
-						arg := argRegex.FindString(header)
-						info.args[arg] = true
 						argIndentation = newIndentation
+						// Extract the arg name
+						arg := argRegex.FindString(header)
```
laurentlb: argHeaderRegex has groups (parenthesis), but you ignore them and you run a second regexp.
I think `FindStringSubmatch` on argHeaderRegex will give you the argument name.

vladmos: Done.

PR #538, opened 2019-02-07
Warnings for variable names

vladmos:
All variables should be lower_snake_case, constant should be UPPER_SNAKE_CASE, and providers should be UpperCamelCase ending with `Info`.

### DIFF
--- a/warn/warn_naming.go
+++ b/warn/warn_naming.go
@@ -68,16 +68,6 @@ func confusingNameWarning(f *build.File, fix bool) []*Finding {
 	return findings
 }
 
-// isProviderNode returns whether the node is a call of `provider()`
-func isProviderNode(expr build.Expr) bool {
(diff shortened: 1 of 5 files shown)


warn/warn_naming.go
```diff
@@ -65,3 +67,65 @@ func confusingNameWarning(f *build.File, fix bool) []*Finding {
...
+
+// isProviderNode returns whether the node is a call of `provider()`
+func isProviderNode(expr build.Expr) bool {
+	call, ok := expr.(*build.CallExpr)
+	if !ok {
+		return false
+	}
+	name, ok := call.X.(*build.Ident)
+	return ok && name.Name == "provider"
+}
```
laurentlb: I think we can skip this check.
We can simplify to: if it's UpperCamelCase, it should end with the suffix "Info".

vladmos: Removed.

PR #542, opened 2019-02-11
Fix uninitialized variables warnings

vladmos:
For-loop variables are not guaranteed to be initialized after the loop, but are initialized inside it.

### DIFF
--- a/warn/warn_control_flow.go
+++ b/warn/warn_control_flow.go
@@ -401,7 +401,16 @@ func findUninitializedVariables(stmts []build.Expr, previouslyInitialized map[st
 			// Traverse but ignore the result. Even if something is defined inside a for-loop, the loop
 			// may be empty and the variable initialization may not happen.
 			findUninitializedIdents(stmt.X, callback)
-			findUninitializedVariables(stmt.Body, initialized, callback)
+
+			// The loop can access the variables defined above, and also the for-loop variables.
+			initializedInLoop := make(map[string]bool)
+			for name := range initialized {
+				initializedInLoop[name] = true
+			}
+			for _, ident := range bzlenv.CollectLValues(stmt.Vars) {
+				initializedInLoop[ident.Name] = true
+			}
+			findUninitializedVariables(stmt.Body, initializedInLoop, callback)
 			continue
 		case *build.IfStmt:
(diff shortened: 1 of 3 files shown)


warn/warn_control_flow_test.go
```diff
@@ -641,4 +641,20 @@ def foo(y):
...
+    print(x.attr)
+
+  print(x)
+  print(x.attr)
+`,
+		[]string{
+			":6: Variable \"x\" may not have been initialized.",
+			":7: Variable \"x\" may not have been initialized.",
+		},
+		scopeEverywhere)
```
laurentlb: Add another test:

```
for x in y:
    a = x
    print(a)
print(a)
```

warn/warn_control_flow.go
```diff
@@ -401,7 +401,16 @@ func findUninitializedVariables(stmts []build.Expr, previouslyInitialized map[st
 			// Traverse but ignore the result. Even if something is defined inside a for-loop, the loop
```
laurentlb: Comment is slightly confusing:
```diff
- the loop will be empty or not
+ the collection will be empty or not
```

vladmos: Done.

warn/warn_control_flow_test.go
vladmos: Done.

allevato:
The 0.22 release should include this fix, correct? But we're still seeing the incorrect lint warnings using this version: https://travis-ci.org/bazelbuild/rules_swift/jobs/493391714

Unfortunately this has made our CI status go red. Could someone look into it and cut a new release if necessary?

vladmos:
This looks like a different bug, I'll send a fix asap.

Has it started to fail only with 0.22?

allevato:
It looks like it. We have a daily cron and the buildifier test started failing today, after the 0.22 release was cut.

It looks like there may be a couple new "legitimate" failures in the mix as well (we have to use the legacy provider syntax in a couple places because `instrumented_files` doesn't have a declared provider counterpart yet), so I'll have to see if I can selectively disable those.

vladmos:
FYI if you don't need a warning you can exclude it by providing `--warnings=-rule-impl-return`.

Issue #143, opened 2017-09-18
Remove dependency on io_bazel for build_proto


jayconrod:
I'd like to be able to use the `edit` package in Gazelle, but it transitively depends on `@io_bazel//src/main/protobuf:srcs` via `build_proto`. That's a large download (~95MB) just to copy one proto, and I'd prefer not to pull that in.

How bad would it be to check in a copy of build.proto and maybe have a test (that downstream rules wouldn't depend on) to verify it matches?

Looks like there are a couple other proto files in packages (deps_proto, extra_actions_base_proto) that could use the same treatment.

pmbethe09:
I see, so reference those protos via _test rules, but not in the _library rules to avoid download if not running tests?

jayconrod:
Right, exactly.

I'd be happy with any solution that lets us depend on the `edit` library without download Bazel though.

pmbethe09:
We do have build.gen.pb.go there for pure go users -- so you could apply your own BUILD.bazel file which is just a go_library

keith:
This can be closed by https://github.com/bazelbuild/buildtools/pull/545

PR #545, opened 2019-02-13
Remove dependency on bazel

### LINKED ISSUE #526: `buildifier` rule fetches the whole of https://github.com/bazelbuild/bazel.git
I sadly don’t have a CLI output because bazel makes it vanish after the fetch is complete (and you can’t copy auto-updating terminal).

Our rule looks like this:
```
# Run this to fix the errors in BUILD files.
buildifier(
    name = "buildifier-fix",
    mode = "fix",
    verbose = True,
)
```

and it starts fetches the bazel repo.

The whole repo takes >200s on our office connection and is basically impossible to wait on on a less potent one (>5 minutes on 1Mb/s). That’s bad.

My checkout of bazel is 750Mb, it’s a bit silly to download all that just to lint a bit of pseudo-python code.

vladmos:
The protofiles are copied from bazelbuild/bazel to avoid depending on a heavy repository.

Fixes #526


extra_actions_base_proto/BUILD.bazel
```diff
@@ -7,14 +7,6 @@ go_proto_checkedin_test(
     src = "extra_actions_base.gen.pb.go",
 )
 
-genrule(
```
pmbethe09: Should we consider leaving a version of this that is not in the main codepath, but makes it easy to 're-vendor'?

extra_actions_base_proto/BUILD.bazel
```diff
@@ -7,14 +7,6 @@ go_proto_checkedin_test(
...
-    name = "copy_and_fix",
-    srcs = ["@io_bazel//src/main/protobuf:srcs"],
-    outs = ["extra_actions_base.proto"],
-    cmd = ("SRCS=($(locations @io_bazel//src/main/protobuf:srcs));" +
-           "cp $$(dirname $$SRCS)/extra_actions_base.proto $@"),
-)
-
 proto_library(
     name = "blaze_proto",
     srcs = ["extra_actions_base.proto"],
```
pmbethe09: Now that it is vendored -- should call out via comment the exact bazel version/commit that this came from

deps_proto/BUILD.bazel
```diff
@@ -7,14 +7,6 @@ go_proto_checkedin_test(
...
-    name = "copy_and_fix",
-    srcs = ["@io_bazel//src/main/protobuf:srcs"],
-    outs = ["deps.proto"],
-    cmd = ("SRCS=($(locations @io_bazel//src/main/protobuf:srcs));" +
-           "cp $$(dirname $$SRCS)/deps.proto $@"),
-)
-
 proto_library(
     name = "blaze_deps_proto",
     srcs = ["deps.proto"],
```
pmbethe09: same comments as below -- add version/commit as a comment, should we leave a variation on copy_and_fix that writes to 'deps.proto.vendor' and then a shell script that can be run to clone, copy and replace the vendored files?

ash2k:
This also fixes https://github.com/bazelbuild/buildtools/issues/143

keith:
Any update on this?

extra_actions_base_proto/BUILD.bazel
vladmos: Do you have an example of what you mean? How do I remove io_bazel from the external dependencies but still make it easy to re-vendor?

extra_actions_base_proto/BUILD.bazel
vladmos: Added a comment.

PR #546, opened 2019-02-14
Add support for lint command line args with bazel

keith:
When running with bazel this allows you to pass arguments to enable /
disable lint rules

### DIFF
--- a/buildifier/def.bzl
+++ b/buildifier/def.bzl
@@ -11,6 +11,10 @@ def _buildifier_impl(ctx):
 
     if ctx.attr.lint_mode:
         args.append("-lint=%s" % ctx.attr.lint_mode)
+    if ctx.attr.lint_args:
+        if not ctx.attr.lint_mode:
+            fail("Cannot pass 'lint_args' without a 'lint_mode'")
+        args.extend(ctx.attr.lint_args)
     if ctx.attr.multi_diff:
         args.append("-multi_diff")
     if ctx.attr.diff_command:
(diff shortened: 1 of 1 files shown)


vladmos:
Are there any other usages for `lint_args` except for passing the `--warnings=...` argument? Perhaps replace it with `warnings` to keep the 1:1 relation between rule and command-line arguments?

keith:
Not right now. I can make that change if you'd like. I only did it this way to keep it more generic for the future.

If you do prefer that way are you thinking a list of warnings where users prepend the +/- or 2 lists for included and excluded?


--
Keith Smiley

vladmos:
I don't like the current way because it's possible to pass any arguments to `lint_args`, not necessarily `--warnings`.

About the second question, perhaps it's better to pass just one list of strings, especially because it can be either `+foo,+bar,-baz` (all prepended with `+` or `-`, meaning "default list + something - something") or `foo,bar,baz` (none prepended with `+` or `-`, meaning "just these categories"). I prefer keeping the logic in one place than partially duplicating it in the rules.

keith:
I've updated this to take a list of warnings prefixed with +/-. Let me know what you think!

buildifier/def.bzl
```diff
@@ -58,6 +62,10 @@ _buildifier = rule(
             doc = "Linting mode",
             values = ["", "fix", "warn"],
         ),
+        "lint_warnings": attr.string_list(
+            allow_empty = True,
+            doc = "A list of warnings to pass to the linter, prefixed with +/- depending on if you want to include or exclude",
```
vladmos: This is not exactly correct, maybe replace with something like

"all prefixed with +/- if you want to include in or exclude from the default set of warnings, or none prefixed with +/- if you want to override the default set, or "all" for all avaliable warnings"

keith: Ah yea, updated with your string

keith:
Thanks!

Issue #550, opened 2019-02-14
uninitialized for list comp if the same variable name is used later


thomasvl:
Given:

```python
# testing.bzl
"""module comment"""

def foo(
    deps = []):
    """Comment

    Args:
        deps: The dependencies of the target

    Returns:
        String
    """
    objc_providers = [dep[apple_common.Objc] for dep in deps if apple_common.Objc in dep]

    abc = []
    for dep in deps:
        if SwiftCcLibsInfo in dep:
          abc.append(dep[SwiftCcLibsInfo].libraries)

    return struct(
       foo = objc_providers,
       bar = abc,
    )
```

Results in:

```
testing.bzl:13: uninitialized: Variable "dep" may not have been initialized. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#uninitialized)
testing.bzl:13: uninitialized: Variable "dep" may not have been initialized. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#uninitialized)
testing.bzl:13: uninitialized: Variable "dep" may not have been initialized. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#uninitialized)
```

Removing the loop to build up `abc` makes it go away.

thomasvl:
@vladmos fyi

vladmos:
I'll temporarily disable the check for comprehensions for now to fix the breakage, in general it can be quite complicated: comprehensions can be nested, the variables defined in comprehensions are not available after them, and also available inside partially:

    [
        y  # y is defined here
        for x in y  # but not here
    ]

It may require some tome to handle such situations correctly.

PR #552, opened 2019-02-14
Fix wrong unitialized warnings

### LINKED ISSUE #549: uninitialized false positive between struct fields and future variables.
Given:

```python
# testing.bzl
"""module comment"""

def foo(
    something = False):
    """Declares output files (and optional output file map) for a compile action.

    Args:
        something: a boolean

    Returns:
        A `struct`
    """
    if not something:
        return struct(
            abc = [],
        )
    
    abc = ["something"]

    return struct(
        abc = abc,
    )
```

Results in:

```
testing.bzl:15: uninitialized: Variable "abc" may not have been initialized. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#uninitialized)
```

That's the first return getting the error, it see the `abc` with the struct as the problem.

vladmos:
Fixes #549

Workaround for #550: the check is temporarily disabled for comprehensions.

### DIFF
--- a/warn/warn_control_flow.go
+++ b/warn/warn_control_flow.go
@@ -376,14 +376,33 @@ func findUninitializedVariables(stmts []build.Expr, previouslyInitialized map[st
 	// findUninitializedIdents traverses an expression (simple statement or a part of it), and calls
 	// `callback` on every *build.Ident that's not mentioned in the map of initialized variables
 	findUninitializedIdents := func(expr build.Expr, callback func(ident *build.Ident)) {
+		// Collect lValues, they shouldn't be taken into account
+		// For example, if the expression is `a = foo(b = c)`, only `c` can be an unused variable here.
+		lValues := make(map[*build.Ident]bool)
 		build.Walk(expr, func(expr build.Expr, stack []build.Expr) {
-			if ident, ok := expr.(*build.Ident); ok && !initialized[ident.Name] {
+			if bin, ok := expr.(*build.BinaryExpr); ok && bin.Op == "=" {
+				for _, ident := range bzlenv.CollectLValues(bin.X) {
+					lValues[ident] = true
+				}
+			}
+		})
+
+		build.Walk(expr, func(expr build.Expr, stack []build.Expr) {
+			// TODO: traverse comprehensions properly
+			for _, node := range stack {
+				if _, ok := node.(*build.Comprehension); ok {
+					return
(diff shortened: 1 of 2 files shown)


warn/warn_control_flow.go
```diff
@@ -376,14 +376,33 @@ func findUninitializedVariables(stmts []build.Expr, previouslyInitialized map[st
 	// findUninitializedIdents traverses an expression (simple statement or a part of it), and calls
 	// `callback` on every *build.Ident that's not mentioned in the map of initialized variables
 	findUninitializedIdents := func(expr build.Expr, callback func(ident *build.Ident)) {
+		// Collect lValues, they shouldn't be taken into account
+		// For example, if the expression is `a = foo(b = c)`, only `c` can be an unused variable here.
+		lValues := make(map[*build.Ident]bool)
 		build.Walk(expr, func(expr build.Expr, stack []build.Expr) {
-			if ident, ok := expr.(*build.Ident); ok && !initialized[ident.Name] {
+			if bin, ok := expr.(*build.BinaryExpr); ok && bin.Op == "=" {
```
laurentlb: It would be useful to have an AssignStmt in the future, to avoid this kind of confusion.

warn/warn_control_flow_test.go
```diff
@@ -669,4 +669,57 @@ def foo():
...
+
+	checkFindings(t, "uninitialized", `
+def f():
+    if foo:
+        x = foo
+
+    f(x = y)
+`,
+		[]string{},
+		scopeEverywhere)
```
laurentlb: Add test:

```
if True:
  f(x = 3)
else:
  x = 3
print(x)  # warning
```

warn/warn_control_flow.go
vladmos: I agree, but this case also covers assignment expressions (as for keyword arguments). Buildifier should warn about uninitialized `x` in `f(x = y)`.

Perhaps I'll create an AssignmentExpr node (we don't have separate statements nodes anyway).

warn/warn_control_flow_test.go
vladmos: Done.

warn/warn_control_flow.go
laurentlb: In `f(x = y)`, we shouldn't warn about x, only y.

PR #559, opened 2019-02-25
Fix for legacy named arguments

vladmos:
Replaces legacy keyword arguments with positional arguments in some builtin functions, e.g. `bool(x = foo)` -> `bool(foo)`.

Fixes https://github.com/bazelbuild/bazel/issues/5010

### DIFF
--- a/WARNINGS.md
+++ b/WARNINGS.md
@@ -407,7 +407,7 @@ or at the beginning of a rule.
   * Category_name: `keyword-parameters`
   * Automatic fix: yes
 
-Some parameters for builtin functions in Skyalrk are keyword for legacy reasons,
+Some parameters for builtin functions in Starlark are keyword for legacy reasons,
 their names are not meaningful (e.g. `x`), making them positional will improve
 the readability.
 
--- a/WARNINGS.md
+++ b/WARNINGS.md
@@ -407,8 +407,8 @@ or at the beginning of a rule.
   * Category_name: `keyword-parameters`
   * Automatic fix: yes
 
-Some parameters for builtin functions in Starlark are keyword for legacy reasons,
-their names are not meaningful (e.g. `x`), making them positional will improve
+Some parameters for builtin functions in Starlark are keyword for legacy reasons;
+their names are not meaningful (e.g. `x`). Making them positional will improve
 the readability.
 
 --------------------------------------------------------------------------------
--- a/WARNINGS.md
+++ b/WARNINGS.md
@@ -20,6 +20,7 @@ Warning categories supported by buildifier's linter:
   * [git-repository](#git-repository)
   * [http-archive](#http-archive)
   * [integer-division](#integer-division)
+  * [keyword-parameters](#keyword-parameters)
   * [load](#load)
   * [load-on-top](#load-on-top)
   * [module-docstring](#module-docstring)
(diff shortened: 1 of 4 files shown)


laurentlb:
cc @alandonovan, fyi

WARNINGS.md
```diff
@@ -401,6 +402,17 @@ or at the beginning of a rule.
 
 --------------------------------------------------------------------------------
 
+## <a name="keyword-parameters"></a>Keyword parameter should be positional
+
+  * Category_name: `keyword-parameters`
+  * Automatic fix: yes
+
+Some parameters for builtin functions in Skyalrk are keyword for legacy reasons,
```
alandonovan: typo: Skylark.
but really: Starlark.

WARNINGS.md
```diff
@@ -401,6 +402,17 @@ or at the beginning of a rule.
...
+their names are not meaningful (e.g. `x`), making them positional will improve
```
alandonovan: The two commas in this sentence are ungrammatical. 
Replace the first with ";" or "and", and the second with a period and capital.

warn/warn_bazel_api.go
```diff
@@ -610,3 +610,58 @@ func ruleImplReturnWarning(f *build.File, fix bool) []*Finding {
...
+	"len": "x",
+	"str": "x",
+	"repr": "x",
+	"bool": "x",
+	"int": "x",
+	"dir": "x",
+	"type": "x",
+	"hasattr": "x",
+	"getattr": "x",
+	"select": "x",
```
alandonovan: Most often I see string.split(maxsplit=-1) or dict.get(key, default=...).  Be sure to handle methods too.

WARNINGS.md
vladmos: Thanks, fixed.

laurentlb:
This change looks limited to 1-arg functions?
What about these cases?

```python
int(2, base = 3)
range(2, 5, step = 2)
```

WARNINGS.md
vladmos: Done.

vladmos:
It's not limited to 1-arg functions, but for now only converts the first argument (see `getattr` and `hasattr` in the tests), it just only converts the first argument. As we haven't agreed yet which parameters to make positional only, I've only implemented the most obvious fixes (e.g. a parameter called `x` never improves readability of the code).

I'll update the PR (or send a separate if this one gets merged) once we finalize the list of parameters to fix.

warn/warn_bazel_api.go
vladmos: I've collected the potential cases here: https://github.com/bazelbuild/bazel/issues/5010#issuecomment-466472598

I'll implement more fixes once we agree on which keyword parameters we want like to abandon.

alandonovan: To repeat what I said on that issue, the important distinction is between core Starlark (e.g. getattr, splitlines) and the build language (e.g. fail, aspect). In the core, following Python, very few functions accept named arguments; int(base=...) is one of the rare exceptions. In the build language, and for user-defined macros, it's much more common for functions to accept or even require named arguments. Let me know if you disagree.

laurentlb:
What's the status of this?

vladmos:
I think I deleted the branch by mistake and Github closed the pull request. But it shouldn't be hard to restore.

Is there a decision which arguments we're going to make positional-only?

laurentlb:
Once we have a way to say: "This argument is always positional" or "This argument is always named", we can extend the list of functions to fix.

For the core Starlark builtins, I think the only one that can be named is `base` in `int(x, base = 2)`. In Blaze functions, we can make the `exclude` more explicit:
```diff
- glob(["*.cc"], ["test*])
+ glob(["*.cc"], exclude = ["test*])
```

vladmos:
Reopened as https://github.com/bazelbuild/buildtools/pull/692

alandonovan:
> For the core Starlark builtins, I think the only one that can be named is base in int(x, base = 2). In Blaze functions, we can make the exclude more explicit:

I think the following built-in functions allow some named arguments: print, sorted, fail, int, min, max.
We should make the docs more explicit.

PR #560, opened 2019-02-25
Introduce separate Bzl and Default filetypes

vladmos:
Previously buildifier distinguished BUILD, WORKSPACE, and all other files (the latter were .bzl files). However buildifier can be applied to other types of Starlark files (e.g. copybara configuration files), and its bzl-specific lints shouldn't be applied to them.

This commit introduces TypeBzl reserved for Bazel's .bzl files, and fixes some checks (mostly the checks for Bazel build API usages) to not be applied to generic Starlark files.

### DIFF
--- a/build/lex.go
+++ b/build/lex.go
@@ -33,22 +33,26 @@ import (
 type FileType int
 
 const (
-	// TypeDefault represents .bzl or other Starlark files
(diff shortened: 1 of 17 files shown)


vladmos:
Updated a test, please re-approve.

PR #564, opened 2019-02-26
Introduce an AssignmentExpr node

vladmos:
Assignment expressions are special cases of binary expressions with operator == "=", they are often treated specially in various lint checks: as statements they define new (local) variables, as expressions inside other statements they correspond to keyword parameters.

### DIFF
--- a/build/syntax.go
+++ b/build/syntax.go
@@ -366,6 +366,7 @@ type AssignmentExpr struct {
 	Comments
 	X         Expr
 	OpStart   Position
+	Op        string
 	LineBreak bool // insert line break between Op and Y
 	Y         Expr
 }
--- a/build/syntax.go
+++ b/build/syntax.go
@@ -361,6 +361,21 @@ func (x *BinaryExpr) Span() (start, end Position) {
 	return start, end
 }
 
+// An AssignmentExpr represents a binary expression with `=`: X = Y.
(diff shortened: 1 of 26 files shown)


build/syntax.go
```diff
@@ -361,6 +361,22 @@ func (x *BinaryExpr) Span() (start, end Position) {
...
 
+// An AssignmentExpr represents a binary expression with `=`: X = Y.
+type AssignmentExpr struct {
+	Comments
+	X         Expr
+	OpStart   Position
+	Op        string
+	LineBreak bool // insert line break between Op and Y
+	Y         Expr
+}
```
laurentlb: Use the same definition as https://github.com/google/starlark-go/blob/master/syntax/syntax.go?

vladmos: I'll rename the fields, but the nodes won't be equivalent: in stalark-in-go keyword parameters are still BinaryExpr, in this PR they become AssignmentExpr. I did that intentionally, the linter often needs to distinguish between random binary expressions (e.g. `a + b`) and keyword parameters(`f(a = b)`). This also technically affects some weird cases that don't exist in real code (and can't be generated by the parser) like `f(a += b)`.

vladmos: Changed the definition.

Issue #565, opened 2019-02-27
Adding a non-string to a list


WillerZ:
It seems that [the logic for the `add` command](https://github.com/bazelbuild/buildtools/blob/34d7371037136c478031f03a11692fb16ec5e11a/edit/buildozer.go#L81-L92) to buildozer adds `"` quotes around the item to be added unless it knows that the list it's adding to is a list of ints.

Can you add an `add_raw` command for the (admittedly rare) scenario that someone wants to add a non-string item to a list. For my use-case all of the `add`s I want to do are of this nature, so a commandline flag to change the behavior of `add` for the duration of one invocation would also work for me.

adzenith:
I am running into this issue right now. I'm trying to add `requirement("package_name")` to a `deps` list for use with [rules_python](https://github.com/bazelbuild/rules_python), but it ends up like:
```
$ buildozer 'add deps foo("bar")' //:target
fixed BUILD
$ buildozer 'print deps' //:target
[
    "foo(\"bar\")",
]

vladmos:
The problem is that buildozer knows that `deps` contains a list of strings and converts any input regardless of whether it was quoted or not to a string literal.

If I change that to "don't use string literals unless the input is quoted" that will be a breaking change, many scripts that use buildozer will be broken.

What do you think about the following syntax:

    buildozer 'add deps `foo("bar")`' //:target

to emphasize that the argument should be inserted as is?

adzenith:
Yeah, that could work - or the `add_raw` command suggested in the initial comment. Right now I’m working around the issue by adding aliases for the requirements that I’m using so that the aliases are just strings.

WillerZ:
We are using `add` with a reasonably unique suffix and then `sed` to fix the results.

wyattanderson:
This would be a really positive change for me and would eliminate a lot of hacky workarounds. Either the `_raw` variants or the backticks would be fine; I think a command-line flag that was applied to all options would work as well—I could modify my tools to add quotes where necessary.

adzenith:
I would still love to be able to do this

PR #567, opened 2019-02-28
Introduce an add_raw command

WillerZ:
Resolves: #565

Some of the build-rule functions used within Facebook are higher-order functions that accept lists of parameters which are not basic strings. This change to buildozer allows me to programmatically insert new values of this type into those rules without introducing duplicates.

### DIFF
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -78,13 +78,20 @@ type CmdEnvironment struct {
 
 // The cmdXXX functions implement the various commands.
 
-func cmdAdd(opts *Options, env CmdEnvironment) (*build.File, error) {
+func cmdAddRaw(opts *Options, env CmdEnvironment) (*build.File, error) {
 	attr := env.Args[0]
 	for _, val := range env.Args[1:] {
-		if IsIntList(attr) {
-			AddValueToListAttribute(env.Rule, attr, env.Pkg, &build.Ident{Name: val}, &env.Vars)
-			continue
-		}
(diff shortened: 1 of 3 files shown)


WillerZ:
I should be an authorized contributor covered by our corporate agreement

WillerZ:
I really am an authorized contributor and I should be covered by our corporate agreement

laurentlb:
I wonder if there can be a different interface, without introducing a new command. I expect other people will need `set_raw`, `set_raw_if_absent`, `replace_raw`, `dict_add_raw`, `dict_set_raw`.

WillerZ:
@laurentlb You're right of course, and I expect I'll need most of those sooner or later.

I tried (quite briefly I admit) and couldn't come up with something else that would be perfectly backwards compatible and not risk breaking existing scripts. Best I came up with is some "magic" prefix to the string parameter like `^` to mean "this is raw" but I don't have enough context on other people's builds to know if that character choice would be safe enough.

vladmos:
Would it make sense to add a new flag (like `--raw_input` that would be applied to all commands (`add`, `set`, `set_if_absent`, `replace`, `dict_add`, `dict_set`)?

WillerZ:
I'm abandoning this PR.

The workaround we're using is to use `add` to insert `"wanted_part_REMOVE_THE_QUOTES_AND_THIS_MESSAGE"` and chasing that with two perl one-liners that remove the duplicate entries and the quotes.

Issue #568, opened 2019-03-03
Format different for file vs stdin


keith:
With a bzl file with this contents:

```python
def _foo_impl(ctx):
    pass

foo = rule(
    fragments = ["apple"],
    implementation = _foo_impl,
    attrs = {},
)
```

Then running this command:

```
buildifier foo.bzl
cat foo.bzl | buildifier > bar.bzl
diff foo.bzl bar.bzl
```

You end up with this diff:

```diff
--- foo.bzl     2019-03-02 17:12:52.038689963 -0800
+++ bar.bzl     2019-03-02 17:13:04.429933521 -0800
@@ -2,7 +2,7 @@
     pass

 foo = rule(
+    attrs = {},
     fragments = ["apple"],
     implementation = _foo_impl,
-    attrs = {},
 )
```

I installed buildifier with `go get` so the version is redacted, but I just updated so it should be 4bcdbd1064fcc48180fa30400e39f7a940fdb8f9

vladmos:
That is intentional for compatibility, but eventually it should format string as bzl files because their format is more flexible.

Now that there's the "default" file type for generic Starlark files (not necessarily for Bazel), buildifier should mark parsed files that come to stdin as default, they are the same as bzl in terms of formatting but some bzl-specific lints won't be applied to them.

keith:
Ah makes sense, thanks for the context, here's a PR to do that https://github.com/bazelbuild/buildtools/pull/569

thomasvl:
Shouldn't the `-type` also come into play?  i.e. - no matter what you send to stdin, you really should use `-type` to indicate if you expect it to format a BUILD/bzl/WORKSPACE?

keith:
If you pass type this does work as expected

vladmos:
Correct, stdin should be treated as default Starlark file (unless `--type` is provided explicitly), files should be parsed based on their filename/extension (also unless `--type` is provided explicitly).

PR #569, opened 2019-03-03
Set inputType to default for stdin

keith:
In the case the user doesn't pass a specific inputType for files passed
with stdin we now change the inputType to default

Fixes https://github.com/bazelbuild/buildtools/issues/568


vladmos:
That's not how it should be done, `--type=auto` means that buildifier should detect the file type based on its name/extension, it shouldn't force the default type in this case.

This line should be changed instead: https://github.com/bazelbuild/buildtools/blob/master/build/lex.go#L110

keith:
Updated!

keith:
Ah sorry about that, updated

PR #572, opened 2019-03-05
buildozer: Better handle Bazel's "glob" helper function

thundergolfer:
PR https://github.com/bazelbuild/buildtools/pull/100 allowed for setting basic `glob(...)`s with the `set` command, but I experienced a failure when attempting to set a more complex example that is close to the full function signature:  

```python
glob(include, exclude=[], exclude_directories=1)
```

The spaces in the function cause the program to produce undesired behaviour. Problems begin when `SplitOnSpaces(arg)` is done. 

The solution in this PR is not great, but it at least improves `buildozer`'s handling of glob.


thundergolfer:
Closing to refactor

PR #581, opened 2019-03-15
Enable the last test on Windows

meteorcloudy:


meteorcloudy:
The blocker for this test bazelbuild/bazel#7222 has been released in 0.23.0

PR #587, opened 2019-03-26
Find relevant Starlark files recursively in directories

### LINKED ISSUE #516: Buildifier: allow to run on a directory (+ find files recursively)
This is a feature request sparked in PR #495 by @brown.

Currently, `buildifier` only operates on single files. In order to check a whole project with plenty of Bazel-related files, it requires users to 'feed' it all relevant files. It would be awesome to have `buildifier` have the ability to ...:

* find files it understands automatically (`BUILD`, `BUILD.bazel`, `WORKSPACE`, `*.bzl`, etc.)
* recurse into subdirectories.

vladmos:
If an argument to buildifier is not a file but a directory, recursively find all relevant Starlark files in it (i.e. with extensions `.bzl`, `.sky`, `.build`, or `.workspace`, or with names `build.*` or `workspace.*`, case insensitively).

Fixes #516.

### DIFF
--- a/buildifier/buildifier.go
+++ b/buildifier/buildifier.go
@@ -70,7 +70,7 @@ func stringList(name, help string) func() []string {
 }
 
 func usage() {
-	fmt.Fprintf(flag.CommandLine.Output(), `usage: buildifier [-d] [-v] [-diff_command=command] [-help] [-multi_diff] [-mode=mode] [-lint=lint_mode] [-path=path] [files...]
+	fmt.Fprintf(flag.CommandLine.Output(), `usage: buildifier [-d] [-v] [-r] [-diff_command=command] [-help] [-multi_diff] [-mode=mode] [-lint=lint_mode] [-path=path] [files...]
 
 Buildifier applies standard formatting to the named Starlark files.  The mode
 flag selects the processing: check, diff, fix, or print_if_changed.  In check
--- a/buildifier/buildifier.go
+++ b/buildifier/buildifier.go
@@ -266,7 +266,7 @@ func main() {
 			files, err = ExpandDirectories(args)
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "buildifier: %v\n", err)
-				os.Exit(1)
+				os.Exit(3)
 			}
 		}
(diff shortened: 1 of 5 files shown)


buildifier/buildifier.go
```diff
@@ -292,6 +292,13 @@ func processFiles(files []string, inputType, lint string, warningsList []string)
 		data []byte
 		err  error
 	}
+
+	files, err := expandDirectories(files)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "buildifier: %v\n", err)
+		os.Exit(1)
```
pmbethe09: style wise is it usually cleaner if only 'main' is allowed to os.Exit

here, return fmt.Errorf(...)
[changing the function signature to return error]

vladmos: Done (moved the block out of the processFiles function).

buildifier/integration_test.sh
```diff
@@ -15,11 +15,13 @@ INPUT="load(':foo.bzl', 'foo'); foo(tags=['b', 'a'],srcs=['d', 'c'])"  # formatt
 echo -e "$INPUT" > test/build  # case doesn't matter
 echo -e "$INPUT" > test/test.bzl
 echo -e "$INPUT" > test/subdir/test.bzl
+echo -e "$INPUT" > test.bzl  # outside the test directory
 echo -e "not valid +" > test/foo.bar
 cp test/foo.bar golden/foo.bar
 
 "$buildifier" < test/build > stdout
-"$buildifier" test
+"$buildifier" -r test
```
laurentlb: Do you have a test for `$buildifier" test`?

vladmos: Added a test

vladmos: The test fails because the error message is "is a directory" on Linux/Mac and "The handle is invalid." on Windows. Is there a simple workaround (e.g. a way to specify different test strings on different platforms)?

laurentlb: The test is not so important (matching against the system provided string is brittle).
Maybe just check the error code.

PR #590, opened 2019-03-28
Provide the functionality to disable some findings of duplicated-name

vladmos:


laurentlb:
I'd prefer to avoid adding this.
Users should first try to fix their code. :)

vladmos:
I've put this CL on hold, will probably discard it.

PR #598, opened 2019-04-04
Allow --lint=warn to be combined with any value of --mode

### LINKED ISSUE #453: Do not reformat with --lint=warn
```bash
$ echo 'x=1' > /tmp/BUILD

$ buildifier --lint=warn /tmp/BUILD
/tmp/BUILD:1: unused-variable: Variable "x" is unused. Please remove it.
To disable the warning, add '@unused' in a comment. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#unused-variable)

$ cat /tmp/BUILD
x = 1
```

When used as a linter, buildifier shouldn't modify the files.

vladmos:
Currently it's not allowed to combine `--lint=warn` with any value of `--mode` except `--mode=fix` (which is default). It makes it not possible to print lint warnings without reformatting a file.

`--lint=fix` still requires `--mode=fix` because applying fixes involves reformatting.

Fixes #453

### DIFF
--- a/buildifier/buildifier.go
+++ b/buildifier/buildifier.go
@@ -129,92 +129,15 @@ func main() {
 	build.DisableRewrites = disable()
 	build.AllowSort = allowSort()
 
-	// Check input type.
-	switch *inputType {
-	case "build", "bzl", "workspace", "default", "auto":
-		// ok
-
-	default:
-		fmt.Fprintf(os.Stderr, "buildifier: unrecognized input type %s; valid types are build, bzl, workspace, default, auto\n", *inputType)
+	if err := utils.ValidateModes(inputType, mode, lint, dflag); err != nil {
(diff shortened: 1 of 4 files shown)


laurentlb:
Is there any test to add/update or documentation to update?

Will `buildifier --lint=warn myfile` reformat the file?

vladmos:
> Is there any test to add/update or documentation to update?

I'll add a test to integration_test.sh. It wasn't mentioned in the documentation that `--lint=warn` required `--mode=fix`, so nothing to update.

> Will `buildifier --lint=warn` myfile reformat the file?

Yes, the default value for `--mode` is still `fix`. Perhaps I can add deprecate the `--lint` flag by adding `--mode=lint` as an alias for `--mode=print_if_changed --lint=warn` but I'll do that in a separate CL.

PR #600, opened 2019-04-04
Fix golint warnings

vladmos:


buildifier/utils/flags.go
```diff
@@ -95,4 +97,4 @@ func ValidateWarnings(warnings *string) ([]string, error) {
 		}
 	}
 	return warningsList, nil
-}
+}
```
laurentlb: Was the file formatted with gofmt?

vladmos: Fixed.

PR #605, opened 2019-04-05
Move some functionality from buildifier.go to utils.go

vladmos:
The goal is to make different implementations of `buildifier.go` having as less repeated code as possible. This CL moves some of the methods implemented in `buildifier.go` to `buildifier/utils` so that it can be reused.

There will be more follow-up CLs.

### DIFF
--- a/buildifier/utils/flags.go
+++ b/buildifier/utils/flags.go
@@ -5,17 +5,19 @@ import (
 	"strings"
 )
 
-// ValidateModes validates flags --type, --mode, --lint, and -d
-func ValidateModes(inputType, mode, lint *string, dflag *bool) error {
-	// Check input type.
+// ValidateModes validates the value of --type
+func ValidateInputType(inputType *string) error {
 	switch *inputType {
 	case "build", "bzl", "workspace", "default", "auto":
-		// ok
+		return nil
 
 	default:
 		return fmt.Errorf("unrecognized input type %s; valid types are build, bzl, workspace, default, auto", *inputType)
 	}
+}
 
+// ValidateModes validates flags --mode, --lint, and -d
+func ValidateModes(mode, lint *string, dflag *bool, additionalModes ...string) error {
 	if *dflag {
 		if *mode != "" {
 			return fmt.Errorf("cannot specify both -d and -mode flags")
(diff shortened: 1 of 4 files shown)


buildifier/utils/utils.go
```diff
@@ -68,3 +72,47 @@ func GetParser(inputType string) func(filename string, data []byte) (*build.File
...
+// WriteTemp writes data to a temporary file and returns the name of the file.
+func WriteTemp(data []byte, toRemove *[]string) (file string, err error) {
+	f, err := ioutil.TempFile("", "buildifier-tmp-")
+	if err != nil {
+		return "", fmt.Errorf("creating temporary file: %v", err)
+	}
+	name := f.Name()
+	*toRemove = append(*toRemove, name)
+	defer f.Close()
+	_, err = f.Write(data)
```
pmbethe09: push into if, use :=

buildifier/utils/utils.go
```diff
@@ -68,3 +72,47 @@ func GetParser(inputType string) func(filename string, data []byte) (*build.File
...
+
+// WriteTemp writes data to a temporary file and returns the name of the file.
+func WriteTemp(data []byte, toRemove *[]string) (file string, err error) {
+	f, err := ioutil.TempFile("", "buildifier-tmp-")
+	if err != nil {
+		return "", fmt.Errorf("creating temporary file: %v", err)
+	}
+	name := f.Name()
+	*toRemove = append(*toRemove, name)
+	defer f.Close()
```
pmbethe09: by convention, this goes right after the '}' where you checked for error

buildifier/utils/utils.go
```diff
@@ -68,3 +72,47 @@ func GetParser(inputType string) func(filename string, data []byte) (*build.File
 		return build.ParseDefault
 	}
 }
+
+// WriteTemp writes data to a temporary file and returns the name of the file.
+func WriteTemp(data []byte, toRemove *[]string) (file string, err error) {
```
pmbethe09: C++ style "out" arguments like toRemove* are strongly discouraged.  

Why not add an extra return value?

buildifier/utils/flags.go
```diff
@@ -24,22 +26,22 @@ func ValidateModes(inputType, mode, lint *string, dflag *bool) error {
...
-		*mode = "fix"
-
-	case "check", "diff", "fix", "print_if_changed":
-		// ok
+	validModes := []string{"check", "diff", "fix", "print_if_changed"}
+	validModes = append(validModes, additionalModes...)
 
-	default:
-		return fmt.Errorf("unrecognized mode %s; valid modes are check, diff, fix", *mode)
+	recognizedMode := false
```
pmbethe09: factoring these sort of linear search into a standalone method returning bool usually reads cleaner.

But why not just make validModes a map[string]bool, and then "append" by looping and adding to the set?

Then here just do 'set' membership?

vladmos: I tried to, but it makes the order of the valid arguments in the error message randomized.

buildifier/utils/utils.go
vladmos: Removed the argument, delegated the logic to the caller of `WriteTemp`.

buildifier/utils/utils.go

buildifier/utils/utils.go
vladmos: Done (I've just copied the code here, but good to refactor it at the same time)

PR #610, opened 2019-04-16
Fix tests in //api_proto on Windows

laszlocsomor:
Fix the Starlark rule generating
//api_proto:api.gen.pb.go_checkshtest so it works
on Windows, without a runfiles tree.

To fix, the generated shell script now uses the
Bash runfiles library to locate the files to
compare.

See https://github.com/bazelbuild/bazel/issues/5508


laszlocsomor:
Thanks @vladmos ! Could you please merge it? I have no write access.

PR #611, opened 2019-04-17
Refactor WriteTemp

vladmos:
Remove the global variable `toRemove` by replacing it with a class that keeps track of created temp files and cleans them up in the end.

### DIFF
--- a/buildifier/buildifier.go
+++ b/buildifier/buildifier.go
@@ -176,7 +176,6 @@ func main() {
 	diff = differ
 
 	tf := &utils.TempFile{}
-	defer tf.Clean()
 
 	if len(args) == 0 || (len(args) == 1 && args[0] == "-") {
 		// Read from stdin, write to stdout.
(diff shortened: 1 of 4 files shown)


buildifier/buildifier.go
```diff
@@ -175,6 +175,9 @@ func main() {
 	}
 	diff = differ
 
+	tf := &utils.TempFile{}
+	defer tf.Clean()
```
pmbethe09: defers are not run in case of os.Exit, only when a func returns

vladmos: Fixed. Although `tf.Clean()` may be not called because of `os.Exit()`, it doesn't make sense to call it anyway if `processFile`(`s`) hasn't been called.

pmbethe09: maybe the answer is to make main() a lot shorter and move most of this code into a 'run' method that has this defer, and also a 'return exitCode', where only main calls os.Exit(exitCode)

vladmos: Done. `main` is not a lot shorter though, but it only contains parsing and validating of the arguments and a call to `run()` now.

PR #612, opened 2019-04-17
Implement diagnostics in json

### LINKED ISSUE #477: Provide a buildifier option for more tooling-friendly output
I've integrated buildifier linting/formatting into https://github.com/bazelbuild/vscode-bazel, and I think there are a couple enhancements that could be made to improve lint integration with IDEs:

**A more structured output mode.** Protos would be good here, similar to how I can run `bazel query --output=proto` and get an easily consumable binary proto stream back. Right now I take the text output and scrape it for line numbers, categories, and messages, but this is a little unclean because some messages can span multiple lines. (https://github.com/bazelbuild/vscode-bazel/blob/master/src/buildifier/buildifier.ts#L159-L180)

**Finer-grained location information about where lint warnings occur.** Right now a lint warning only indicates the line number of the issue. In VS Code, I register a diagnostic that occupies (and thus, underlines) the entire line (starting from the first non-whitespace character). It would be nice to get precise start and end line/column information for warnings so that I can highlight the exact issue. For example, if one of many bindings in a `load(...)` statement is unused, I would like to highlight just the one that should be removed.

vladmos:
Users can specify `--format=json`, but that requires `--mode=check`. Can be combined with `-v` to print a readable json.

If the `--format` flag is provided (with values either `text` or `json`), diagnostics will be printed to stdout and the return code will be 0 (so that the tools know that the output is parseable). Otherwise diagnostics will be printed to stderr, as stdout is reserved for the file output.

Fixes #477


buildifier/buildifier.go
```diff
@@ -319,41 +348,22 @@ func processFile(filename string, data []byte, inputType, lint string, warningsL
...
+			fileDiagnostics.SetRewrites(info.Stats())
 			exitCode = 4
 		}
-		return
 
 	case "diff":
 		// diff mode: run diff on old and new.
 		if bytes.Equal(data, ndata) {
-			return
+			break
```
pmbethe09: instead of all these break statements (which require to to follow down to see what happens next

make these all
return fileDiagnostics, exitCode


and in the case where it was

exitCode = 3
return

use 
return fileDiagnostics, 3

vladmos: Done.

buildifier/utils/diagnostics.go
```diff
@@ -0,0 +1,137 @@
...
+			result, _ = json.MarshalIndent(*d, "", "    ")
+		} else {
+			result, _ = json.Marshal(*d)
+		}
+		return string(result) + "\n"
+	}
+	return ""
+}
+
+type FileDiagnostics struct {
```
laurentlb: Ad documentation

buildifier/utils/diagnostics.go
```diff
@@ -0,0 +1,137 @@
...
+	Filename  string         `json:"filename"`
+	Formatted bool           `json:"formatted"`
```
laurentlb: What is `Formatted`?
Is it set to true because the file has been reformatted? Is it set to true when the file was already formatted?

vladmos: Means the file is formatted correctly. Currently the diagnostics is only available with `--mode=check`, so can't mean "reformatted".

buildifier/buildifier.go
```diff
@@ -319,41 +348,21 @@ func processFile(filename string, data []byte, inputType, lint string, warningsL
...
-						last = s
-						uniq = append(uniq, s)
-					}
-				}
-				log = " " + strings.Join(uniq, " ")
-			}
-			fmt.Printf("%s #%s %s%s\n", filename, reformat, &info, log)
+			fileDiagnostics.MarkNotFormatted()
+			fileDiagnostics.SetRewrites(info.Stats())
 			exitCode = 4
```
pmbethe09: could just 
return fileDiagnostics, 4

allevato:
This is great, thanks! I'm looking forward to integrating this into vscode-bazel once it's released.

buildifier/utils/diagnostics.go
vladmos: Done.

buildifier/buildifier.go
vladmos: Done.

PR #619, opened 2019-04-24
Bug fix. Preserve inline comments within an attribute list as inline …

pdoyle12:
…comments, even when there is only one list item remaining. Do not promote them to end-of-line comments.

Example rule:
	
java_library(
    name = "testrule",
    deps = [
      "depone",
      "deptwo",  # keep
    ],
)

Previously if depone was removed by buildozer, it would reformat like this:

java_library(
    name = "testrule",
    deps = ["deptwo"],  # keep
)

moving the "keep" comment outside the list, but now preserves the comment in the list:

java_library(
    name = "testrule",
    deps = [
      "deptwo",  # keep
    ]
)

This change was requested because tools depend on comments associated with specific deps.

### DIFF
--- a/build/print.go
+++ b/build/print.go
@@ -697,7 +697,7 @@ func (p *printer) useCompactMode(start *Position, list *[]Expr, end *End, mode s
 	// If there are line comments, use multiline
 	// so we can print the comments before the closing bracket.
 	for _, x := range *list {
-		if len(x.Comment().Before) > 0 {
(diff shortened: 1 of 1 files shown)


laurentlb:
This breaks the tests: https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/ec3e02f0-00fa-4198-a81e-1d06ea6b6970/build/go_default_test/attempt_1.log

pdoyle12:
Updated; I hadn't realized the internal and external tests were non-overlapping, sorry about that.

PR #625, opened 2019-05-02
Fix golint warnigns

vladmos:


buildifier/utils/tempfile.go
```diff
@@ -6,6 +6,7 @@ import (
 	"os"
 )
 
+// Tempfile creates temporary files and cleans them up in the end
```
pmbethe09: nit: case Tempfile or TempFile -- probably the second and then be consistent.

And the doc here 'creates' reads like a func, not a struct

buildifier/utils/tempfile.go
```diff
@@ -25,8 +26,10 @@ func (tf *TempFile) WriteTemp(data []byte) (file string, err error) {
 	return name, nil
 }
 
+// Clean removes all created temporary files
```
pmbethe09: should this func be Close() error ? 

Since os.Remove returns error (that should be handled but is not currently)?

buildifier/utils/tempfile.go
vladmos: Fixed

buildifier/utils/tempfile.go
vladmos: Done

buildifier/utils/tempfile.go
```diff
@@ -27,9 +27,12 @@ func (tf *TempFile) WriteTemp(data []byte) (file string, err error) {
 }
 
 // Clean removes all created temporary files
-func (tf *TempFile) Clean() {
+func (tf *TempFile) Clean() error {
 	for _, file := range tf.filenames {
-		os.Remove(file)
+		if err := os.Remove(file); err != nil {
+			return err
```
pmbethe09: note that this leaves tf.filenames in a precarious state as some are removed and some are not.
Also, if Remove returns os.ErrNotExist -- we can probably drop that as our job is done.

vladmos: I thought about that, but the object is used and cleaned exactly once by the buildifier binary. If it fails to delete some temporary files right before exiting there's nothing meaningful to do anyway.

PR #626, opened 2019-05-02
Fix -r flag

### LINKED ISSUE #614: ExpandDirectories not working as intended?
From what I can tell the `utils.ExpandDirectories` is supposed to return a list of regular starlark files.  As currently written, it also includes inputs like:

`{WORKSPACE_ROOT}/foo/bar/build` (a directory named build)
`{WORKSPACE_ROOT}/foo/bar/workspace` (a directory named workspace)
`{WORKSPACE_ROOT}/proto/build/build.proto` (a proto file)

### LINKED ISSUE #624: buildifier -r should not scan .git directory
Buildifer tries to parse Git files such as .git/logs/refs/remotes/origin/user/HackAttack/build as BUILD files.

You could argue that this is not a bug per se, but it is definitely never the desired behavior.

vladmos:
Skip directories named as Starlark files (e.g. `build/` or `workspace/`).
Skip .git directories entirely.

Fixes #614
Fixes #624

### DIFF
--- a/buildifier/integration_test.sh
+++ b/buildifier/integration_test.sh
@@ -42,7 +42,11 @@ echo -e "$INPUT" > test_dir/subdir/test.bzl
 echo -e "$INPUT" > test.bzl  # outside the test_dir directory
 echo -e "not valid +" > test_dir/foo.bar
 mkdir test_dir/workspace  # name of a starlark file, but a directory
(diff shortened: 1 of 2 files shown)


vladmos:
Implemented skipping of .git directories.

PR #630, opened 2019-05-07
Allow setting empty arguments

### LINKED ISSUE #616: set_if_absent to empty list
If I have a rule like this:

```python
unit_test(name = "PluginsTests")
```

And I want to add an empty `deps` attribute (currently because of this issue https://github.com/bazelbuild/buildtools/issues/615 but I think this is relevant regardless)

It looks like `set_if_absent` doesn't support setting something to nothing, without warning.

If I run:

```sh
buildozer "set_if_absent deps" Modules/Plugins:PluginsTests
```

It does actually do what I want:

```python
unit_test(
    name = "PluginsTests",
    deps = [],
)
```

But it also warns with:

```
Too few arguments for command 'set_if_absent', expected at least 2.
```

And exits 0. Is this behavior expected or should it not warn in this case?

vladmos:
* Set the minimum amount of arguments to `set` and `set_if_absent` to 1 (was 2).
  * Exit prematurely if there are not enough or too many arguments (was: warn and continue the execution)
  * If `set` or `set_if_absent` are called without a value provided, set it to an empty list if a list was expected, on `None` otherwise (was: empty list or crash with runtime error).

Fixes #616

### DIFF
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -82,7 +82,7 @@ func cmdAdd(opts *Options, env CmdEnvironment) (*build.File, error) {
 	attr := env.Args[0]
 	for _, val := range env.Args[1:] {
 		if IsIntList(attr) {
-			AddValueToListAttribute(env.Rule, attr, env.Pkg, &build.Ident{Name: val}, &env.Vars)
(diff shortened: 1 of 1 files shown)


vladmos:
I was thinking about setting an empty string if a string is expected but nothing is provided, but a user can always use `'set foo ""'` for that.

Issue #375, opened 2018-09-16
Release buildtools for Windows as well


oferb:
The repo releases:
https://github.com/bazelbuild/buildtools/releases

only contain Linux and Mac binaries. Would be nice to also have Windows there :-)

laszlocsomor:
+1

I successfully built Buildozer and Buildifier at https://github.com/bazelbuild/buildtools/commit/220349f891794f407450d6eada3a2cddf370040a using Bazel 0.18.0 on Windows.

pmbethe09:
I would be very happy to include Windows builds as part of an automated release process.
However we don't have one yet -- no one has had the time to follow up on this.
Right now I do linux/Mac builds manually and don't have access to a Windows platform to do that build.

oferb:
What constitutes a release process? Just building it, or are there additional steps?

oferb:
BTW - for my particular use, I'm going for WSL - Windows Subsystem for Linux, so currently I don't support plain Windows for https://github.com/google/startup-os.

laurentlb:
cc @philwo for any recommendation (is it easy to automate? could it be part of the Bazel release process?)
cc @aiuto

philwo:
This is difficulty is that we only have two sets of machines where we can build releases on:

- The "trusted" Buildkite organization that will build Bazel releases in the future.
- The "untrusted" Buildkite organization where all other CI jobs run.

Only code that has been 100% reviewed by Bazel core team members can run in the trusted zone in order to reduce risk of infecting the machines with malware and then accidentally shipping an infected version of Bazel to our users.

You could run release builds in the untrusted zone and while I think we've done a pretty good job of keeping these machines clean, stateless and somewhat trustable, in theory someone could escape the sandbox on these machines and affect builds that come afterwards.

On Linux / Windows we could somehow solve this by never recycling machines (always delete and start a fresh one), but for macOS we only have physical machines that we can't easily wipe clean after each job.

Maybe that's still better than any other option. We can look into it, but currently have to do other tasks first.

evankanderson:
It sounds like we already have release builds for macOS (which mentions physical machines as a limit), but not Windows. I'm guessing the problem with Windows is getting the automation set up?

I'm running into this trying to build https://github.com/jetstack/cert-manager, which depends on these release binaries through some platform-matching magic which fails on Windows. I'm building on a VM in the meantime, but it would be nice to have a faster edit-compile cycle that didn't involve syncing files between machines. 😬

[...]

PR #631, opened 2019-05-15
Rewrite status.sh in Python

vladmos:
`status.sh` is used by BazelCI to build buildtools-related binaries, but it can't be used on Windows.

`status.py` is a multiplatform alternative, it will completely replace `status.sh` once the latter is not used anymore by the CI configuration.

#375

### DIFF
--- a/status.py
+++ b/status.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-from __future__ import print_function
+from __future__ import print_function, unicode_literals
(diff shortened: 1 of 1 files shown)


vladmos:
@fweikert PTAL, do I need to do anything else with the script?

vladmos:
That's not covered by tests (because `status.sh` will be removed anyway), but `diff <(./status.sh) <(./status.py)` returns an empty result.

PR #632, opened 2019-05-15
Fix status.py for Python 3

vladmos:
Python 3 prints byte strings as their `repr`:

    >>> print(b'asd')
    b'asd'

They need to be converted to unicode before printing.


vladmos:
Yes, `diff <(python2 status.py) <(python3 status.py)` (the output is empty).

PR #635, opened 2019-05-16
Fix docstring warnings ranges

vladmos:
In case there is no module or function docstring, buildifier attaches a warning to the first non-comment statement in the sequence (of file or function statements). This may look fine in the console (because the console only shows line numbers of the findings beginnings, however looks weird in CIs where a warnings is usually shown right after its range ends. Thus, if a module has no docstring and starts with a long function definition, the warning is shown after the last line of the function. If a function has no docstring and starts with a long if- or for-block, the warning will be shown somewhere in the middle of the function.

This PR fix the issue: "no module docstring" will always be attached to the range of the first symbol of the first non-comment statement of the file, "no function docstring" warnings will be attached to function headers (from `def` till `:`).

### DIFF
--- a/warn/warn_docstring.go
+++ b/warn/warn_docstring.go
@@ -12,7 +12,7 @@ import (
 // a public function is required to have a docstring.
 const FunctionLengthDocstringThreshold = 5
 
-// getDocstrings returns a docstring of the statemenets and true if it exists.
+// getDocstrings returns a docstring of the statements and true if it exists.
 // Otherwise it returns the first non-comment statement and false.
 func getDocstring(stmts []build.Expr) (build.Expr, bool) {
 	for _, stmt := range stmts {
@@ -33,9 +33,9 @@ func moduleDocstringWarning(f *build.File, fix bool) []*Finding {
 		return []*Finding{}
 	}
 	if stmt, ok := getDocstring(f.Stmt); stmt != nil && !ok {
-		start, end := stmt.Span()
+		start, _ := f.Span()
 		return []*Finding{
-			makeFinding(f, start, end, "module-docstring", "The file has no module docstring.", true, nil),
+			makeFinding(f, start, start, "module-docstring", "The file has no module docstring.", true, nil),
 		}
 	}
 	return []*Finding{}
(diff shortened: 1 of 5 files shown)


laurentlb:
An empty range can be problematic - what should a tool underline?

vladmos:
A tool could show a warning attached to the first line (and the message should be clear to the user so that there's no need to see where exactly the problem in the line is). But I can't say for all possible tools.

warn/warn_docstring.go
```diff
@@ -12,7 +12,7 @@ import (
 // a public function is required to have a docstring.
 const FunctionLengthDocstringThreshold = 5
 
-// getDocstrings returns a docstring of the statemenets and true if it exists.
+// getDocstrings returns a docstring of the statements and true if it exists.
 // Otherwise it returns the first non-comment statement and false.
 func getDocstring(stmts []build.Expr) (build.Expr, bool) {
```
laurentlb: If you don't need it, don't return the first statement anymore.

laurentlb:
I'd prefer to have a real span.
Maybe the first line? (from 1:1 to 2:1)

warn/warn_docstring.go
vladmos: It's technically still used (no docstring warnings are not shown if the file has no non-comment statements), but I don think that makes much sense, will remove.

vladmos:
What if a file contains just one line (no `\n`s at all)? Can such span break something?

warn/warn_docstring.go
vladmos: I'll remove this logic, but the first returned argument is still used when I need to analyze the docstring contents.

laurentlb:
Not sure. You can skip the warning if the file is empty, if you want.

vladmos:
Another problem is that the `Position` object contains not only `Line` and `LineRune`, but also `Byte`. Different clients can use different combinations to determine the positions in the file, and parsed files don't have the information about how many bytes are in the first line (to return 2:1 correctly).

Maybe if there's a non-comment statement, I should return the span for its first symbol? That can be calculated easily given that the first statement can't be indented.

vladmos:
Replaced the empty span with the first symbol of the first non-comment statement. Editors and other clients will have something to underline, it'll also be clear that the problem is not with the statement itself.

PR #636, opened 2019-05-17
Update rules_go and gazelle

hlopko:
Rules go and gazelle are sensitive to order of *_dependencies calls, so I'm copy pasting the snippet from https://github.com/bazelbuild/bazel-gazelle (+ adding mirror.bazel.build).

Thanks!


laurentlb:
... except that tests fail

hlopko:
Not anymore :) Had to fight with pregenerated protos a bit.

PR #637, opened 2019-05-17
Refactor the interface of warnings

vladmos:
Currently the interface of warning/fix functions in Buildifier is not optimal for several reasons:

  1. There's a lot of boilerplate code in the warnings returned: the return objects are top-level `Finding` objects which are supposed to contain the range of the finding (which in almost all cases is a span of a node, so each warning has `start, end := node.Span()` prepended), they're supposed to contain a fix (it's not feasible to calculate them independently in each warning function), they return the category name which is a duplication (each warning function is called when the corresponding category name is known anyway).
  2. The functions that are able to make a fix edit the tree. There are two major problems with it:
    a. Returned warnings can be disabled by a special comment (`# buildozer: disable=...`), but fixes cant (unless each function looks for such a comment around the edited place).
    b. It's not possible to suggest individual fixes for each finding (the functions fix everything they can find). It's also not possible to provide correct diffs for different types of findings (by the time the second warning function is called, the tree may have already been edited by the first function).

The proposed solution is to return lower-level `LinterFinding` objects instead of `Finding` (which can be converted to `Finding` later, e.g. after calculating an appropriate diff for each of the findings). The functions also don't touch the tree anymore, instead they provide pairs (`&oldExpr`, `newExpr`) which can be applied conditionally (e.g. only if the expressions are not covered by disabling comments), they can also be rolled back (so that to calculate a diff one can roll them forward, calculate a diff and roll back to make it possible for the next function to operate on an untouched tree).

This PR only converts a couple of warning functions, marking the old style warning functions as "legacy", because it's easier to review, and also there are warning functions defined elsewhere (not in this repo) that haven't been refactored yet, buildifier should be able to handle both interface types at the same time.

This PR also doesn't implement replacements (atomic diffs for each finding), that will be implemented later.


warn/warn.go
```diff
@@ -184,8 +212,20 @@ func FileWarnings(f *build.File, pkg string, enabledWarnings []string, fix bool)
 
 	for _, warn := range warnings {
 		if fct, ok := FileWarningMap[warn]; ok {
+			for _, w := range fct(f) {
```
pmbethe09: for if for if if for ?  Too much nesting.

Can you refactor?

vladmos: Simplified by creating two subfunctions (`runFileWarningsFunction` and `runRuleWarningsFunction`). The legacy code for running the functions with the old interface will be removed soon(ish) anyway, but I'm going to implement replacements on `attr-cfg` first (if that works correctly and I don't need to alter the interface anymore I'll convert all other functions and replacements will become available for them automatically).

PR #638, opened 2019-05-17
Add native-android fix to buildifier

timpeut:
Also fixes a bug where the same global would be loaded multiple times.


timpeut:
Done

warn/warn_bazel_api.go
```diff
@@ -168,7 +178,12 @@ func notLoadedFunctionUsageCheck(f *build.File, category string, globals []strin
 	walk(&expr, bzlenv.NewEnvironment())
 
 	if fix && len(toLoad) > 0 {
-		f.Stmt = edit.InsertLoad(f.Stmt, loadFrom, toLoad, toLoad)
+		loads := []string{}
+		for k := range toLoad {
+			loads = append(loads, k)
+		}
+		sort.Slice(loads, func(i, j int) bool { return loads[i] < loads[j] })
```
vladmos: You can use `sort.Strings` to sort a slice of strings without providing a comparison function.

WARNINGS.md
```diff
@@ -456,6 +457,16 @@ UPPER_SNAKE_CASE, and providers should be UpperCamelCase ending with `Info`.
 
 --------------------------------------------------------------------------------
 
+## <a name="native-android"></a>All Android build rules should be loaded from Starlark
+
+  * Category name: `native-android`
+  * Automatic fix: no
```
vladmos: There's an automatic fix for this category.

vladmos:
Sorry, I've just realized: your fix will work for BUILD files, but in .bzl files the usage you need to match is not just `android_library` but `native.android_library` (and the other functions as well). In that case you'll need to both add a load and replace the node with a simple function call.

An example fix that matches `native.<something>`: https://github.com/bazelbuild/buildtools/blob/master/warn/warn_bazel.go#L47 . You'll basically need to do the same but in .bzl files, and you should only match `CallExpr` nodes with the `X` attribute equal to `native.<your function>` (which is a `DotExpr` node with `X` == `Ident` (`Name` == "native") and `Name` == "<your function>".

timpeut:
Added support for native.X

PR #642, opened 2019-05-28
Support raw strings in the formatter

vladmos:
The information whether a string was raw or not is currently not parsed and is lost during re-formatting.


laurentlb:
It looks like the previous behavior was to convert the raw string into a normal string. Did you change the behavior? Just adding a `r` in front might change the meaning.

Also, your tests miss the interesting cases, those with `\`.

PR #648, opened 2019-06-11
Implement automatic suggestions

vladmos:
The suggestions are implemented for all new-style warnings that return replacements. They are not yet implemented in the command-line binary but only in the buildifier library. Command-line support will be implemented in a follow-up CL to make it easier to import changes.


warn/warn_test.go
```diff
@@ -114,3 +114,145 @@ func checkFindingsAndFix(t *testing.T, category, input, output string, expected
...
+	}{
+		{
+			start:       28,
+			end:         42,
+			replacement: "",
+		},
+		{
+			start:       59,
+			end:         73,
+			replacement: "",
```
laurentlb: why you only have tests without replacement?

vladmos: Because currently the only supported replacement is `attr-cfg` which removes `cfg = "data"`. Empty string doesn't mean there's no replacement, it means that the given range should be replaced with an empty string. I'll add more tests here once more warning functions support replacements including some that actually add something new to the file.

PR #654, opened 2019-06-17
Update warning functions in bazel_warn.go to the new signature

vladmos:
This is the first CL in a series, the rest will be submitted separately.


vladmos:
There's just one RuleWarning function here, but more are defined internally, so I've created a wrapper that can easilly convert per-rule to per-file functions, that will help avoid some duplicated logic (i.e. check that a file is a collection of rules (BUILD or WORKSPACE), iterate over top-level statements, check whether they are function calls).

PR #657, opened 2019-06-19
Restore the old behavior

vladmos:
#655 and #656 slightly changed the behavior of some functions (i.e. differently formatted warnings, different Load statements were modified if multiple loads from the same file were found, etc). This PR restores the previous behavior.


vladmos:
There are tests but they are internal, bulldozer tests are not opensourced yet.

PR #659, opened 2019-06-21
Cleanup the code after the refactorings

vladmos:
Cleaning up the unused chunks of the code, also making it possible to have rule warning and file warning functions with different signatures (file warning functions don't need the `pkg string` parameter, rule warning functions return at most one warning per rule).

The PR is required for the compatibility with some internal-only warnings.

### DIFF
--- a/warn/warn.go
+++ b/warn/warn.go
@@ -163,13 +163,13 @@ func fileWarningWrapper(fct func(f *build.File) []*LinterFinding) func(f *build.
 }
 
 // ruleWarningWrapper is a wrapper that converts a per-rule function to a per-file function.
-// It also doesn't run on .bzl of default files.
+// It also doesn't run on .bzl or default files, only on BUILD and WORKSPACE files.
 func ruleWarningWrapper(ruleWarning func(call *build.CallExpr, pkg string) *LinterFinding) func(f *build.File, pkg string) []*LinterFinding {
 	return func(f *build.File, pkg string) []*LinterFinding {
 		if f.Type != build.TypeBuild && f.Type != build.TypeWorkspace {
 			return nil
 		}
-		findings := []*LinterFinding{}
+		var findings []*LinterFinding
 		for _, stmt := range f.Stmt {
 			switch stmt := stmt.(type) {
 			case *build.CallExpr:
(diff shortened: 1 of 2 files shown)


warn/warn.go
```diff
@@ -142,39 +147,86 @@ var FileWarningMap = map[string]func(f *build.File) []*LinterFinding{
...
+// fileWarningWrapper is a wrapper that converts a file warning function to a generic function.
+// A generic function takes a `pkg string` argument which is not used for file warnings, so it's just removed.
+func fileWarningWrapper(fct func(f *build.File) []*LinterFinding) func(f *build.File, pkg string) []*LinterFinding {
+	return func(f *build.File, pkg string) []*LinterFinding {
+		return fct(f)
+	}
+}
+
+// ruleWarningWrapper is a wrapper that converts a per-rule function to a per-file function.
+// It also doesn't run on .bzl of default files.
```
laurentlb: of default?

warn/warn.go
```diff
@@ -90,7 +94,9 @@ func makeLinterFinding(node build.Expr, message string, replacement ...LinterRep
 
 // RuleWarningMap lists the warnings that run on a single rule.
 // These warnings run only on BUILD files (not bzl files).
-var RuleWarningMap = map[string]func(f *build.File, pkg string, expr build.Expr) *Finding{}
+var RuleWarningMap = map[string]func(call *build.CallExpr, pkg string) *LinterFinding{
+	"positional-args": positionalArgumentsWarning,
+}
```
laurentlb: Is it still useful to have a separate map?
(instead of using `RuleWarning(positionalArgumentsWarning)`)

vladmos: Yes, because some rule warnings also require a `pkg` argument, and none of file warning functions need it.

I introduced `RuleWarning(...)` a couple of weeks ago, but now that I've also refactored the internal functions I see that this approach doesn't always work.

warn/warn.go
vladmos: there are 4 file types defined: BUILD, WORKSPACE, .bzl, and "default" (for just Starlark files, no Bazel-specific warnings should be raised for them).

vladmos: I see the typo now, will fix.

PR #660, opened 2019-06-21
Add a test for WARNINGS.md

vladmos:
All warnings contain a link to `https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#<category_name>`, the new test ensures that all registered warnings are documented in the file (at least that there's a link in the header and an anchor for each warning).


warn/documentation_test.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		t.Fatal(err)
+		return
+	}
+
+	contents := string(data)
+	for _, warning := range AllWarnings {
+		link := fmt.Sprintf("  * [%s](#%s)", warning, warning)
+		if !strings.Contains(contents, link) {
+			t.Errorf("No link found for the warning %q in WARNINGS.md, is it documented?", warning)
```
laurentlb: nit. the error message should include the pattern you match (`link`).

warn/documentation_test.go
```diff
@@ -0,0 +1,37 @@
...
+		}
+
+		anchor := fmt.Sprintf(`<a name="%s"></a>`, warning)
+		if !strings.Contains(contents, anchor) {
+			t.Errorf("No anchor found for the warning %q in WARNINGS.md, is it documented?", warning)
```
laurentlb: here, include `anchor`

warn/documentation_test.go
vladmos: Done.

warn/documentation_test.go
vladmos: Done.

PR #667, opened 2019-07-15
Support new operators

vladmos:
The unary `+` operator is implemented in Bazel in https://github.com/bazelbuild/bazel/pull/8890, bitwise operators are supported by https://github.com/bazelbuild/bazel/pull/8903. Buildifier should support them too.


build/testdata/052.golden
```diff
@@ -1,3 +1,11 @@
+# Unary operators
+
+-foo
+
++foo
```
laurentlb: add test for `+ + foo` or `- + foo`

vladmos: Done, also added support for bitwise operators (in the same PR)

build/print.go
```diff
@@ -334,6 +334,10 @@ var opPrec = map[string]int{
 	"//":     precMultiply,
 	"%":      precMultiply,
 	"|":      precMultiply,
+	"&":      precMultiply,
+	"^":      precMultiply,
+	"<<":     precMultiply,
+	">>":     precMultiply,
```
laurentlb: This doens't seem to match 
https://docs.python.org/3/reference/expressions.html#operator-precedence or https://github.com/google/starlark-go/blob/master/doc/spec.md#binary-operators

vladmos: Fixed.

PR #668, opened 2019-07-15
Add `remove_comment` command to Buildozer.

Zilulil:
This command removes the comment attached to the specified rule, attribute, or value in an attribute.

### DIFF
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -353,18 +353,25 @@ func cmdRemoveComment(opts *Options, env CmdEnvironment) (*build.File, error) {
 	case 0: // Remove comment attached to rule
 		env.Rule.Call.Comments.Before = nil
 		env.Rule.Call.Comments.Suffix = nil
+		env.Rule.Call.Comments.After = nil
 	case 1: // Remove comment attached to attr
 		if attr := env.Rule.AttrDefn(env.Args[0]); attr != nil {
 			attr.Comments.Before = nil
 			attr.Comments.Suffix = nil
+			attr.Comments.After = nil
 			attr.LHS.Comment().Before = nil
+			attr.LHS.Comment().Suffix = nil
+			attr.LHS.Comment().After = nil
+			attr.RHS.Comment().Before = nil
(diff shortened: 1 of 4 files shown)


edit/buildozer.go
```diff
@@ -348,6 +348,31 @@ func cmdRemove(opts *Options, env CmdEnvironment) (*build.File, error) {
 	return nil, nil
 }
 
+func cmdRemoveComment(opts *Options, env CmdEnvironment) (*build.File, error) {
+	switch len(env.Args) {
+	case 0: // Remove comment attached to rule
+		env.Rule.Call.Comments.Before = nil
+		env.Rule.Call.Comments.Suffix = nil
```
vladmos: A rule (and in theory any node in the AST) may have an after-comment, please nullify `Rule.Call.Comments.After` too.

    my_rule(
        name = "foo" ,
    )
    # this comment is attached to the rule above

edit/buildozer.go
```diff
@@ -348,6 +348,31 @@ func cmdRemove(opts *Options, env CmdEnvironment) (*build.File, error) {
...
+			attr.Comments.Before = nil
+			attr.Comments.Suffix = nil
+			attr.LHS.Comment().Before = nil
+			attr.RHS.Comment().Suffix = nil
+		}
+	case 2: // Remove comment attached to value
+		if attr := env.Rule.Attr(env.Args[0]); attr != nil {
+			if expr := ListFind(attr, env.Args[1], env.Pkg); expr != nil {
+				expr.Comments.Before = nil
+				expr.Comments.Suffix = nil
```
vladmos: The same as above (although arguments normally have just before or suffix comments, it's cleaner to remove everything).

Zilulil: Done, additionally nil'd out every field on the attr and attr.LHS and attr.RHS.

vladmos: Good catch, a comment like ` attr = value  # comment` will actually be attached to the RHS, not to the assignment node.

PR #669, opened 2019-07-16
Docstring warning

vladmos:
A recent change has broken the position of "no function docstring" warnings. Their span should be just the function's header (from `def` till `:`), not the whole body.


vladmos:
Seems I've pushed a wrong branch, will reopen the PR.

PR #672, opened 2019-07-19
Update buildifier setup instructions.

g-easy:
The previous instructions don't work with new versions of bazel like
0.28.0.


limdor:
@g-easy @vladmos is it intended in this change that buildifier_dependencies is no longer called?

g-easy:
I don't remember why that change was made, but it worked at the time.

PR #674, opened 2019-07-19
Add native-proto fix to buildifier

Yannic:
This change adds a fix for the native-to-starlark migration of Protobuf
rules. Also, the implementation is refactored to add support for warnings
when symbols should be loaded.

/cc @hlopko @lberki @iirina


lberki:
/cc @lberki 

I'm not fully qualified to review this, so I'll let the others shine, but I'm happy that this is happening!

Yannic:
@vladmos Can we get this in before Bazel 0.29.0 is released? https://github.com/bazelbuild/bazel/issues/8922

PR #677, opened 2019-07-24
buildifier: Support keyword-only parameters

jmillikin-stripe:
Keyword-only parameters are now permitted in Starlark (https://github.com/bazelbuild/starlark/issues/23), and are supported by the parser in starlark-go. This PR adds the new syntax to `parse.y` for `buildifier`, and upgrades the starlark-go dependency for `buildifier2`.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -29,6 +29,6 @@ buildifier_dependencies()
 
 go_repository(
     name = "skylark_syntax",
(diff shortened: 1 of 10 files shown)


jmillikin-stripe:
Done, added a `def` with keyword-only params to `build/testdata/053.in`.

vladmos:
Thanks!

PR #680, opened 2019-07-29
Provide a warning for overly nested depsets

### LINKED ISSUE #679: Warn against potentially overly nested depsets
In Bazel, depsets are not always used efficiently. API changes (https://github.com/bazelbuild/bazel/issues/5817) encourage users to write the efficient solution. 

We can statically detect bad patterns. For example, this is sometimes used to add things to a depset:

```python
x = depset(..., transitive = [..., x, ...])
```

If this pattern is used inside a `for` loop, we should report a warning because it can lead to overly nested depsets (which lead to performance problems).

vladmos:
Fixes #679.


lberki:
this is really cool

PR #686, opened 2019-07-30
Update bzl escape for incompatible flags

keith:
This adds support for `--incompatible_restrict_string_escapes` https://github.com/bazelbuild/bazel/issues/8380


keith:
cc @laurentlb

PR #671, opened 2019-07-17
Add native-java fix to buildifier.

iirina:

PR #691, opened 2019-08-05
Add native-py fix to buildifier

brandjon:
This is pretty much a copy/paste of #671 for Python rules. It migrates for the
incompatible change `--incompatible_load_python_rules_from_bzl` (bazelbuild/bazel#9006).


brandjon:
Missing from this automated migration is adding `@rules_python` to the WORKSPACE. Does/should buildifier help automate that? If not, we should probably update the warnings file to document this and link to the rules_python readme (same for java and cc).

warn/warn.go
```diff
@@ -130,6 +130,7 @@ var FileWarningMap = map[string]func(f *build.File) []*LinterFinding{
 	"native-build":              nativeInBuildFilesWarning,
 	"native-cc":                 nativeCcRulesWarning,
 	"native-java":               nativeJavaRulesWarning,
+	"native-py":                 nativePyRulesWarning,
```
vladmos: Move two lines below

brandjon: Done.

brandjon:
Thanks!

PR #692, opened 2019-08-06
Fix for legacy named arguments

vladmos:
Replaces legacy keyword arguments with positional arguments in some builtin functions, e.g. `bool(x = foo)` -> `bool(foo)`.

bazelbuild/bazel#5010

### DIFF
--- a/warn/warn_bazel_api.go
+++ b/warn/warn_bazel_api.go
@@ -833,103 +833,139 @@ func ruleImplReturnWarning(f *build.File) []*LinterFinding {
 	return findings
 }
 
-var legacyNamedParameters = map[string]string{
-	"all":     "elements",
-	"any":     "elements",
-	"tuple":   "x",
-	"list":    "x",
-	"len":     "x",
-	"str":     "x",
-	"repr":    "x",
-	"bool":    "x",
-	"int":     "x",
-	"dir":     "x",
-	"type":    "x",
-	"hasattr": "x",
-	"getattr": "x",
-	"select":  "x",
+type signature struct {
+	Positional []string // These parameters are typePositional-only
+	Keyword    []string // These parameters are typeKeyword-only
 }
 
-var legacyPositionalParameters = map[string]map[int]string{
-	"glob": {
-		1: "exclude",
-	},
+var signatures = map[string]signature{
+	"all":     {[]string{"elements"}, []string{}},
+	"any":     {[]string{"elements"}, []string{}},
+	"tuple":   {[]string{"x"}, []string{}},
+	"list":    {[]string{"x"}, []string{}},
+	"len":     {[]string{"x"}, []string{}},
+	"str":     {[]string{"x"}, []string{}},
+	"repr":    {[]string{"x"}, []string{}},
+	"bool":    {[]string{"x"}, []string{}},
+	"int":     {[]string{"x"}, []string{}},
+	"dir":     {[]string{"x"}, []string{}},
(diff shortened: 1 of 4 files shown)


warn/warn_bazel_api.go
```diff
@@ -832,3 +832,105 @@ func ruleImplReturnWarning(f *build.File) []*LinterFinding {
...
+	"str":     "x",
+	"repr":    "x",
+	"bool":    "x",
+	"int":     "x",
+	"dir":     "x",
+	"type":    "x",
+	"hasattr": "x",
+	"getattr": "x",
+	"select":  "x",
+}
```
laurentlb: How are you going to fix functions with more than 1 argument?

vladmos: I was going to implement it in portions, but perhaps I should support flexible amount of arguments now and just add functions/methods later.

vladmos: Refactored to handle both legacy keyword and legacy positional parameters, even for the same function

warn/warn_bazel_api_test.go
```diff
@@ -650,3 +650,67 @@ def macro():
...
+int(x = "13", base = 8)
+dir(x = foo)
+type(x = foo)
+hasattr(x = foo, name = "bar")
+getattr(x = foo, name = "bar", default = "baz")
+select(x = {})
+glob(["*.cc"], ["test*"])
+native.glob(["*.cc"], ["test*"])
+glob(["*.cc"])
+native.glob(["*.cc"])
```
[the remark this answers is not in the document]
vladmos: Done.

warn/warn_bazel_api_test.go
```diff
@@ -650,3 +650,99 @@ def macro():
...
+list([1, 2, 3])
+len([1, 2, 3])
+str(foo)
+repr(foo)
+bool(3)
+int("3")
+int("13", base = 8)
+dir(foo)
+type(foo)
+hasattr(foo, name = "bar")
```
laurentlb: Should be: `hasattr(foo, "bar")`

warn/warn_bazel_api_test.go
```diff
@@ -650,3 +650,99 @@ def macro():
...
+getattr(foo, name = "bar", default = "baz")
```
laurentlb: Should be:
`getattr(foo, "bar", "baz")`

warn/warn_bazel_api_test.go
```diff
@@ -650,3 +650,99 @@ def macro():
...
+glob(include = [], exclude = [])
+glob([], exclude = [])
+glob([], [], 1)
+glob(*args, [])
+`, `
+glob(["*.cc"], exclude = ["test*"])
+glob(["*.cc"])
+glob([], exclude = [])
+glob([], exclude = [])
+glob([], [], 1)
```
laurentlb: Add exclude_directories:
`glob([], exclude = [], exclude_directories = 1)`

warn/warn_bazel_api_test.go
vladmos: Done.

warn/warn_bazel_api_test.go
vladmos: Done.

warn/warn_bazel_api_test.go
vladmos: Done.

vladmos:
I'll run it on the existing files right before importing.

PR #693, opened 2019-08-07
WIP: Add rewrite for @bazel_tools//tools/build_defs/pkg to @rules_pkg

aiuto:
Add rewrite for load("@bazel_tools//tools/build_defs/pkg:...", ...) 
to load("@rules_pkg//:...", ...)

See: https://github.com/bazelbuild/bazel/issues/8857

It can't do WORKSPACE automatically, so it is sort of moot. Maybe not even legitimate to automate this kind of migration.  The only valid update is to add one of these stanza to the workspace

if bazel_federation is NOT mentioned in WORKSPACE

```
maybe(
        http_archive,
        name = "rules_pkg",
        urls = [
            "https://mirror.bazel.build/github.com/bazelbuild/rules_pkg/rules_pkg-0.2.0.tar.gz",
            "https://github.com/bazelbuild/rules_pkg/releases/download/0.2.0/rules_pkg-0.2.0.tar.gz",
        ],
        sha256 = "5bdc04987af79bd27bc5b00fe30f59a858f77ffa0bd2d8143d5b31ad8b1bd71c",
    )
```

Or, if @bazel_federation is already in the WORKSPACE

```
load("@bazel_federation//:repositories.bzl", "rules_pkg")
rules_pkg()
```

### DIFF
--- a/build/rewrite.go
+++ b/build/rewrite.go
@@ -80,6 +80,7 @@ type RewriteInfo struct {
 	FormatDocstrings int      // number of reindented docstrings
 	ReorderArguments int      // number of reordered function call arguments
 	EditOctal        int      // number of edited octals
+	UseRulesPkg      int      // number of uses of tools/build_defs/pkg converted
 	Log              []string // log entries - may change
 }
 
@@ -95,6 +96,7 @@ func (info *RewriteInfo) Stats() map[string]int {
 		"formatdocstrings": info.FormatDocstrings,
 		"reorderarguments": info.ReorderArguments,
 		"editoctal":        info.EditOctal,
+		"userulespkg":      info.UseRulesPkg,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


aiuto:
On second thought this is a dumb idea. buildifier should not do things that update WORKSPACE. That is complex enough so that it needs a much more sophisticated tool. I'm just keeping it until I capture the knowledge and turn it into docs for buildifier.

vladmos:
Perhaps it's better to implement such trasformations as a part of the fixer instead (in the `warn/` package). There are a few differences:

  * Fixes there can be applied conditionally (i.e. you can specify what categories you're interested in),
  * They can have warnings (if you don't want to fix a file but just see what are the potential problems)
  * They can be suppressed with a special comment if you're sure you don't want a certain finding to be fixed
  * They don't affect the formatter (e.g. if you have a presubmit check that makes sure that your files are formatted correctly, it won't be broken if you just implement a new fix)

Issue #696, opened 2019-08-09
Buildozer support for selects


cristiancreteanu:
So, I have this bazel file (a BUCK file used by the buck build tool):

```
cxx_library(
    name = "1",
)

cxx_library(
    name = "2",
)

cxx_library(
    name = "mock",
    public_include_directories = [
        ".",
    ],
    raw_headers = [
        "mock.h",
    ],
    deps = select({
        "macos": [
            ":1",
            ":2",
        ],
        "DEFAULT": [
            ":1",
        ],
    }),
)
```

The following command:
```
buildozer //mock:mock remove deps :1 :2
```
... returns the file as it was before, without making any changes. I would like to change one or multiple branches of that `select()`. Does buildozer not provide support for this?

pmbethe09:
Seems like a good idea.  while buildozer re-uses the buildifier parser, so it should be aware of selects, afaik support for selects is not implemented.

Any interest in contributing?

cristiancreteanu:
Yes, this feature would be quite helpful for me. I will start looking into the code.

cristiancreteanu:
Yes, this feature could be quite useful. I will start looking into the code.

cristiancreteanu:
I have the following scenarios in mind which should be handled:
```
deps = select({
        "macos": [
            ":1",
            ":2",
        ],
        "DEFAULT": [
            ":1",
        ],
    }),
```

1. `remove deps ":1"` results in:
```
deps = select({
        "macos": [
            ":2",
        ],
        "DEFAULT": [],
    }),
```

2. `add deps ":1"` results in:
```
deps = select({
        "macos": [
            ":2",
        ],
        "DEFAULT": [],
    }) + [":1"],
```

3. if, after changing a select, it ends up like:
```
deps = select({
        "macos": [
            ":2",
            ":3",
            ":4",
        ],
        "linux": [
            ":2",
            ":3",
            ":5",
        ],
        "DEFAULT": [
            ":2",
            ":3",
        ],
    }),
```
It should be resolved to:
```
deps = select({
        "macos": [":4"],
        "linux": [":5"],
        "DEFAULT": [],
    }) + [":2", ":3"],
```
Basically, get the intersection of all the lists and take it out of the select.

4. If all the dictionary list values are empty, then the select is turned into empty list.

5. For `deps = x + y`, if any of the operands of `+` is an empty list, then the attribute value should not contain it. Namely, if x == [], then `deps = y`.

PR #706, opened 2019-08-28
Buildozer support for select add/remove

cristiancreteanu:
Fix for #696

### DIFF
--- a/edit/edit.go
+++ b/edit/edit.go
@@ -393,6 +393,21 @@ func AllLists(e build.Expr) []*build.ListExpr {
 	return nil
 }
 
+// AllSelects returns all the selects concatenated in an expression.
+func AllSelects(e build.Expr) []*build.CallExpr {
+	switch e := e.(type) {
+	case *build.BinaryExpr:
+		if e.Op == "+" {
+			return append(AllSelects(e.X), AllSelects(e.Y)...)
+		}
+	case *build.CallExpr:
+		if x, ok := e.X.(*build.Ident); ok && x.Name == "select" {
+			return []*build.CallExpr{e}
+		}
+	}
+	return nil
+}
+
 // FirstList works in the same way as AllLists, except that it
 // returns only one list, or nil.
 func FirstList(e build.Expr) *build.ListExpr {
@@ -461,6 +476,22 @@ func ContainsComments(expr build.Expr, str string) bool {
 	return false
 }
 
+// RemoveFromList removes one element from a ListExpr and stores
+// the deleted StringExpr at the address pointed by the last parameter
+func RemoveFromList(li *build.ListExpr, item, pkg string, deleted **build.StringExpr) {
+	var all []build.Expr
+	for _, elem := range li.List {
+		if str, ok := elem.(*build.StringExpr); ok {
+			if LabelsEqual(str.Value, item, pkg) && (DeleteWithComments || !hasComments(str)) {
+				*deleted = str
+				continue
+			}
+		}
+		all = append(all, elem)
+	}
+	li.List = all
+}
+
 // ListDelete deletes the item from a list expression in e and returns
 // the StringExpr deleted, or nil otherwise.
 func ListDelete(e build.Expr, item, pkg string) (deleted *build.StringExpr) {
@@ -470,18 +501,25 @@ func ListDelete(e build.Expr, item, pkg string) (deleted *build.StringExpr) {
 	deleted = nil
 	item = ShortenLabel(item, pkg)
 	for _, li := range AllLists(e) {
-		var all []build.Expr
-		for _, elem := range li.List {
-			if str, ok := elem.(*build.StringExpr); ok {
-				if LabelsEqual(str.Value, item, pkg) && (DeleteWithComments || !hasComments(str)) {
-					deleted = str
-					continue
+		RemoveFromList(li, item, pkg, &deleted)
+	}
+
+	for _, sel := range AllSelects(e) {
+		if len(sel.List) == 0 {
+			continue
+		}
+
+		if dict, ok := sel.List[0].(*build.DictExpr); ok {
+			for _, keyVal := range dict.List {
+				if keyVal, ok := keyVal.(*build.KeyValueExpr); ok {
+					if val, ok := keyVal.Value.(*build.ListExpr); ok {
+						RemoveFromList(val, item, pkg, &deleted)
+					}
 				}
 			}
-			all = append(all, elem)
 		}
-		li.List = all
 	}
+
 	return deleted
 }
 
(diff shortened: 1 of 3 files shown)


cristiancreteanu:
@googlebot I signed it!

cristiancreteanu:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here with `@googlebot I signed it!`) and we'll verify it.
> 
> #### What to do if you already signed the CLA
> ##### Individual signers
> * It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
> ##### Corporate signers
> * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google.com/docs/cla/#troubleshoot)).
> * The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Fbazelbuild%2Fbuildtools%2Fpull%2F706) for more info**.

@googlebot I signed it!

cristiancreteanu:
@googlebot I signed it!

laurentlb:
Thanks! From a quick glance, it looks reasonable. I'll let @vladmos do the review next week.

@vladmos Do we have tests for this? (or is it still not open-sourced?)

vladmos:
Overall the PR looks good, could you please add some tests? There are some tests for similar methods in `edit/edit_test.go`, for example, `TestAddValueToListAttribute`, you can create a similar test for the newly added methods.

edit/edit.go
```diff
@@ -461,6 +476,146 @@ func ContainsComments(expr build.Expr, str string) bool {
...
+// the dictionary argument of the select expression given as a parameter
+func SelectListsIntersection(sel *build.CallExpr, pkg string) (allEmpty bool, intersection []build.Expr) {
+	dict := sel.List[0].(*build.DictExpr)
+	firstList := dict.List[0].(*build.KeyValueExpr).Value.(*build.ListExpr).List
+	intersection = make([]build.Expr, len(firstList))
+	copy(intersection, firstList)
+	allEmpty = len(firstList) == 0
+
+	for _, keyVal := range dict.List[1:] {
+		keyVal := keyVal.(*build.KeyValueExpr)
```
vladmos: Although DictExpr nodes shouldn't have anything than KeyValueExpr as subnodes, in theory that's still possible, e.g. if there's a bug in another method in this file. It's safer to use a checked type cast and just `continue` if the current element is not what's expected, otherwise buildozer will just crash if something goes wrong.

cristiancreteanu: Wouldn't it be advisable to expose such a bug if it exists? Or print a warning if the type is different from KeyValueExpr.

cristiancreteanu:
> Overall the PR looks good, could you please add some tests? There are some tests for similar methods in `edit/edit_test.go`, for example, `TestAddValueToListAttribute`, you can create a similar test for the newly added methods.

Will do!

edit/edit.go
vladmos: It makes sense, however it may indicate a bug somewhere else (especially if this module is used by a completely different project as a library, failing here would mean that the main project crashes.

edit/edit.go
```diff
@@ -461,6 +476,150 @@ func ContainsComments(expr build.Expr, str string) bool {
...
+				return e.Y
+			}
+
+			if yIsList && len(y.List) == 0 {
+				return e.X
+			}
+		}
+	case *build.CallExpr:
+		if x, ok := e.X.(*build.Ident); ok && x.Name == "select" {
+			dict := e.List[0].(*build.DictExpr)
```
vladmos: This will crash if the `select` call has no arguments or the first argument is not a dict expression. That shoudn't usually happen, but it's still not a syntax error and buildozer shouldn't crash in this case.

Also, the argument to the `select` function can be a constant defined elswhere: `select(MY_CONFIGS)`. Please make sure buildozer doesn't crash there (and add a test).

edit/edit.go
```diff
@@ -461,6 +476,150 @@ func ContainsComments(expr build.Expr, str string) bool {
...
+	}
+
+	return e
+}
+
+// SelectListsIntersection returns the intersection of the lists of strings inside
+// the dictionary argument of the select expression given as a parameter
+func SelectListsIntersection(sel *build.CallExpr, pkg string) (intersection []build.Expr) {
+	dict := sel.List[0].(*build.DictExpr)
+	firstList := dict.List[0].(*build.KeyValueExpr).Value.(*build.ListExpr).List
```
vladmos: Same here (in the two lines above). You're assuming that 1. the select has an argument, 2. it's a dict literal, 3. the dict is non-empty, 4. its subnodes are KeyValueExpr nodes, 5. the value of the virst key-value pair is a list literal. If any of these assumptions are not correct, buildozer will fail.

Although if the 4th assumption is wrong it means there's a bug somewhere, all other cases can occur in input files (even if some of these files are not valid for Bazel).

Even if you're sure that the value must be a list, it's not necessarily a list literal, it can be a constant or an expression, the computed value of which is a list, e.g. `getParams(foo).split(";")`, where `getParams` returns a string.

Please add tests for these cases to make sure they don't crash buildozer.

cristiancreteanu: Yes, I didn't take such possibilities into account. Thanks! :D

edit/edit.go
```diff
@@ -461,6 +476,186 @@ func ContainsComments(expr build.Expr, str string) bool {
...
+	if len(list2) == 0 {
+		return []build.Expr{}
+	}
+
+	i := 0
+	for j, common := range list1 {
+		found := false
+		for _, elem := range list2 {
+			if str, ok := elem.(*build.StringExpr); ok {
+				if str.Value == common.(*build.StringExpr).Value {
```
vladmos: There's one more unchecked type cast, please check whether `common` is actually a `StringExpr`.

cristiancreteanu: Sorry, missed that one :D

vladmos:
Thanks!

PR #728, opened 2019-10-05
unused_deps: Support BUILD.bazel files

### LINKED ISSUE #204: unused_deps doesn't support BUILD.bazel files
According to https://github.com/bazelbuild/bazel/issues/552, `BUILD.bazel` is a valid name for BUILD file, but `unused_deps` called on the target with `BUILD.bazel` file tries to open `BUILD` and fails:

```
2018/02/08 19:00:32 open /some/path/BUILD: no such file or directory when parsing /some/path/BUILD
```

### LINKED ISSUE #441: unused_deps not detecting unused dependencies at all
Hi everyone,

I'm so sorry for posting this again, right after @pmbethe09 closes issue #396. But I've upgraded `unused_deps` to the latest release, `0.19.2` today, and `unused_deps` is still not detecting any unused dependencies.

You can find our code at [`haikalpribadi/grakn` branch `unused-deps-not-working`](https://github.com/haikalpribadi/grakn/tree/unused-deps-not-working), which you can clone and run:
```
bazel build //client-java:all
bazel run //dependencies/tools:unused-deps -- //client-java
```

You will find that `unused_deps` returns: `No unused deps found.`

Meanwhile, we know for a fact that the following dependencies of `//client-java:client-java` are no longer needed (and we have confirmed by removing them afterwards).
```
//dependencies/maven/artifacts/com/codahale/metrics:metrics-core
//dependencies/maven/artifacts/com/github/rholder:guava-retrying
//dependencies/maven/artifacts/com/netflix/hystrix:hystrix-core
//dependencies/maven/artifacts/com/sun/jersey:jersey-client
//dependencies/maven/artifacts/io/reactivex:rxjava
//dependencies/maven/artifacts/javax/ws/rs:jsr311-api
//dependencies/maven/artifacts/org/sharegov:mjson
```

Can someone help us look into this?

aljoscha:
Before, we would only support BUILD files. Now, we check if the BUILD
file exists, if not we add the .bazel extension and return that.

I didn't add a test because there previously didn't exist a test for `InterpretLabelForWorkspaceLocation` and it's a bit more involved because it includes parsing/finding the `WORKSPACE` files, among other things. If you have a suggestion for a possible test I'm happy to take a stab at it.

In my local testing (with Bazel 0.29.1) `build_deps` now successfully finds unused dependencies and prints `buildozer` commands for removing them. 

This fixes #204
This fixes #441

I have a Google ICLA on file.


aljoscha:
I added a test case but I couldn't figure out what the second part of the tested function is meant to do. I don't know if any production code or test code ever has the case that the prefix of `target` is not `//`. Anyone know what's up with this?

aljoscha:
I added another commit that fixes path handling on Windows. This was a pre-existing problem but was only discovered by the newly added test.

edit/edit.go
```diff
@@ -59,7 +58,7 @@ func ParseLabel(target string) (string, string, string) {
 	if len(parts) == 1 {
 		if strings.HasPrefix(target, "//") || tables.StripLabelLeadingSlashes {
 			// "//absolute/pkg" -> "absolute/pkg", "pkg"
-			return repo, parts[0], path.Base(parts[0])
+			return repo, parts[0], filepath.Base(parts[0])
```
vladmos: Why `filepath`? Labels always have forward slashes regardless of current operating system.

aljoscha: Ah, you're right. It seems I went a bit overboard when I fixed file path handling for windows.

Reverting 👌

edit/edit.go
```diff
@@ -136,21 +136,34 @@ func InterpretLabelForWorkspaceLocation(root string, target string) (buildFile s
...
 	if pkg != "" {
 		buildFile = pkg + "/BUILD"
+		if !isFile(buildFile) {
+			// try it with the .bazel extension
+			buildFile += ".bazel"
+		}
+
 	} else {
 		buildFile = "BUILD"
+		if !isFile(buildFile) {
```
vladmos: Instead of duplicating the check, you can move it after the if-else block.

edit/edit_test.go
```diff
@@ -673,3 +676,51 @@ x = 2`,
...
+func runTestInterpretLabelForWorkspaceLocation(t *testing.T, buildFileName string) {
+	tmp, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmp)
+	if err := os.MkdirAll(filepath.Join(tmp, "a", "b"), 0755); err != nil {
+		t.Fatal(err)
+	}
+	// todo(aljoscha): See if we should hardcode "WORKSPACE" here or should use wspace.workspaceFile
```
vladmos: You have two pending TODOs, are you going to keep them? If so, it's better to file corresponding bugs and use links for them instead of your username, this way it'll be much easier to track them.

aljoscha: Ah yes, sorry about that. I meant them mostly as questions for the reviewer or someone familiar with the code

So for the first one: should I leave `"WORKSPACE"` here and assume that this will always be the filename of workspace files?

For the second one: I don't know when the second part of the `InterpretLabelForWorkspaceLocation` function would be used. If I remove the second part of the method, that is those lines:
```
if isFile(pkg) {
	// allow operation on other files like WORKSPACE
	buildFile = pkg
	pkg = filepath.Join(relativePath, filepath.Dir(pkg))
	return
}
if pkg != "" {
	buildFile = pkg + "/BUILD"
	if !isFile(buildFile) {
		// try it with the .bazel extension
		buildFile += ".bazel"
	}

} else {
	buildFile = "BUILD"
	if !isFile(buildFile) {
		// try it with the .bazel extension
		buildFile += ".bazel"
	}
}
pkg = filepath.Join(relativePath, pkg)
```
`bazel test /...` still passes. Which doesn't mean that we don't need them, it might just be missing test coverage.

edit/edit.go
aljoscha: pushed the change for that

edit/edit_test.go
vladmos: 1. I think it's ok to hardcode the "WORKSPACE" string for now.

2. The second part of the function will be executed if the following check doesn't pass: `strings.HasPrefix(target, "//")`, i.e. the label is not absolute but relative, e.g. `":a"` or `":a/b"`.

aljoscha: Thanks! And I'm sorry, this is taking up quite a bit of your time.

I understand the flow of the function, I wasn't sure about the behaviour because it was not previously covered by tests. And the existing behaviour is a bit weird (or doesn't throw errors). For example:
 - `/path/to/WORKSPACE:foo` gives (`"/path/to/WORKSPACE"`, `"/path/to"`, `"foo"`)
 - `/path/to/WORKSPACE` gives (`"BUILD"`, `""`, `"/path/to/WORKSPACE"`)
(the tuple is the result of `InterpretLabelForWorkspaceLocation`, which is `(buildFile, pkg, rule)`

The first one seems (is?) correct, the second result seems to be bogus, but this is what the existing code spits out.

vladmos: I think it worth investigating, but right now it's safer to leave it as it is.

edit/edit.go
```diff
@@ -136,21 +136,29 @@ func InterpretLabelForWorkspaceLocation(root string, target string) (buildFile s
...
 	if isFile(pkg) {
 		// allow operation on other files like WORKSPACE
 		buildFile = pkg
-		pkg = path.Join(relativePath, filepath.Dir(pkg))
+		pkg = filepath.Join(relativePath, filepath.Dir(pkg))
 		return
 	}
 	if pkg != "" {
-		buildFile = pkg + "/BUILD"
+		buildFile = filepath.Join(pkg, "/BUILD")
```
vladmos: Good catch, but now the tests fail on Windows, you probably need to update them by replacing a slash with `filepath.Separator`.

vladmos: Also, shouldn't it be `"BUILD"` instead of `"/BUILD"`?

aljoscha: I only need to remove the `/` because `Join` already adds that. (I don't have windows so I only notice it from the buildkite run. Getting this to pass on windows is a bit tricky, I'll notify you again once I have it figured out.

I think the gist of it is that `ParseLabel` does not work for absolute windows paths. It was only tested on unix-style paths: https://github.com/bazelbuild/buildtools/blob/96b69a8e017b0943fa3003d118794324e2b75344/edit/edit_test.go#L46. The reason is that the code simply splits on the `:`, which is a part of the path on windows.

vladmos: That's probably the reason, at least it can explain the following error: https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/bf49dbeb-eb10-4ee6-bbdc-669c127fbc3f/edit%5Cgo_default_test%5Cattempt_1.log

Feel free to just revert the change to this line if you don't want to fix it now, it's not directly relevant to your pull request.

aljoscha: Another suggestion is to ignore this issue for now, remove the last two test cases, and create a new issue for fixing the parsing issues on windows.

aljoscha: I removed the unrelated change and also removed the test cases (that I added) which are failing on windows. I can add a new ticket for these issues afterwards if you'd like that.

Also: please let me know if I should rebase/squash this once you're done with the review. I'm not familiar with how you handle that on this project.

vladmos:
Thanks!

I can update and squash with a single click if there are no merging conflicts, so that’s not necessary.

PR #729, opened 2019-10-08
buildozer: Add command for appending to lists in a dict attribute

chingjun:
We've recently added a build attribute which is a [string_list_dict](https://docs.bazel.build/versions/master/skylark/lib/attr.html#string_list_dict), just to find out that there is no way to operate on such an attribute from buildozer, so I'm sending this PR. Thanks.

Example:

```
buildozer 'dict_list_add attr key1 value1 value2' 'dict_list_add attr key2 value3' //some/target
```

Should generate:
```
a_rule(
    name = "target",
    attr = {
        "key1": [
            "value1",
            "value2",
        ],
        "key2": ["value3"],
    },
)
```

Internal bug: b/22484279

### DIFF
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -573,6 +573,33 @@ func cmdDictRemove(opts *Options, env CmdEnvironment) (*build.File, error) {
 	return env.File, nil
 }
 
+// cmdDictListAdd adds an item to a list in a dict.
+func cmdDictListAdd(opts *Options, env CmdEnvironment) (*build.File, error) {
+	attr := env.Args[0]
+	key := env.Args[1]
+	args := env.Args[2:]
+
+	dict := &build.DictExpr{}
+	if currDict, ok := env.Rule.Attr(attr).(*build.DictExpr); ok {
+		dict = currDict
+	}
+
+	prev := DictionaryGet(dict, key)
+	if prev == nil {
+		prev = &build.ListExpr{}
+	}
+
+	for _, val := range args {
+		expr := getStringExpr(val, env.Pkg)
+		prev = AddValueToList(prev, env.Pkg, expr, true)
+	}
+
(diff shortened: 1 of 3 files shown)


edit/buildozer.go
```diff
@@ -573,6 +573,34 @@ func cmdDictRemove(opts *Options, env CmdEnvironment) (*build.File, error) {
...
 
+// cmdDictListAdd adds a item to a list in a dict.
+func cmdDictListAdd(opts *Options, env CmdEnvironment) (*build.File, error) {
+	attr := env.Args[0]
+	key := env.Args[1]
+	args := env.Args[2:]
+
+	dict := &build.DictExpr{}
+	currDict, ok := env.Rule.Attr(attr).(*build.DictExpr)
+	if ok {
```
vladmos: You can combine these two lines:

    if currDict, ok := env.Rule.Attr(attr).(*build.DictExpr); ok {

edit/buildozer.go
```diff
@@ -573,6 +573,34 @@ func cmdDictRemove(opts *Options, env CmdEnvironment) (*build.File, error) {
 	return env.File, nil
 }
 
+// cmdDictListAdd adds a item to a list in a dict.
```
vladmos: "an item"

vladmos:
Update: it's actually better to use the `buildozer_test.go` file, there's currently just one command test.

edit/buildozer.go
chingjun: Updated, thanks.

edit/buildozer.go
chingjun: Fixed

chingjun:
Updated and added tests. PTAL, thanks.

chingjun:
Conflict resolved. PTAL, thanks!

PR #732, opened 2019-10-10
Fix recursive traversal of BUILD files for "..." targets

### LINKED ISSUE #731: File not found error with buildozer
Latest commit (a795ddb) in buildozer is causing file not found error and blocking presubmit.  An example error can be seen in https://source.cloud.google.com/results/invocations/e09ece88-4c14-4d0b-92bd-eb5f122cbe09/targets.  Could you please take a look?  Thanks.

Running check buildozer fix... ++ run_buildozer
++ /tmpfs/src/tools/bin/buildozer -quiet 'fix movePackageToTop unusedLoads usePlusEqual' //...:__pkg__
/tmpfs/src/github/gapid/.../BUILD.bazel: file not found or not readable
++ local r=2
++ '[' 2 -eq 3 ']'
++ return 2
++ echo FAILED
FAILED
++ echo '  Error executing: run_buildozer'
  Error executing: run_buildozer
++ exit 1

aljoscha:
This fixes #731

### DIFF
--- a/edit/buildozer_test.go
+++ b/edit/buildozer_test.go
@@ -14,8 +14,13 @@ distributed under the License is distributed on an "AS IS" BASIS,
 package edit
 
 import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
+
+	"reflect"
 
 	"github.com/bazelbuild/buildtools/build"
 )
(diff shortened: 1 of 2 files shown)


edit/buildozer_test.go
```diff
@@ -114,3 +119,69 @@ func TestCmdRemoveComment(t *testing.T) {
...
+		{tmp, "//a/b", []string{filepath.Join(tmp, "a", "b", buildFileName)}},
+		{tmp, "//a/b:foo", []string{filepath.Join(tmp, "a", "b", buildFileName)}},
+		{tmp, "//...", []string{filepath.Join(tmp, buildFileName), filepath.Join(tmp, "a", buildFileName), filepath.Join(tmp, "a", "b", buildFileName), filepath.Join(tmp, "a", "c", buildFileName)}},
+		{tmp, "//a/...", []string{filepath.Join(tmp, "a", buildFileName), filepath.Join(tmp, "a", "b", buildFileName), filepath.Join(tmp, "a", "c", buildFileName)}},
+		{tmp, "//a/b/...", []string{filepath.Join(tmp, "a", "b", buildFileName)}},
+		{tmp, "//a/c/...", []string{filepath.Join(tmp, "a", "c", buildFileName)}},
+		{tmp, "//a/c/...:foo", []string{filepath.Join(tmp, "a", "c", buildFileName)}},
+	} {
+		buildFiles := targetExpressionToBuildFiles(tc.rootDir, tc.target)
+		expectedBuildFilesMap := make(map[string]string)
```
aljoscha: I'm not sure if this is the right way to do this comparison. (I only learned Go this weekend to fix the `BUILD.bazel` bug, so my code might not be that idiomatic...)

edit/buildozer_test.go
```diff
@@ -14,9 +14,14 @@ distributed under the License is distributed on an "AS IS" BASIS,
 package edit
 
 import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
```
vladmos: Please remove the empty line, also I believe `gofmt` will sort the imports lexicographically (except the last github import because it's separated by an empty line), so please also do that too.

aljoscha: Fixed! I was assuming that `goimports`/`gofmt` would take care of all the formatting but it doesn't remove empty lines if I accidentally split the imports into sections.

vladmos: Empty lines are allowed for separating blocks of imports, like here imports from built-in libraries are separated from the github import below.

vladmos:
If this PR fixes #731 please mention it in the description. Also please replace the maps with `map[string]bool` (you probably have missed my comment above).

aljoscha:
Sorry about that, I assumed you meant it's ok to leave it as is in the tests.

aljoscha:
Thanks for the review!

vladmos:
@aljoscha: to fix the behavior I had to partially revert `targetExpressionToBuildFiles` which surprisingly didn't break any tests including those you've added to `buildozer_test.go` with the same pull request. Let me know if that still breaks your usecases (I've added a new test case that was broken by your change but fixed with the revertion).

vladmos:
Sorry, forgot to mention the PR: #738

aljoscha:
I'm in vacation but I'll have a look at this once I'm back. 👌

aljoscha:
@vladmos You did an alternative fix by adding these checks: https://github.com/bazelbuild/buildtools/blob/9f3978593d3ee017a9badc0adde9c552390cf7bf/edit/edit.go#L140

This way, the "hack" of keeping the `.../BUILD` suffix and then trimming it in `targetExpressionToBuildFiles` works again. That's why the tests still pass. 👌

PR #733, opened 2019-10-11
Don't warn about positional arguments if a macro has just one argument

### LINKED ISSUE #727: buildifier wrongly warns about positional arguments for variadic function
Given this workspace content:

```python
load("@rules_antlr//antlr:repositories.bzl", "rules_antlr_dependencies")

rules_antlr_dependencies("4.7.2")
```
Running buildifier yields the following warning:

```
WORKSPACE:31: positional-args: All calls to rules or macros should pass arguments by keyword (arg_name=value) syntax. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#positional-args)
```

But the function looks like this:

```python
def rules_antlr_dependencies(*versions):
    """Loads the dependencies for the specified ANTLR releases.

    Args:
      *versions: the ANTLR releases to make available.
    """
    ...
```

Ideally, buildifier should analyze the actual function definition in order to differentiate positional args vs variadic args.

Also, if a function only has 1 argument, enforcing the keyword seems a bit overkill to me.

vladmos:
Fixes #727.

### DIFF
--- a/warn/warn_bazel.go
+++ b/warn/warn_bazel.go
@@ -133,6 +133,9 @@ func positionalArgumentsWarning(call *build.CallExpr, pkg string) *LinterFinding
 	if id, ok := call.X.(*build.Ident); !ok || functionsWithPositionalArguments[id.Name] {
 		return nil
 	}
+	if len(call.List) <= 1 {
+		return nil
+	}
 	for _, arg := range call.List {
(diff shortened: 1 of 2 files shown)


laurentlb:
Normally, macros should have a name argument (which should be a keyword argument).

PR #738, opened 2019-10-24
Fix targetExpressionToBuildFiles

vladmos:
The behavior was partially broken by #732. `targetExpressionToBuildFiles` was returning incorrect results if it was called with an empty `rootDir` parameter and a `target` containing `"..."` (e.g. when buildozer was called with `...:*` as a target).


vladmos:
Updated the description.

PR #730, opened 2019-10-09
Fix parsing of dictionary values that contain ":"

nickjmeyer:
If a dictionary value contains a :, then the value is not correct. Need to preserve everything after the first :. One example where this comes up is setting environment variables for paths when using container_image from the docker rules.

PR #739, opened 2019-10-31
Fix parser for adding to dictionary

nmeyer-argo:
This is a repeat of #730.  I opened the original with the wrong account for the CLA.

If a dictionary value contains a :, then the value is not parsed correctly. Need to preserve everything after the first :. One example where this comes up is setting environment variables for paths when using container_image from the docker rules. For example, when running dict_add env PATH:/a/b/c:/d/e/f it ends up only including /a/b/c and it ignores /d/e/f.

### DIFF
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -513,7 +513,7 @@ func cmdDictAdd(opts *Options, env CmdEnvironment) (*build.File, error) {
 	}
 
 	for _, x := range args {
-		kv := strings.Split(x, ":")
+		kv := strings.SplitN(x, ":", 2)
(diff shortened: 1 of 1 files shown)


nmeyer-argo:
Having trouble getting the CLA system to accept my agreement.  I submitted an agreement and both the POC and the signer from my organization approved.  This was yesterday morning.  Anyone know if it just takes some time to get accepted?  If I try and resubmit the CLA it'll say "you have already submitted a CLA", but the main page says "no agreements on file".  And yes I've checked that the email addresses match.

nmeyer-argo:
@googlebot I signed it!

nmeyer-argo:
@googlebot I signed it!

PR #746, opened 2019-11-25
Implement the `export_file` command in buildozer

vladmos:
To add or modify an "exports_files" declaration one may use the following command:

    buildozer 'export_file filename.txt //visible/for:__pkg__' //path/to/target

It's also possible to provide no or several visibility arguments, if a corresponding "exports_files" section with the same visibility parameters exists it will be updated.


oreflow:
Closing this as it's been stale since 2019

PR #747, opened 2019-11-25
Don't reformat raw string literals

### LINKED ISSUE #740: buildifier should not eliminate raw strings
If a string literal is raw (`r` prefix), it's because the author specifically decided it would be more readable that way than backslash-escaping the string contents. We should respect that decision.

```sh
$ cat test.bzl
print(r'\a\b\c')
$ buildifier test.bzl 
$ cat test.bzl
print("\\a\\b\\c")
```

Expected result: File content is rewritten as `print(r"\a\b\c")`.

vladmos:
Users use raw string literals if the value contains some cases of escaping, buildifier shouldn't reformat such strings.

Fixes #740.


build/testdata/060.golden
```diff
@@ -28,4 +28,14 @@ strings = [
     "'\"foo\"'",
     '''\'\'\'"""foo"""\'\'\'''',
     """'''\"\"\"foo\"\"\"'''""",
+
+    # raw strings
+    r'\foo',
+    r'"\foo',
+    r"\foo",
+    r"'\foo",
+    r'''\foo''',
```
laurentlb: why not double-quotes here?

vladmos: Replaced with double quotes (but only if there are no single quotes inside at all). E.g. `r''' ' '''` won't be reformatted as `r""" ' """` (even if in this case it's possible).

PR #762, opened 2020-01-16
Replace the string-escape check with mandatory reformatting

vladmos:
If a string literal contains an unknown escape sequence (e.g. `"\a"`) the backslash symbol should become escaped itself: `"\\a"'. This is consistent with the current behavior of Bazel.


vladmos:
I'm not sure what should happen if a string literal contains `\1` (can often be found in regular expressions). Now buildifier will reformat it as `\001` but it's better to yield `\\1`, right?

laurentlb:
> Now buildifier will reformat it as \001 but it's better to yield \\1, right?

`\001` is the correct output, since `\\1` has a different meaning.
(we can discuss what to do once `\1` is forbidden)

vladmos:
`\001` is correct, but there are probably existing files with `\1` used in regex strings (which currently works) and reformating that with `\001` will change the semantics. Also I can imagine that users forget to escape the backslash properly and buildifier does it for them on autosave. In these situations it's probably more useful to convert `\1` to `\\1` instead of `\001` both now and after `\1` is forbidden, what do you think?

laurentlb:
```python
>> "\1" == "\\1"
False

>> "\1" == "\001"
True
```

Buildifier shouldn't change the behavior of a working code, even if you think the user did a mistake. :)

But at least, replacing with `\001` will make the issue much more obvious. So the new behavior is better than the old one, right?

vladmos:
Yes, you're right, I though interpreting `\1` as `\001` was a special behavior of Buildifier, but it seems Bazel does the same thing, so I can submit it without breaking anything.

PR #772, opened 2020-01-27
Improve the warning message for missing argument documentation

### LINKED ISSUE #768: Allow documentation for **kwargs in wrappers OR turn off lint warning.
Consider this wrapper method:
```
# buildifier: disable=function-docstring-args
def license(name, license_kinds = None, license_kind = None, copyright_notice = None, **kwargs):
    """Wrapper for license rule.

    Args:
      name: str target name.
      license_kinds: list(label) list of license_kind targets.
      license_kind: label a single license_kind. Only one of license_kind or license_kinds may
                    be specified
      copyright_notice: str Copyright notice associated with this package.
    """
```
First of all, the parsing must be broken, because (without the disable line) I see the lint warning
 
```
Argument "**kwargs" is not documented. 
 https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#function-docstring-args
```

I added this to the docstring
`kwargs: additional args`

And *still* get the above warning, plus
```
Argument "kwargs" is documented but doesn't exist in the function signature. 
  https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#function-docstring-args
```

This is a no win.

Suggested solution:  Allow this in the docstring
`kwargs: [forwarded to <other rule name>]`

The magical [fowarded to ...] would tell lint not to complain about it.
The <other rule name> could be used by Stardoc to pull up the definitions from the underlying rule, so we could document the wrapper appropriately.

vladmos:
If a function signature contains arguments with asterisks ("*args" or "**kwargs") users often forget to include the asterisks into the documentation, which causes warnings like `Arguments "*args", "**kwargs" are not documented` and `Argument "args" is documented but doesn't exist in the function signature`. To avoid the confusion, the latter warning message is extended with `Do you mean "*args"?` (or `"**kwargs"`) if such an argument actually exists in the function signature.

Closes #768

### DIFF
--- a/warn/warn_docstring.go
+++ b/warn/warn_docstring.go
@@ -323,22 +323,23 @@ one (preferably two) space more than "Args:", for example:
 
 		// Check whether all documented arguments actually exist in the function signature.
 		for name, pos := range info.args {
-			if !paramNames[name] {
-				msg := fmt.Sprintf("Argument %q is documented but doesn't exist in the function signature.", name)
-				// *args and **kwargs should be documented with asterisks
-				for _, asterisks := range []string{"*", "**"} {
-					if paramNames[asterisks+name] {
-						msg += fmt.Sprintf(` Do you mean "%s%s"?`, asterisks, name)
-						break
-					}
-				}
-				posEnd := pos
-				posEnd.LineRune += len(name)
-				finding := makeLinterFinding(doc, msg)
-				finding.Start = pos
-				finding.End = posEnd
-				findings = append(findings, finding)
+			if paramNames[name] {
+				continue
 			}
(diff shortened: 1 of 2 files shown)


warn/warn_docstring.go
```diff
@@ -324,9 +324,17 @@ one (preferably two) space more than "Args:", for example:
 		// Check whether all documented arguments actually exist in the function signature.
 		for name, pos := range info.args {
 			if !paramNames[name] {
```
pmbethe09: invert/continue ?

vladmos: done

PR #775, opened 2020-02-03
Don't warn about undocumented *

### LINKED ISSUE #774: Buildifier: warning function-docstring-args cannot handle keyword-only function arguments
While working on https://github.com/bazelbuild/rules_proto/pull/22, I noticed that buildifier doesn't know how to handle the separator for keyword-only function arguments.

See https://buildkite.com/bazel/rules-proto/builds/258#76717e64-ae09-4bf7-ac1a-0861e530e45b

```starlark
def direct_source_infos(proto_info, *, provided_sources = []):
    """Returns sequence of `ProtoFileInfo` for `proto_info`'s direct sources.

    Files that are both in `proto_info`'s direct sources and in
    `provided_sources` are skipped. This is useful, e.g., for well-known
    protos that are already provided by the Protobuf runtime.

    Args:
      proto_info: An instance of `ProtoInfo`.
      provided_sources: Optional. A sequence of import paths to ignore.

    Returns: A sequence of `ProtoFileInfo` containing information about
        `proto_info`'s direct sources.
    """

    files = [_info(f, proto_info) for f in proto_info.direct_sources]
    return [f for f in files if f.import_path not in provided_sources]
```

```shell
$ buildifier --lint=warn --warnings=all proto/private/utils/direct_source_infos.bzl 
proto/private/utils/direct_source_infos.bzl:26: function-docstring-args: Argument "" is not documented. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#function-docstring-args)
```

Adding documentation for `*:` doesn't seem to fix it (and TBH, documenting that would be weird).

vladmos:
Fixes #774

### DIFF
--- a/warn/warn_docstring_test.go
+++ b/warn/warn_docstring_test.go
@@ -344,12 +344,14 @@ def f(x, *, y, z = None):
     x: x
     *: a separator
     y: y
+    : argument without a name
     z: z
    """
    pass
 `,
 		[]string{
 			`6: Argument "*" is documented but doesn't exist in the function signature.`,
+			`8: Argument "" is documented but doesn't exist in the function signature.`,
 		},
 		scopeEverywhere)
 
--- a/warn/warn_docstring_test.go
+++ b/warn/warn_docstring_test.go
@@ -337,6 +337,23 @@ def f(x, *, y, z = None):
 		scopeEverywhere)
 
 	checkFindings(t, "function-docstring-args", `
+def f(x, *, y, z = None):
+   """This is a function.
+
+   Args:
+    x: x
+    *: a separator
+    y: y
+    z: z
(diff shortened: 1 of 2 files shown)


warn/warn_docstring_test.go
```diff
@@ -322,6 +322,20 @@ def f(x, y, z = None, *args, **kwargs):
...
+   Args:
+    x: x
+    y: y
+    z: z
+   """
+   pass
+`,
+		[]string{},
+		scopeEverywhere)
+
```
laurentlb: Add a negative test, to check the error message when one argument is *

vladmos: Done.

PR #778, opened 2020-02-03
Allow escaped linebreaks in buildozer command arguments

vladmos:
#773 allows buildozer arguments to be separated by linebreaks. However in some cases it's required to have linebreaks inside arguments (e.g. for adding multiline comments), in such cases it should be allowed to use escaped linebreaks, such as:

    $ buildozer 'comment srcs a.cc Just\ a\
    multiline\ comment' //my:package


vladmos:
It exists but currently it's internal. I'll try to opensource buildozer tests soon.

vladmos:
will do

PR #779, opened 2020-02-06
buildozer: Add command replace_load

Yannic:


buildozer/README.md
```diff
@@ -68,6 +68,8 @@ Buildozer supports the following commands(`'command args'`):
     importing the symbols. Before using this, make sure to run
     `buildozer 'fix movePackageToTop'`. Afterwards, consider running
     `buildozer 'fix unusedLoads'`.
+  * `replace_load <path> <[to=]from(s)>`: Similar to `new_load`, but removes
+    existing load statements for the requested symbols before adding new loads.
```
laurentlb: Adding an example (in the Examples section below) could be useful.

Yannic: Sounds good. I'll send a PR.

Yannic:
This is useful when moving a rule into another file. One use-case where this happens is when switching from a private rule to an Open-Source rule. You would typically change the rule in `private.bzl` to mimic the API of the OSS version, update the implementation of the private rule to an alias for the OSS rule, and then update all loads to use the OSS version directly. However, there may be other rules in `private.bzl`, so you'll only want to change the load for that one rule.

I can definitely see that it could be useful to change the path of a load without listing all symbols, e.g. running as part of a Copybara import config to change loads from `@rules_foo//foo:defs.bzl` to `//third_party/rules_foo/foo:defs.bzl`. A separate `substitute_load` command that allows replacing loads that match a regular expression sounds more powerful than a simple "replace-if-excact-match", though.

laurentlb:
Thanks! I agree that `replace_load` will be useful.

PR #784, opened 2020-02-10
Disallow implicit string literal concatentation

vladmos:
Unlike Python, Starlark doesn't support implicit string concatenation (e.g. `"aaa" "bbb"`).

Buildifier currently supports it and automatically joins the strings with a `+`, however if an editor runs buildifier on save automatically, a file with a typo can be reformatted incorrectly, e.g.

    [
        "aaa",
        "bbb"
        "ccc"
    ]

is automatically reformatted as

    [
        "aaa",
        "bbb" +
        "ccc",
    ]
 
instead of throwing a syntax error.


laurentlb:
Tests?

vladmos:
There are currently no tests for syntax errors, but the existing tests ensure that common usecases are not affected by the change.

vladmos:
Added negative tests.

PR #787, opened 2020-02-14
Add publishing to npm

alexeagle:
This replaces the mirroring we currently do in https://github.com/bazelbuild/rules_nodejs/blob/master/scripts/mirror_buildtools.sh

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,1 @@
+build:release -c opt --stamp --workspace_status_command="$PWD/status.py"
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,6 +1,11 @@
 load("@bazel_gazelle//:def.bzl", "gazelle")
 load("@com_github_bazelbuild_buildtools//buildifier:def.bzl", "buildifier")
 
+exports_files([
+    "LICENSE",
+    "launcher.js",
+])
+
 config_setting(
     name = "windows",
     values = {"cpu": "x64_windows"},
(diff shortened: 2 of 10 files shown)


alexeagle:
Note, I can't find anywhere in the repo that you use `--stamp` to create your existing release, so I'm not sure where to add the command to publish these packages.

Let me know where to add those...

vladmos:
The releases are currently created by Buildkite and a script there is supposed to use `--stamp` (however probably it doesn't because the latest versions of buildifier and buildozer fail to print the correct version when called with `--version`).

I'm going to rewrite the release script anyway, meanwhile perhaps you can create a .bazelrc file like the one for bazelisk.

alexeagle:
@vladmos thanks, added a release.sh that has just the npm bits for now

vladmos:
Thanks!

mdvorsky:
Hi! We use buildifier in our WORKSPACE and we fetch it from master (as recommended in https://github.com/bazelbuild/buildtools/blob/master/buildifier/README.md).

It looks like after this PR, buildifier stopped working and we had to both:
- Upgrade bazel_skylib to at least 0.8
- Add build_bazel_rules_nodejs to our WORKSPACE

Especially the second change is surprising, because we don't use build_bazel_rules_nodejs in our WORKSPACE at all, but somehow, buildifier doesn't bring that dependency on its own.

It looks like that build_bazel_rules_nodejs is both missing from buildifier deps.bzl and the instructions to install buildifier to the WORKSPACE are missing the call to buildifier_dependencies().

vladmos:
There's an issue for this: #796

PR #790, opened 2020-02-19
buildifier: Remove exit code printout for diff mode (#349)

Rasrack:
Solves issue 349

### DIFF
--- a/differ/diff.go
+++ b/differ/diff.go
@@ -23,6 +23,7 @@ import (
 	"os/exec"
 	"runtime"
 	"strings"
+	"syscall"
 )
 
 // Invocation of different diff commands, according to environment variables.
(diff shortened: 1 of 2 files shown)


differ/diff.go
```diff
@@ -56,11 +57,17 @@ func (d *Differ) run(command string, args ...string) error {
...
-	// As long as the program started (above), we don't care about the
-	// exact exit status. In the most common case, the diff command
-	// will exit 1, because there are diffs, causing bash to exit 1.
-	return cmd.Wait()
+	// Assume that the diff program adhere to ordinary diff commands
+	// exit codes i.e. 0 no differences, 1 differences found, >1 an
+	// error occoured.
+	fmt.Fprintf(os.Stdout, "differ: %s\n", cmd)
+	if err := cmd.Wait(); err != nil {
+		if status, ok := err.(*exec.ExitError).Sys().(syscall.WaitStatus); ok && status.ExitStatus() != 1 {
```
vladmos: This will crash is the `err` is not of type `ExitError`.  According to the documentation that may happen: https://golang.org/pkg/os/exec/#Cmd.Wait

Also I can't find the documentation for the `.Sys()` method, can it return nill or an object of something else than type `syscal.WaitStatus`?

PR #792, opened 2020-02-20
Fix .bzl visibility warnings

vladmos:
The warning didn't work correctly with relative file paths.

### DIFF
--- a/buildifier/utils/utils.go
+++ b/buildifier/utils/utils.go
@@ -77,9 +77,22 @@ func GetParser(inputType string) func(filename string, data []byte) (*build.File
 	}
 }
 
+// hasWorkspaceFile checks whether a given directory contains a file called
+// WORKSPACE or WORKSPACE.bazel. The contents are not checked
+func hasWorkspaceFile(path string) bool {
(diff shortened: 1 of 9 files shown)


vladmos:
That's Windows- (case-insensitive) and MacOS- (different interface of `sed`) specific in the shell test. I had to update the shell test because the unit test for the warning didn't catch everything, i.e. the location of the WORKSPACE file is supposed to be detected not by the warning function but by the main buildifier binary.

I'll fix the test soon.

buildifier/utils/utils.go
```diff
@@ -77,9 +77,22 @@ func GetParser(inputType string) func(filename string, data []byte) (*build.File
 	}
 }
 
+// hasWorkspaceFile checks whether a given directory contains a file called
+// WORKSPACE or WORKSPACE.bazel. The contents are not checked
```
laurentlb: Remove "The contents are not checked".

vladmos:
The tests are passing now.

buildifier/utils/utils.go

Issue #796, opened 2020-02-24
buildifier and buildozer have an implicit dependency on rules_nodejs


sudoforge:
`98244f725d46ab37614ac6bbb653a6827a46fabd` introduced an implicit dependency on `rules_nodejs` when running `buildifier`:

```
➜ bazel run //:buildifier
ERROR: /home/username/path/to/repo/BUILD:23:1: every rule of type _buildifier implicitly depends upon
rror loading package '@com_github_bazelbuild_buildtools//buildifier': Unable to find package for @build_bazel_rules_node
ERROR: /home/username/path/to/repo/BUILD:23:1: every rule of type _buildifier implicitly depends upon
ause of: error loading package '@com_github_bazelbuild_buildtools//buildifier': Unable to find package for @build_bazel_
ERROR: Analysis of target '//:buildifier' failed; build aborted: Analysis failed
INFO: Elapsed time: 0.354s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (1 packages loaded, 0 targets configured)
FAILED: Build did NOT complete successfully (1 packages loaded, 0 targets configured)
```

PR #797, opened 2020-02-24
fix(buildifier): add rules_nodejs to //buildifier:deps.bzl target

sudoforge:
Commit `98244f725d46ab37614ac6bbb653a6827a46fabd` introduced an implicit
dependency on `rules_nodejs` within `//buildifier:BUILD`, which breaks
downstream user's ability to run Buildifier.

This patch uses the `_maybe()` rule (defined within
`//buildifier:deps.bzl`) to conditionally register
`@build_bazel_rules_nodejs` so that downstream users do not need to
define this internal dependency within their `WORKSPACE`.

closes bazelbuild/buildifier#796

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -70,13 +70,6 @@ load("@com_github_bazelbuild_buildtools//buildifier:deps.bzl", "buildifier_depen
 
 buildifier_dependencies()
 
-# We don't use any nodejs but this includes a rule for publishing releases to npm
-http_archive(
-    name = "build_bazel_rules_nodejs",
-    sha256 = "b6670f9f43faa66e3009488bbd909bc7bc46a5a9661a33f6bc578068d1837f37",
-    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/1.3.0/rules_nodejs-1.3.0.tar.gz"],
-)
-
 load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories")
 
 node_repositories()
--- a/buildifier/deps.bzl
+++ b/buildifier/deps.bzl
@@ -9,6 +9,13 @@ def buildifier_dependencies():
         url = "https://github.com/bazelbuild/bazel-skylib/archive/c00ef493869e2966d47508e8625aae723a4a3054.tar.gz",  # 2018-12-06
     )
 
+    _maybe(
+        http_archive,
+        name = "build_bazel_rules_nodejs",
+        sha256 = "b6670f9f43faa66e3009488bbd909bc7bc46a5a9661a33f6bc578068d1837f37",
+        urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/1.3.0/rules_nodejs-1.3.0.tar.gz"],
+    )
+
 def _maybe(repo_rule, name, **kwargs):
     if name not in native.existing_rules():
         repo_rule(name = name, **kwargs)


sudoforge:
This _does_ duplicate the definition of the `build_bazel_rules_nodejs` repository, mostly because I don't want to have to define it within my workspace. To simplify future management of this particular external repository, we could define variables for the version, name, etc -- or simply update the "using buildifier via bazel" documentation to specify that users should also define the `build_bazel_rules_nodejs` repository within their workspace.

vladmos:
Thanks!

The repository definition is duplicated now, I suppose it should be removed from the `WORKSPACE` file because it already calls `buildifier_dependencies`.

sudoforge:
> Thanks!
> 
> The repository definition is duplicated now, I suppose it should be removed from the `WORKSPACE` file because it already calls `buildifier_dependencies`.

Yeah, I thought about removing the repository definition from the root-level `WORKSPACE` but that puts us in a bit of a weird dependency model due to the fact that the repository is also used in `//buildozer:BUILD.bazel` for the `pkg_npm` rule.

The way I see it, we have a few choices:

- Add a `_maybe` definition of the under the `buildozer` package (but then we'd have to manage two definitions of the repository; perhaps not a problem if `bazelbuild/buildtools` adds [`renovate`](https://github.com/marketplace/renovate) for automatically upgrading external dependency definitions).

- Remove the `_maybe` definition I've added here, instead keeping the definition in the root-level `WORKSPACE` file. This has the downside of requiring downstream users to define yet _another_ repository in their WORKSPACE, but keeps a clean depdendency model within `bazelbuild/buildtools`, since the `build_bazel_rules_nodejs` repository is indeed used for multiple packages (`buildifier` and `buildozer`)

- Keep the root-level `WORKSPACE` definition _and_ this `_maybe` definition (what the original patch for this PR did). This has the downside of managing two definitions for the repository, which is additional management overhead (without the use of something like [`renovate`](https://github.com/marketplace/renovate) as described above)

Let me know what you'd like me to do and I'll take care of it this afternoon.

vladmos:
Sorry for the late answer.

Buildifier and buildozer are already interconnected and depend on each other. `"bazel_skylib"` which is currently being registered in `buidlifier_dependencies()` is also used by `buildozer` as well.

sudoforge:
@vladmos Thanks for the green light. I'll amend this patch and remove the declaration of the repository from the root `WORKSPACE`.

WORKSPACE
```diff
@@ -69,14 +69,3 @@ rules_proto_toolchains()
...
 buildifier_dependencies()
-
-# We don't use any nodejs but this includes a rule for publishing releases to npm
-http_archive(
-    name = "build_bazel_rules_nodejs",
-    sha256 = "b6670f9f43faa66e3009488bbd909bc7bc46a5a9661a33f6bc578068d1837f37",
-    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/1.3.0/rules_nodejs-1.3.0.tar.gz"],
-)
-
-load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories")
```
vladmos: I assume this and the next line should be kept, otherwise

sudoforge: Whoops, you're absolutely right. `dG` at it's finest :)

vladmos:
Thanks!

laurentlb:
Why should users of Buildifier depend on rules_nodejs?

sudoforge:
@laurentlb As the commit body describes (as well as the original comment post for this PR), an earlier commit in this repository added a workspace-level definition for `rules_nodejs`. This was added in order to use the NPM packaging rule in `//buildifier:BUILD.bazel` and `//buildozer:BUILD.bazel`. 

Downstream users do not pull in the `WORKSPACE` file of this repository, because repository rules exclude the file to avoid conflicting or duplicate repository declarations. Because of this fact, updating their repository rule to use the aforementioned commit would break their ability to build their project, unless they had already defined a `@build_bazel_rules_nodejs` repository.

This patch avoids that downstream breakage for users who do not wish to add `rules_nodejs` to their workspace.

Let me know if this isn't clear, and I'll write up an example for you.

laurentlb:
First, thanks for reporting and fixing the problem, this is very appreciated. Also, sorry for the unclear message.

With your fix, I think every user of Buildifier will be downloading rules_nodejs, and I think we could avoid that. I haven't looked at the details, but we could maybe move the calls to pkg_npm to a separate BUILD file (so that end users of Buildifier who don't need the mpm don't need the dependency)?

sudoforge:
> With your fix, I think every user of Buildifier will be downloading rules_nodejs, and I think we could avoid that.

Actually, prior to this patch, anyone who wanted to use Buildifier through Bazel would have been forced to add `@build_bazel_rules_nodejs` to their workspace explicitly due to the implicit dependency in Buildifier's `BUILD` file (after commit `98244f725d46ab37614ac6bbb653a6827a46fabd`, that is).

This patch moves the repository definition to `buidlfier_dependencies()`, which will conditionally define the `@build_bazel_rules_nodejs` repository if the user isn't already explicitly defining that repository in their workspace. This has the benefit of not requiring the downstream user to define a `@build_bazel_rules_nodejs` repository in their workspace.

> ...we could maybe move the calls to pkg_npm to a separate BUILD file (so that end users of Buildifier who don't need the npm don't need the dependency)?

If I'm understanding you correctly, this would involve moving the `pkg_npm()` call from (at least) `//buildifier:BUILD.bazel` to some other location in the workspace. This could definitely work if we moved it somewhere else, say, `//npm:buildifier`, but might be considered anti-pattern, as the `pkg_npm()` call "belongs" to Buildifier.

sudoforge:
For additional context of the issue introduced by 98244f7 that this patch resolves, take a look at issue #796 and [this comment](https://github.com/bazelbuild/buildtools/pull/787#issuecomment-590885988) from #787.

PR #799, opened 2020-03-02
buildozer: Add parameter to substitute commands to exclude certain expressions

Yannic:
A common use-case for `buildozer 'substitute ...'` is to move everything
in a package to a different package (e.g. mapping packages from
`//...` to `//third_party/project/...` during Copybara import).
However, some packages have special meaning for Bazel and should
(usually) not be renamed.

This change adds an optional last parameter to `substitute` and
`substitute_load` that is used to exclude expressions if they match the
passed regexp.


oreflow:
Generally I would say this PR make sense, but closing it due to being stale. Please sync this to latest and reopen if it's still relevant.

PR #802, opened 2020-03-06
add nodejs binding to conveniently use buildozer from a node program

alexeagle:
/cc @kwonoj this is to support automating the ts project references -> BUILD file translation

### DIFF
--- a/buildozer/npm/index.ts
+++ b/buildozer/npm/index.ts
@@ -1,0 +1,55 @@
+/**
+ * @fileoverview NodeJS binding to run the buildozer command through a convenient API
+ */
+import {spawnSync} from 'child_process';
+// Use require() here since the file doesn't exist at design-time;
+// bazel will copy the /launcher.js in this repo while building the package
+const {getNativeBinary} = require('./buildozer');
+
+/**
+ * Models the data structure of a buildozer command file.
+ * You should group as many commands as possible into one call to `run` for efficiency.
+ * https://github.com/bazelbuild/buildtools/tree/master/buildozer#do-multiple-changes-at-once
+ */
+export interface CommandBatch {
+    /**
+     * Each entry is a buildozer edit command like 'new cc_library foo' or a print command
+     * @see https://github.com/bazelbuild/buildtools/tree/master/buildozer#edit-commands
+     */
+    commands: string[];
+    /**
+     * Each entry is like a Bazel label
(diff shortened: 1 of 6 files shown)


buildozer/npm/index.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * @fileoverview NodeJS binding to run the buildozer command through a convenient API
+ */
+const {spawnSync} = require('child_process');
```
alan-agius4: Since you are using `@types/node` is there a reason why not use `import` ? IE;
```
import {spawnSync} from 'child_process';
```

buildozer/npm/index.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Models the data structure of a buildozer command file.
+ * You should group as many commands as possible into one call to `run` for efficiency.
+ * https://github.com/bazelbuild/buildtools/tree/master/buildozer#do-multiple-changes-at-once
+ */
+export type CommandBatch = {
```
alan-agius4: NIT: shouldn't this be an `interface` since `CommandBatch` is restricted to a object?

buildozer/npm/index.ts
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Each entry is a buildozer edit command like 'new cc_library foo' or a print command
+     * see https://github.com/bazelbuild/buildtools/tree/master/buildozer#edit-commands
+     */
+    commands: string[];
+    /**
+     * Each entry is like a Bazel label
+     * See https://github.com/bazelbuild/buildtools/tree/master/buildozer#targets
```
alan-agius4: Nit: See -> `@see`

buildozer/npm/index.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Models the data structure of a buildozer command file.
+ * You should group as many commands as possible into one call to `run` for efficiency.
+ * https://github.com/bazelbuild/buildtools/tree/master/buildozer#do-multiple-changes-at-once
+ */
+export type CommandBatch = {
+    /**
+     * Each entry is a buildozer edit command like 'new cc_library foo' or a print command
+     * see https://github.com/bazelbuild/buildtools/tree/master/buildozer#edit-commands
```
alan-agius4: Nit: See -> `@see`

alexeagle:
@vladmos thanks!

PR #807, opened 2020-03-19
Implement deprecated function warning

vladmos:
Deprecated funcitons are defined in other .bzl files, so buildifier is now also able to read from loaded .bzl files.

### DIFF
--- a/buildifier/utils/utils.go
+++ b/buildifier/utils/utils.go
@@ -124,6 +124,8 @@ func getFileReader(workspaceRoot string) *warn.FileReader {
 	}
 
 	readFile := func(filename string) ([]byte, error) {
+	    // Use OS-specific path separators
+        filename = strings.ReplaceAll(filename, "/", string(os.PathSeparator))
 		path := filepath.Join(workspaceRoot, filename)
 
 		data, err := ioutil.ReadFile(path)
--- a/buildifier/utils/utils.go
+++ b/buildifier/utils/utils.go
@@ -3,6 +3,7 @@
 package utils
 
 import (
+    "io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
@@ -115,13 +116,37 @@ func SplitFilePath(filename string) (workspaceRoot, pkg string) {
 	return workspaceRoot, pkg
 }
 
+// getFileReader returns a *FileReader object that reads files from the local
+// filesystem if the workspace root is known.
+func getFileReader(workspaceRoot string) *warn.FileReader {
+	if workspaceRoot == "" {
+		return nil
+	}
+
+	readFile := func(filename string) ([]byte, error) {
+		path := filepath.Join(workspaceRoot, filename)
+
+		data, err := ioutil.ReadFile(path)
+		if err != nil {
+			return nil, err
+		}
+		return data, nil
+	}
+
+	fileReader := &warn.FileReader{}
+	fileReader.Init(readFile)
+	return fileReader
+}
+
 // Lint calls the linter and returns a list of unresolved findings
 func Lint(f *build.File, lint string, warningsList *[]string, verbose bool) []*warn.Finding {
+	fileReader := getFileReader(f.WorkspaceRoot)
+
 	switch lint {
 	case "warn":
-		return warn.FileWarnings(f, *warningsList, nil, warn.ModeWarn)
+		return warn.FileWarnings(f, *warningsList, nil, warn.ModeWarn, fileReader)
 	case "fix":
-		warn.FixWarnings(f, *warningsList, verbose)
+		warn.FixWarnings(f, *warningsList, verbose, fileReader)
 	}
 	return nil
 }
--- a/buildifier/utils/utils.go
+++ b/buildifier/utils/utils.go
@@ -131,9 +131,7 @@ func getFileReader(workspaceRoot string) *warn.FileReader {
 		return ioutil.ReadFile(path)
 	}
 
-	fileReader := &warn.FileReader{}
-	fileReader.NewFileReader(readFile)
-	return fileReader
+	return warn.NewFileReader(readFile)
 }
 
 // Lint calls the linter and returns a list of unresolved findings
--- a/buildifier/utils/utils.go
+++ b/buildifier/utils/utils.go
@@ -3,7 +3,7 @@
 package utils
 
 import (
-    "io/ioutil"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
@@ -124,19 +124,15 @@ func getFileReader(workspaceRoot string) *warn.FileReader {
 	}
 
 	readFile := func(filename string) ([]byte, error) {
-	    // Use OS-specific path separators
-        filename = strings.ReplaceAll(filename, "/", string(os.PathSeparator))
+		// Use OS-specific path separators
+		filename = strings.ReplaceAll(filename, "/", string(os.PathSeparator))
 		path := filepath.Join(workspaceRoot, filename)
 
-		data, err := ioutil.ReadFile(path)
-		if err != nil {
-			return nil, err
-		}
-		return data, nil
+		return ioutil.ReadFile(path)
 	}
 
 	fileReader := &warn.FileReader{}
-	fileReader.Init(readFile)
+	fileReader.NewFileReader(readFile)
 	return fileReader
 }
 
--- a/warn/multifile.go
+++ b/warn/multifile.go
@@ -11,12 +11,15 @@ type FileReader struct {
 	readFile func(string) ([]byte, error)
 }
 
-// NewFileReader initializes a FileReader instance with an instance of function
-// that can read an arbitrary file in the repository using a path relative
-// to the workspace root.
-func (fr *FileReader) NewFileReader(readFile func(string) ([]byte, error)) {
-	fr.cache = make(map[string]*build.File)
-	fr.readFile = readFile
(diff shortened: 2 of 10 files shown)


warn/warn_test.go
```diff
@@ -19,6 +19,26 @@ const (
 	scopeBazel      = scopeBuild | scopeBzl | scopeWorkspace
 )
 
+// A global FileReader object that can be used by tests. If a test redefines it it must
+// reset it when it finishes.
+var testFileReader *FileReader
+
+func setUpFileReader(data map[string]string) func() {
```
pmbethe09: name the return value 'cleanup'

buildifier/utils/utils.go
```diff
@@ -115,13 +116,39 @@ func SplitFilePath(filename string) (workspaceRoot, pkg string) {
 	return workspaceRoot, pkg
 }
 
+// getFileReader returns a *FileReader object that reads files from the local
+// filesystem if the workspace root is known.
+func getFileReader(workspaceRoot string) *warn.FileReader {
+	if workspaceRoot == "" {
+		return nil
```
pmbethe09: is there anyway to return a 'no-op' Filereader that doesn't require any special nil handling?

buildifier/utils/utils.go
```diff
@@ -115,13 +116,39 @@ func SplitFilePath(filename string) (workspaceRoot, pkg string) {
...
+
+		data, err := ioutil.ReadFile(path)
+		if err != nil {
+			return nil, err
+		}
+		return data, nil
+	}
+
+	fileReader := &warn.FileReader{}
+	fileReader.Init(readFile)
```
pmbethe09: If Init is always needed, just name a method

warn.NewFileReader

that does what Init does

buildifier/utils/utils.go
```diff
@@ -115,13 +116,39 @@ func SplitFilePath(filename string) (workspaceRoot, pkg string) {
...
+	if workspaceRoot == "" {
+		return nil
+	}
+
+	readFile := func(filename string) ([]byte, error) {
+	    // Use OS-specific path separators
+        filename = strings.ReplaceAll(filename, "/", string(os.PathSeparator))
+		path := filepath.Join(workspaceRoot, filename)
+
+		data, err := ioutil.ReadFile(path)
```
pmbethe09: since you do nothing with the returned values, just 

return ioutil.ReadFile(path)

buildifier/utils/utils.go
```diff
@@ -115,13 +116,39 @@ func SplitFilePath(filename string) (workspaceRoot, pkg string) {
...
+// filesystem if the workspace root is known.
+func getFileReader(workspaceRoot string) *warn.FileReader {
+	if workspaceRoot == "" {
+		return nil
+	}
+
+	readFile := func(filename string) ([]byte, error) {
+	    // Use OS-specific path separators
+        filename = strings.ReplaceAll(filename, "/", string(os.PathSeparator))
+		path := filepath.Join(workspaceRoot, filename)
```
pmbethe09: why is this indented?

warn/multifile.go
```diff
@@ -0,0 +1,48 @@
...
+// from the repository and cache the results.
+type FileReader struct {
+	cache    map[string]*build.File
+	readFile func(string) ([]byte, error)
+}
+
+// Init initializes a FileReader instance with an instance of function
+// that can read an arbitrary file in the repository using a path relative
+// to the workspace root.
+func (fr *FileReader) Init(readFile func(string) ([]byte, error)) {
```
pmbethe09: as I mentioned above, instead 

func NewFileReader(readFile func(string) ([]byte, error)) *FileReader {

warn/multifile.go
```diff
@@ -0,0 +1,48 @@
...
+	fr.cache = make(map[string]*build.File)
+	fr.readFile = readFile
+}
+
+// retrieveFile reads a Starlark file using only the readFile method
+// (without using the cache).
+func (fr *FileReader) retrieveFile(filename string) *build.File {
+	contents, err := fr.readFile(filename)
+	if err != nil {
+		return nil
```
pmbethe09: log this and the below error?

warn/multifile.go
```diff
@@ -0,0 +1,48 @@
...
+
+// GetFile reads a Starlark file from the repository or the cache.
+// Returns nil if the file is not found or not valid.
+func (fr *FileReader) GetFile(filename string) *build.File {
+	// Try to retrieve from the cache
+	if file, ok := fr.cache[filename]; ok {
+		return file
+	}
+	file := fr.retrieveFile(filename)
+	fr.cache[filename] = file
```
pmbethe09: do you want to cache a nil value?

warn/warn_deprecated.go
```diff
@@ -0,0 +1,87 @@
...
+		return pkg + "/" + label[1:]
+	default:
+		// External repositories are not supported
+		return ""
+	}
+}
+
+func readBzlFile(path string, fileReader *FileReader) (*build.File, bool) {
+	file := fileReader.GetFile(path)
+	if file == nil {
```
pmbethe09: return file, file != nil

warn/warn_deprecated.go
```diff
@@ -0,0 +1,87 @@
...
+		if !ok {
+			continue
+		}
+
+		loadedSymbols := make(map[string]*build.Ident)
+		for _, from := range load.From {
+			loadedSymbols[from.Name] = from
+		}
+
+		for _, stmt := range loadedFile.Stmt {
```
pmbethe09: do the body in a separate method to reduce indent?

warn/warn_deprecated.go
```diff
@@ -0,0 +1,87 @@
...
+			if !ok {
+				continue
+			}
+			docstringInfo := parseFunctionDocstring(str)
+			if !docstringInfo.deprecated {
+				continue
+			}
+
+			findings = append(findings,
+				makeLinterFinding(loadedSymbols[def.Name], fmt.Sprintf(
```
pmbethe09: a lot of linebreaks.  maybe if this was in a separate method would be clean on one line?

warn/warn_test.go

buildifier/utils/utils.go
vladmos: I chose to use a nil FileReader because 1. if a client doesn't need such functionality it doesn't need to provide a dummy FileReader object, just `nil`; 2. if a check doesn't make sense without a file reader it can easily check for `if fileReader == nil` and return.

buildifier/utils/utils.go

buildifier/utils/utils.go
vladmos: Because my WFH setup doesn't have gofmt running automatically on save yet :)

warn/multifile.go
vladmos: I tried to, but log messages appear in the stderr where the error messages normally appear, I thought it would be confusing for the users.

warn/multifile.go
vladmos: Yes, because if a file doesn't exist or is not a valid Starlark file, there's no point in reading it again.

warn/warn_deprecated.go
vladmos: Done.

warn/warn_deprecated.go
vladmos: Done.

warn/warn_deprecated.go
vladmos: Done.

vladmos:
Thanks for the thorough review :)

buildifier/utils/utils.go
pmbethe09: expecting/checking for nil is not a normal Go-idiom.

More often you provide a Noop struct/interface, and/or return (*foo, ok bool) where the bool tells you if the struct is valid.  This could be the case for an interface

buildifier/utils/utils.go

buildifier/utils/utils.go
vladmos: Should the callers that need to fix/lint a single file explicitly provide a dummy object that implements an interface with a single GetFile method? If so, should there be a singleton object, or should the callers explicitly create one, e.g. `warn.FixWarnings(..., warn.DummyFileReader{})`?

warn/multifile.go
```diff
@@ -0,0 +1,48 @@
...
+// from the repository and cache the results.
+type FileReader struct {
+	cache    map[string]*build.File
+	readFile func(string) ([]byte, error)
+}
+
+// NewFileReader initializes a FileReader instance with an instance of function
+// that can read an arbitrary file in the repository using a path relative
+// to the workspace root.
+func (fr *FileReader) NewFileReader(readFile func(string) ([]byte, error)) {
```
pmbethe09: this should not be a receiver, but return *FileReader

func  NewFileReader(readFile func(string) ([]byte, error)) *FileReader {

vladmos: Done.

buildifier/utils/utils.go
vladmos: Another question is even if I create an interface and several implementations (real, dummy), a caller may still pass `nil` as its value, and then   its methods will cause a segmentation fault error. I'll still need to check for nil to avoid that.

Buildifier is supposed to be used as a library, so by callers I actually mean other people who implement their own tools using buildifier. How can I enforce the usage of noop objects, especially given that the FileReader object is optional and using `nil` won't cause any breakages unless users actually try to run one of the multifile warnings?

PR #811, opened 2020-04-02
Remove string-escape from warnings, this option is no longer supported.

chatura-atapattu:
Support was removed in #762.


chatura-atapattu:
@googlebot I fixed it.

chatura-atapattu:
@googlebot I fixed it.

PR #813, opened 2020-04-04
feat: allow flags to be passed to buildozer via the NodeJS API

mattem:
Allows flags to be passed to the buildozer when consumed via the NodeJS API.

It may be nice to strongly type the flags instead of passing a list of strings, so there's a potential change here, but would add a little maintenance overhead.

It's a shame the options and flags args aren't swapped, as I think that provides a nicer method signature, but would introduce a breaking change.

cc @alexeagle


vladmos:
@alexeagle I'm not an expert in NodeJS, could you please take a look?

PR #814, opened 2020-04-09
Implement the unnamed macro warning

vladmos:
Warn if a macro function does not accept (either directly or indirectly via **kwargs) a keyword argument called `name`.

A macro is a function that calls a rule (either directly or indirectly via another macro).

The warning prints an example stack trace (statically analyzed) from the current function to the detected rule definition or a native rule call.

The warning tries to not load unnecessary extra files (e.g. if a function calls other functions from already read files, they will be checked before functions from not yet read files), it also doesn't analyze the same function twice.


warn/stacktrace.go
```diff
@@ -0,0 +1,65 @@
+package warn
```
laurentlb, line 1: The package needs a comment.
Please explain what it is and how it's used.

warn/warn_macro.go
```diff
@@ -0,0 +1,334 @@
...
+		}
+	}
+	return false
+}
+
+// fileData represents information about rules and functions extracted from a file
+type fileData struct {
+	rules     map[string]int
+	functions map[string]map[string]funCall
+	aliases   map[string]alias
```
laurentlb, line 40: Is it useful to have a separate field for aliases?

I could imagine that aliases would show up as rules/functions.

warn/warn_macro.go
```diff
@@ -0,0 +1,334 @@
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// fileData represents information about rules and functions extracted from a file
+type fileData struct {
+	rules     map[string]int
```
laurentlb, line 38: What's the int?

warn/warn_macro.go
```diff
@@ -0,0 +1,334 @@
...
+	functions map[string]map[string]funCall
```
laurentlb, line 39: Why are there nested maps?

warn/warn_macro_test.go
```diff
@@ -0,0 +1,463 @@
...
+  [native.cc_library() for i in x]
+`,
+		[]string{
+			`5: Macro function "macro" doesn't accept a keyword argument "name".
+
+Example stack trace (statically analyzed):
+  File "test/package/test_file.bzl", line 7, in macro
+    my_rule(...)
+  File "test/package/test_file.bzl", line 3
+    my_rule = rule(...)`,
```
laurentlb, line 33: I'm not convinced showing the stacktrace is really useful.

Maybe show just a line:
> "macro" is considered a macro because it calls the (rule|macro) my_rule on line 7.

warn/stacktrace.go
vladmos: Obsolete (removed the file together with stacktraces).

warn/warn_macro.go
vladmos: Yes, because an alias can refer to a loaded function (if it's been re-exported).

warn/warn_macro.go
vladmos: Line number (obsolete, not needed if a full stack trace is not printed).

warn/warn_macro.go
vladmos: The outer map is for functions defined in the file, the inner is for each distinct function call from inside the function definition (if a function calls `foo` several times, it's not useful to store `foo` more than once).

warn/warn_macro_test.go
vladmos: Done.

warn/warn_macro.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+// isMacroPrivate is the same as IsMacro except that it takes a set of already visited nodes to prevent
+// infinite recursion (recursion is not allowed in Starlark but can still crash Buildifier if not
+// handled properly)
+func (ma macroAnalyzer) isMacroPrivate(fn function, visited map[function]bool) (isMacro bool, fc *funCall) {
+	if visited[fn] {
+		// Don't visit the function again to prevent infinite recursion
+		return false, nil
+	}
```
laurentlb, line 214: Do we need both `visited` and `ma.cache`?
If we put a value in ma.cache before doing the recursion, and then update it, it should replace visited?

vladmos: The cache is only populated when buildifier knows about the current node whether it's a macro or not. It's useful if multiple functions in the file call the same helper functions, buildfier doesn't need to check them multiple times. It's a shared object for each function in the file.

`visited` prevents visiting the same node twice during the analysis (e.g. if there's recursion it will crash buildifier with stack overflow otherwise).

vladmos: In fact you're right, writing a negative result to the cache before analyzing is equivalent to using `visited`, so I've got rid of the latter.

PR #815, opened 2020-04-09
Refactor rewrites

vladmos:
Currently to format a file one should call two separate functions:

  * `build.Rewrite(file, &rewriteInfo)`: to modify the AST in-place (e.g. to sort the load arguments or to replace the division operator `/` with the integer division operator `//`). This function also takes an output argument of the type `*build.RewriteInfo` which is actually not useful (the diagnostics are being printed to the console but users are usually interested in whether the file has been reformatted or not, rather than how many division operators were fixed).
  * `build.Format(f)`: to actually convert the AST to bytes.

This is confusing for users who use buildifier as a library. If they don't know that they need to call `Rewrite` they may end up with incorrectly sorted load arguments if they add new loads.

The PR refactors the pipeline in the following way:

  * `Rewrite()` doesn't take the output argument anymore.
  * `Format()` rewrites and formats a file.
  *  In some rare cases (mostly in testing) it's required to format a file without rewriting (e.g. because some of the rewrites are not idempotent and it's required to format the same file twice). For such case a `FormatWithoutRewriting()` function is created (it's equal to the former `Format()`).

### DIFF
--- a/build/print.go
+++ b/build/print.go
@@ -39,7 +39,7 @@ func FormatWithoutRewriting(f *File) []byte {
 
 // Format rewrites the file and returns the formatted form of it.
 func Format(f *File) []byte {
-	rewrite(f)
+	Rewrite(f)
 	return FormatWithoutRewriting(f)
 }
 
--- a/build/print.go
+++ b/build/print.go
@@ -361,9 +361,9 @@ func (p *printer) expr(v Expr, outerPrec int) {
 	// If we are in the middle of an expression but not inside ( ) [ ] { }
 	// then we cannot just break the line: we'd have to end it with a \.
 	// However, even then we can't emit line comments since that would
-	// end the expression. This is only a concern if we have IsRewritten
+	// end the expression. This is only a concern if we have rewritten
 	// the parse tree. If comments were okay before this expression in
-	// the original input they're still okay now, in the absense of rewrites.
+	// the original input they're still okay now, in the absence of rewrites.
 	//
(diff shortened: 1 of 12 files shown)


build/print.go
```diff
@@ -354,7 +361,7 @@ func (p *printer) expr(v Expr, outerPrec int) {
 	// If we are in the middle of an expression but not inside ( ) [ ] { }
 	// then we cannot just break the line: we'd have to end it with a \.
 	// However, even then we can't emit line comments since that would
-	// end the expression. This is only a concern if we have rewritten
+	// end the expression. This is only a concern if we have IsRewritten
```
laurentlb, line 364: revert?

build/print.go
```diff
@@ -514,12 +521,12 @@ func (p *printer) expr(v Expr, outerPrec int) {
 
 	case *BinaryExpr:
 		// Precedence: use the precedence of the operator.
-		// Since all binary expressions format left-to-right,
+		// Since all binary expressions FormatWithoutRewriting left-to-right,
```
laurentlb, line 524: revert?

Please double-check all the other comments below, there seems to be some unintentional renaming.

vladmos: Thanks! That wasn't even Replace All, I used the Idea's refactoring tool to rename a function together with all its calls, but it also edits some random comments.

build/print.go

build/print.go
vladmos: Done.

PR #818, opened 2020-04-16
Fix file packages

vladmos:
Currently buildifier determines a file's package as its base name (relative to the workspace root). Although that's always true for BUILD files, .bzl files may be located not necessarily in a package root but in a subdirectory.

That didn't matter before, but now with the functionality of loading additional .bzl files that's being used by some analyzers, it's important to know the current file's package to resolve relative labels correctly.

To achieve that it's enough to locate a BUILD file closest to the initial file; all other loaded files packages can be determined by analyzing their label: e.g. if "//some/package:label/name.bzl" is loaded, the package is set to "some/package" and the label to "label/name.bzl" without even looking at BUILD files.

### DIFF
--- a/build/syntax.go
+++ b/build/syntax.go
@@ -20,6 +20,7 @@ package build
 // Syntax data structure definitions.
 
 import (
+	"fmt"
 	"strings"
 	"unicode/utf8"
 )
@@ -93,7 +94,8 @@ func stmtsEnd(stmts []Expr) Position {
 // A File represents an entire BUILD or .bzl file.
 type File struct {
 	Path          string // absolute file path
-	Pkg           string // optional; the package of the file
+	Pkg           string // optional; the package of the file (always forward slashes)
+	Label         string // optional; file path relative to the package name (always forward slashes)
 	WorkspaceRoot string // optional; path to the directory containing the WORKSPACE file
 	Type          FileType
 	Comments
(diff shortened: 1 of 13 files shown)


build/syntax.go
```diff
@@ -108,6 +110,14 @@ func (f *File) DisplayPath() string {
 	return f.Path
 }
 
+// Full label of the file, e.g. "//package:file.bzl"
+func (f *File) FullLabel() string {
+	if f.Pkg == "" {
+		return fmt.Sprintf("//%s", f.Label)
```
laurentlb, line 116: Why do we need this?
Shouldn't the label show up as `//:foo.bzl`?

vladmos: Actually what's more useful is a canonical path of a file (relative to the workspace root, with forward slashes only and no colon for package boundaries. It can be used in warning messages for referring a file (e.g. "the function is defined at //path/to/file.bzl"), but also for diagnostics (e.g. for resolving visibility of .bzl files).

I'll replace the function with CanonicalPath() that returns the same but with slashes instead of colons.

There's a `Path` field in the File struct, but it contains filesystem-specific slashes and may be not relative to the workspace root, e.g. if the workspace root is not known.

laurentlb: Maybe it shouldn't start with `//` if it's not a label?

laurentlb: Also the name of the function is misleading.

PR #820, opened 2020-04-28
remove attributes using regex

wix-andriusb:
This allows the user to remove dependencies, for example:

```
'remove deps @java.+' //server/...:%scala_library
```

will remove all dependencies which begin with the ```@java``` prefix in all bazel build files within //server


wix-andriusb:
@googlebot I signed it!

wix-andriusb:
@vladmos @laurentlb @pmbethe09 Any comments? Useful, not useful? This is something I would like to personally see in the tool to make working with my own repos easier.

PR #822, opened 2020-04-28
Authomatise GitHub release process

vladmos:
Add a script that can automatically create a draft release with an empty description. Generated draft releases contain everything else: correct tag, name and uploaded binaries for all platforms.


vladmos:
Yes, that's defined in .bazelrc.

PR #825, opened 2020-05-04
Don't apply unnamed-macro to private functions

vladmos:
Private functions can't be imported at BUILD files and therefore can't be macros.


laurentlb:
What happens if a public function calls a private function which calls a rule?

vladmos:
The change just doesn't start the top-level check on a function if it's private, that doesn't affect analysis of a function if it's triggered by inspecting another function.

I've added a test about that.

PR #828, opened 2020-05-05
Allow quoted arguments in buildozer's replace command

### LINKED ISSUE #823: `replace` fails with quoted old_value
When using a quoted old value `replace` fails with an exit code of 3 but it's:

1. Really hard to figure out what's going on (is there no `--verbose` flag?)
1. Inconsistent with `add`/`remove` since those work fine with quoted values

Repro:
`BUILD.bazel`
```
filegroup(
    name = "foolib6",
    srcs = ["foo6.py"],
)
```
Starting with this file, these commands all work:
✅ `buildozer 'add srcs "foo7.py"' //:foolib6`
✅ `buildozer 'remove srcs "foo6.py"' //:foolib6` 
✅ `buildozer 'replace srcs foo6.py "foo7.py"' //:foolib6`
✅ `buildozer 'replace srcs foo6.py foo7.py' //:foolib6`
But this does not:
🚫 `buildozer 'replace srcs "foo6.py" "foo7.py"' //:foolib6`

vladmos:
Fixes #823

### DIFF
--- a/buildozer/buildozer_test.sh
+++ b/buildozer/buildozer_test.sh
@@ -88,6 +88,14 @@ function test_add_dep_no_deps() {
 )'
 }
 
+function test_add_dep_quotes() {
+  run "$no_deps" 'add deps "//foo"' '//pkg:edit'
+  assert_equals 'go_library(
+    name = "edit",
+    deps = ["//foo"],
+)'
+}
+
 function test_add_dep_empty_deps() {
   run "$empty_deps" 'add deps //foo' '//pkg:edit'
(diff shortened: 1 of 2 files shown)


laurentlb:
The point of quoted arguments is to have space, right? Should this be tested?

vladmos:
IMO spaces are not supported yet, probably they should be.

The point was initially to allow some arguments to accept both string values (in quotes) and non-string (e.g. constants). For consistency all arguments (even those that can be only strings) should be accepted with quotes.

PR #829, opened 2020-05-07
Clarify unnamed macro warning

MrStevenFeldman:
Make the build warning for lack of `name` in a macro keyword arguments more clear.

### DIFF
--- a/WARNINGS.md
+++ b/WARNINGS.md
@@ -910,7 +910,7 @@ can potentially be empty.
 
 --------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


MrStevenFeldman:
@googlebot I signed it!

vladmos:
Thanks! Could you please also edit this header so that it matches the warning message? Otherwise it may be confusing.

https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#unnamed-macro

MrStevenFeldman:
Thanks for catching this, I have updated that file as well

PR #832, opened 2020-05-12
Add support for configuring build file names

Utsav2:
Dropbox uses `BUILD` file templates extensively so that users don't have to edit `BUILD` files for common attributes like `deps`. It would be useful for us to apply `buildozer` commands on those template files, so it'll be useful for `buildozer` to take an optional argument to modify files other than the hardcoded `BUILD/BUILD.bazel/BUCK` files.


Utsav2:
I notice that https://github.com/bazelbuild/buildtools/pull/168 tried adding the exact same thing, and doesn't seem like this functionality is desired. I wonder if maintainers would be amenable to this PR https://github.com/bazelbuild/buildtools/pull/836 which would unblock our use case since we'd be able to wrap buildozer and write some tools without needing to maintain a fork.

PR #833, opened 2020-05-15
Add option to sort lists by default

davidhao3300:
We work in a repo that does not care about dependency order, and we keep our lists in variables. A simple example of a typical file in our repo looks like:
```
DEPS = [
    ":a"',
    ":b",
    ":c",
]
function_call(deps=DEPS)
```
We'd like to keep lists sorted by default, and if we ever have a case of wanting to preserve order, we can choose to use the "do not sort" comment.

This PR adds an option to sort by lists by default, and it respects the existing "do not sort" comment.

### DIFF
--- a/build/rewrite.go
+++ b/build/rewrite.go
@@ -126,10 +126,16 @@ func doNotSort(x Expr) bool {
 	return hasComment(x, "do not sort")
 }
 
-// keepSorted reports whether x is marked with a comment containing
+// forceKeepSorted reports whether x is marked with a comment containing
 // "keep sorted", case-insensitive.
-func keepSorted(x Expr) bool {
+func forceKeepSorted(x Expr) bool {
 	return hasComment(x, "keep sorted")
+}
+
+// keepSorted checks whether lists are to be sorted by default, and respects comments as overrides of the default table
+// value
+func keepSorted(x Expr) bool {
+	return !doNotSort(x) && (tables.SortListsByDefault || forceKeepSorted(x))
 }
 
 // fixLabels rewrites labels into a canonical form.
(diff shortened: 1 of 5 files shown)


build/rewrite.go
```diff
@@ -450,7 +456,7 @@ func sortStringList(x Expr, context string) {
 		return
 	}
 
-	forceSort := keepSorted(list) || keepSorted(list.List[0])
+	forceSort := forceKeepSorted(list) || forceKeepSorted(list.List[0])
```
davidhao3300, line 459: this bit of logic requires users to explicitly comment "keep sorted" on lists with comments

davidhao3300:
@googlebot I signed it!

pmbethe09:
what problem does this solve that some combination of: 
tables.go and jsontables doesn't already solve in combo with `# buildifier: do not sort`

?

davidhao3300:
my understanding is that the code block
```
DEPS = [
    ":a"',
    ":b",
    ":c",
]
```
is handled by some subset of the following lines: https://github.com/bazelbuild/buildtools/blob/master/build/rewrite.go#L411-L434
These branches will only sort the list only if the list is prepended with a comment saying "keep sorted". In the current state, we would have to prepend every list with that comment, which is not ideal for us.

davidhao3300:
I'm not familiar with what jsontables is, and google doesn't seem to be returning relevant results. What is that?

laurentlb:
It's correct, this is currently not supported.

However, I feel like this is a very niche need and it doesn't match the style recommendations in Bazel. I would encourage you to build your own program (using Buildifier as a library) to perform this transformation; you'll just need to write a few lines of Go. I don't think this feature should belong to Buildifier.

davidhao3300:
Based on https://docs.bazel.build/versions/master/skylark/build-style.html, it's true that we should `Prefer to list dependencies directly, as a single list. Putting the “common” dependencies of several targets into a variable reduces maintainability, makes it impossible for tools to change the dependencies of a target and can lead to unused dependencies.`
I'd push back and say that the style guide doesn't forbid it, so I think there would be use for others who may keep lists in variables for whatever reason?

oreflow:
Closing as it's been stale for multiple years, please reopen and sync this to latest main if it's still relevant

PR #836, opened 2020-05-18
Export build file names variables

Utsav2:
Some organizations contain BUILD file templates or other files that resemble BUILD files. It's useful to be able to use buildifier to modify those variables. This PR simply exports the BUILD file name variables so that libraries can tweak these. This isn't an ideal solution, but would unblock many use cases without users needing to maintain custom patches.

### DIFF
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -851,15 +851,13 @@ func getGlobalVariables(exprs []build.Expr) (vars map[string]*build.AssignExpr) 
 }
 
 // When checking the filesystem, we need to look for any of the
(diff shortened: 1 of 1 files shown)


vladmos:
It can be confusing and error prone that both `BuildFileNames` and `BuildFileNamesSet` are exported, it's easy to update only one of them and forget about the other.

I don't think having a separate map `BuildFileNamesSet` is useful at all given that there are only few elements, could you please get rid of it and use the slice instead?

Utsav2:
@vladmos done!

vladmos:
Thanks!

PR #837, opened 2020-05-19
Implement a warning for the word 'skylark' in comments and docstrings

vladmos:


thomasvl:
fyi - https://github.com/bazelbuild/bazel/issues/8185 likely means this triggers a warning for anyone trying to reference bazel docs on Starlark itself.

PR #838, opened 2020-05-19
Thread through build file names to edit.go

Utsav2:
I tested this locally, and seemed to work for some commands like `set` and `add`

Even though buildozer had support for custom build file names, it didn't work for all commands like `add`. This was due to the assumption of `BUILD` file name in `edit.go`. This PR should fix that.


Utsav2:
@vladmos could you PTAL?

PR #840, opened 2020-05-27
Fix windows handling of path to build file containing "/"

joker-eph:
Copybara invokes buildozer by passing the relative path to a
BUILD file using Unix separator.
This was broken in 19db42aa7a by the introduction
of `filepath.FromSlash` when evaluating the path to BUILD files.
Instead we can make it more tolerant by accepting both forms.

### DIFF
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -1076,6 +1076,8 @@ func appendCommands(opts *Options, commandMap map[string][]commandsForTarget, ar
 		for _, buildFileName := range BuildFileNames {
 			if strings.HasSuffix(target, filepath.FromSlash("/"+buildFileName)) {
 				target = strings.TrimSuffix(target, filepath.FromSlash("/"+buildFileName)) + ":__pkg__"
+			} else if strings.HasSuffix(target, "/"+buildFileName) {
+				target = strings.TrimSuffix(target, "/"+buildFileName) + ":__pkg__"
 			}
 		}
 		var buildFiles []string
--- a/edit/buildozer_test.go
+++ b/edit/buildozer_test.go
@@ -124,7 +124,7 @@ type targetExpressionToBuildFilesTestCase struct {
 	buildFiles      []string
 }
 
-func runTestTargetExpressionToBuildFiles(t *testing.T, buildFileName string) {
+func setupTestTmpWorkspace(t *testing.T, buildFileName string) (tmp string) {
 	tmp, err := ioutil.TempDir("", "")
 	if err != nil {
 		t.Fatal(err)
@@ -136,7 +136,6 @@ func runTestTargetExpressionToBuildFiles(t *testing.T, buildFileName string) {
 		t.Fatal(err)
 	}
 
-	defer os.RemoveAll(tmp)
 	if err := os.MkdirAll(filepath.Join(tmp, "a", "b"), 0755); err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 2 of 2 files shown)


edit/buildozer.go
```diff
@@ -1074,8 +1074,8 @@ func appendCommands(opts *Options, commandMap map[string][]commandsForTarget, ar
 	commands, targets := parseCommands(args)
 	for _, target := range targets {
 		for _, buildFileName := range BuildFileNames {
-			if strings.HasSuffix(target, filepath.FromSlash("/"+buildFileName)) {
-				target = strings.TrimSuffix(target, filepath.FromSlash("/"+buildFileName)) + ":__pkg__"
+			if strings.HasSuffix(target, "/"+buildFileName) {
```
vladmos, line 1077: If `target` is a windows-style path with backslashes, this won't work, or am I wrong?

edit/buildozer_test.go
```diff
@@ -205,6 +210,47 @@ func TestTargetExpressionToBuildFiles(t *testing.T) {
...
+			if value[0].target != tc.target {
+				t.Errorf("appendCommands for buildfile %s yielded target %s, expected %s", tc.buildFiles, value[0].target, tc.target)
+			}
+		}
+	}
+}
+
+func TestAppendCommands(t *testing.T) {
+        runTestAppendCommands(t, "BUILD")
+        runTestAppendCommands(t, "BUILD.bazel")
```
vladmos, line 251: Please iterate over `BuildFileNames` instead, like here: https://github.com/bazelbuild/buildtools/pull/838/files#diff-10c338c9802b1617e469141df410c9ebL204

joker-eph: Sure! I copy-pasted the test right above (TestTargetExpressionToBuildFiles line 203, do you want me to update in the same PR?)

joker-eph: Nevermind I see that the other PR is actually updating the code I copy/pasted!

edit/buildozer.go
joker-eph: Somehow I didn't push the last version I had where I kept the original code + the "unix" variant. Let me know if there is a more elegant way you'd see here.

edit/buildozer.go
```diff
@@ -1076,6 +1076,8 @@ func appendCommands(opts *Options, commandMap map[string][]commandsForTarget, ar
 		for _, buildFileName := range BuildFileNames {
 			if strings.HasSuffix(target, filepath.FromSlash("/"+buildFileName)) {
 				target = strings.TrimSuffix(target, filepath.FromSlash("/"+buildFileName)) + ":__pkg__"
+			} else if strings.HasSuffix(target, "/"+buildFileName) {
+				target = strings.TrimSuffix(target, "/"+buildFileName) + ":__pkg__"
```
vladmos, line 1080: This code now replaces the trailing "/BUILD" or "\BUILD" (but in non-windows environment only "/BUILD") with ":__pkg__", leaving all other slashes as they are. That doesn't look correct to me, the slashes should either always be forward if they represent targets or system specific if they represent files. The `target` variable is later being interpreted as a modified file name but sometimes as a label (that should probably be fixed).

I think it's better to replace forward slashes in `target` with system-specific, and have just one call to `strings.HasSuffix`, that should fix your problem, make the behavior more predictable (`target` will always have system-specific slashes), and also make the function slightly more readable.

joker-eph: I gave it a try, let me know if I got it right. I tried to do:

> I think it's better to replace forward slashes in target with system-specific

but I think this goes against:

> the slashes should either always be forward if they represent targets 

Because now even when `target` is representing a target we have `\` on Windows.

joker-eph:
Anything else here @vladmos ?

vladmos:
Sorry, for some reason I thought I'd merged it already. Merging now.

PR #842, opened 2020-05-28
Change generate_tables from genrule to .bzl rule.

tomlu:
genrules do not work on Windows without a msys dependency.

The tool had to be changed as well, since ctx.actions.run does not
support piping stdin and stdout to files. I couldn't find any
other dependencies on the tool in this repository, but the visibility
is public so if any external dependencies have decided to use
this they will break.


generatetables/generate_tables.go
```diff
@@ -17,6 +17,7 @@ distributed under the License is distributed on an "AS IS" BASIS,
 package main
 
 import (
+    "flag"
```
pmbethe09, line 20: gofmt ?

lang/defs.bzl
```diff
@@ -0,0 +1,24 @@
+def _generate_tables_impl(ctx):
+    args = ctx.actions.args()
+    args.add("--input", ctx.file.src)
```
pmbethe09, line 3: opt: can use single '-' in go flags '-input'

generatetables/generate_tables.go
tomlu: Done.

lang/defs.bzl
tomlu: Done.

PR #844, opened 2020-05-28
Enable buildifier warnings on the CI

vladmos:


vladmos:
I'll need to fix the existing warnings before merging it.

PR #851, opened 2020-05-29
support floats in buildifier2

garymm:


vladmos:
Thanks! `buildifier2` is experimental and is not guaranteed to support everything though.

PR #856, opened 2020-06-04
Restrict dict elements to KeyValueExpr objects

vladmos:
`DictExpr` objects may contain subnodes of any types that implement `Expr`. This commit restricts it to `KeyValueExpr` objects only, which makes the AST more correct and traversing it easier.

### DIFF
--- a/build/parse.y
+++ b/build/parse.y
@@ -568,9 +568,9 @@ primary_expr:
 	}
 |	'{' keyvalues '}'
 	{
-		exprValues := make([]Expr, len($2))
-		for i, kv := range $2 {
-			exprValues[i] = Expr(kv)
+		exprValues := make([]Expr, 0, len($2))
+		for _, kv := range $2 {
+			exprValues = append(exprValues, Expr(kv))
 		}
 		$$ = &DictExpr{
 			Start: $1,
--- a/build/parse.y
+++ b/build/parse.y
@@ -25,6 +25,8 @@ package build
 	// partial syntax trees
 	expr      Expr
 	exprs     []Expr
+	kv        *KeyValueExpr
+	kvs       []*KeyValueExpr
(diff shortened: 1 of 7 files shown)


build/parse.y
```diff
@@ -566,11 +568,15 @@ primary_expr:
 	}
 |	'{' keyvalues '}'
 	{
+		exprValues := make([]Expr, len($2))
```
pmbethe09, line 571: prefer capacity make and append to avoid potential off by 1

exprValues := make([]Expr, 0, len($2))

then 
exprValues = append(exprValues, Expr(kv))

edit/edit.go
```diff
@@ -639,22 +631,16 @@ func SelectListsIntersection(sel *build.CallExpr, pkg string) (intersection []bu
...
 	for _, keyVal := range dict.List[1:] {
-		if keyVal, ok := keyVal.(*build.KeyValueExpr); ok {
-			if val, ok := keyVal.Value.(*build.ListExpr); ok {
-				intersection = ComputeIntersection(intersection, val.List)
-				if len(intersection) == 0 {
-					return intersection
-				}
-			} else {
-				return nil
+		if val, ok := keyVal.Value.(*build.ListExpr); ok {
```
pmbethe09, line 640: prefer to invert and return nil to reduce indent

val, ok := keyVal.Value.(*build.ListExpr)
if !ok {
  return nil
}
intersection = ...

vladmos: Done.

build/parse.y

Why can it be off by 1? Both slices are guaranteed to have the same length.

pmbethe09: I agree it won't here, but if you use [i] style you *might* make a mistake.  If you use append style, if you make a mistake, you just force a re-allocation

PR #873, opened 2020-07-14
Decree package_group.packages as a non-label attribute

vladmos:
The attribute doesn't contain labels, the values shouldn't be shortened.

### DIFF
--- a/edit/types.go
+++ b/edit/types.go
@@ -64,8 +64,8 @@ func IsStringDict(attr string) bool {
 // ContainsLabels returns true for all attributes whose type is a label or a label list.
 func ContainsLabels(kind, attr string) bool {
 	if kind == "package_group" && attr == "packages" {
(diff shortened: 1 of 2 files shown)


edit/types.go
```diff
@@ -62,7 +62,11 @@ func IsStringDict(attr string) bool {
 }
 
 // ContainsLabels returns true for all attributes whose type is a label or a label list.
-func ContainsLabels(attr string) bool {
+func ContainsLabels(kind, attr string) bool {
+	if kind == "package_group" && attr == "packages" {
+	  // "package_group" is a special rule and its "packages" attribute is not a list of labels.
```
laurentlb, line 67: indentation is surprising, make sure you run gofmt

vladmos: I updated IDEA and that disabled autoformatting on save and two spaces look identical to a tabulation. Fixed.

PR #879, opened 2020-07-24
Add warning for calls to `provider` without `doc` or `fields`.

laurentlb:


laurentlb:
(fyi, @alandonovan, @meisterT)

WARNINGS.md
```diff
@@ -796,6 +797,38 @@ may never see the warning.
...
+```python
+provider(
+    doc = "Some documentation.",
+    fields = {
+        "field1": "Documentation for field1",
+        "field2': "Documentation for field2",
+    })
+``
+
+Note that specifying a list of fields is a breaking change. It is an error if a
```
alandonovan, line 825: Does an explicit fields=None work to (a) indicate a schemaless struct and (b) suppress the warning? If so, we should document it. If not, we should implement it (and document it).

warn/warn_bazel_api.go
```diff
@@ -1008,3 +1008,33 @@ func keywordPositionalParametersWarning(f *build.File) []*LinterFinding {
...
+		_, _, fields := getParam(call.List, "fields")
+		_, _, doc := getParam(call.List, "doc")
+		// doc can also be the first positional argument
+		hasPositional := false
+		if len(call.List) > 0 {
+			if _, ok := call.List[0].(*build.AssignExpr); !ok {
+				hasPositional = true
+			}
+		}
+		if fields == nil || (doc == nil && !hasPositional) {
```
alandonovan, line 1033: redundant parens

warn/warn_bazel_api.go
```diff
@@ -1008,3 +1008,33 @@ func keywordPositionalParametersWarning(f *build.File) []*LinterFinding {
...
+			findings = append(findings, makeLinterFinding(call,
```
alandonovan, line 1034: Make the error message specific about which one is missing (or both).

WARNINGS.md
laurentlb: Yes, it's supported. I've updated the text here.

We could actually declare `struct` as `struct = provider()`.

warn/warn_bazel_api.go
laurentlb: Done

WARNINGS.md
```diff
@@ -796,6 +797,45 @@ may never see the warning.
...
 
+## <a name="provider-params"></a>Calls to 'provider' should specify a list of fields and a documentation
+
+  * Category name: `provider-params`
+  * Automatic fix: no
+
+Calls to `provider` should specify a documentation string and a list of fields:
+
+```python
+fooInfo = provider(
```
alandonovan, line 808: Let's give a better example, for instance:
```
server_address = provider(
   "The address of an HTTP server. Fields are host (string) and port (int)."
  fields = ["host", "port"])
```

laurentlb: Thanks, updated.
I've updated the name to ServerAddressInfo, because that's the convention (CamlCase as it's considered as a type).

WARNINGS.md
```diff
@@ -796,6 +797,44 @@ may never see the warning.
...
+```
+
+Fields should also be documented when needed:
+
+```python
+ServerAddressInfo = provider(
+    "The address of an HTTP server.",
+    fields = {
+        "host": "string, e.g. 'example.com'",
+        "port: "int",
```
alandonovan, line 820: Missing quote. Also missing documentation:

```
"port": "int, a TCP port number"
```

laurentlb: done

PR #884, opened 2020-08-12
Automatically remove superfluous parens

laurentlb:
This transformation applies only in a few safe cases. Parens are
preserved in multiline expressions, in presence of a comment, or in
complex expressions.

This change can modify the formatting of existing files, but such cases
should be relatively rare.

### DIFF
--- a/build/rewrite.go
+++ b/build/rewrite.go
@@ -18,12 +18,13 @@ distributed under the License is distributed on an "AS IS" BASIS,
 package build
 
 import (
-	"github.com/bazelbuild/buildtools/tables"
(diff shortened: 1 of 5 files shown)


laurentlb:
Statistics from the Google codebase:
* 1.7% of the bzl files, and
* 0.3% of the BUILD files are affected by this change

@vladmos What do you think?

(googlers can ​see go/starlark-remove-useless-parens to discuss the rollout in Google)

laurentlb:
Fixed a bug, added tests.
PTAL

PR #889, opened 2020-08-25
`Buildozer` return 1 if `NumIO` is greater than 0

danielmirchandani:
The motivation behind this is an infinite wait I ran into when I tried calling `Buildozer()` with `Options{}`. `Buildozer()` doesn't start any go-routines, then waits forever for data to come from any of the go-routines. We can't limit access to the `Options` struct, and someone might set this value less than 1 by mistake, so immediately stopping is probably a better behavior.

No tests because "buildozer_command_file_test.go" and "buildozer_test.go" don't have any tests that cover `Buildozer()`.


vladmos (approved):

vladmos:
Thanks!

PR #891, opened 2020-08-26
Fix wrong error msg & doc about --mode=check

artem-zinnatullin:
It's a typo in both error message and the doc, should be `--mode=check` instead of `--type=check` :)

I tried using it with `--type=check` and was getting following error:

```console
buildifier -v --format=json --type=check myfile.bzl
buildifier: unrecognized input type check; valid types are build, bzl, workspace, default, auto
```


vladmos (approved):
Thanks, the `--type` flag toggles the input file type.

PR #897, opened 2020-09-03
buildozer: return 0 if only readonly commands have been executed

### LINKED ISSUE #888: Feature request: Buildozer option to exit 0 instead of 3 when no changes are made
I'm using buildozer's print command to print some labels and attributes. buildozer ran successfully, but to my surprise, its exit status was 3.  [Checking the documentation](https://github.com/bazelbuild/buildtools/blob/master/buildozer/README.md#error-code) I learnt that buildozer returns "0 on success, if changes were made" and "3 on success, when no changes were made".  This is somewhat counterintuitive if you are intentionally using buildozer with read only commands such a print. 

For these kinds of read only use cases, it would be helpful if buildozer had an option to always return 0 for any success, regardless of whether changes were made.

As it happens, I'm using buildozer inside xargs, so return codes are extra confusing and hard to handle. Everything appeared successful, but xargs returned 123, because buildozer returned 3.

vladmos:
Fixes #888

### DIFF
--- a/buildozer/README.md
+++ b/buildozer/README.md
@@ -194,8 +194,8 @@ buildozer 'add new_attr def_val' //:%cc_binary
 
 ### Print commands
 
-They work just like the edit commands. Expect a return code of 3 as they are not
-modifying any file.
+These commands are not modifying files, Buildifier returns 0 after a successful
+execution.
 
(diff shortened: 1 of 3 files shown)


laurentlb (approved):
lgtm

PR #898, opened 2020-09-03
Format files with a builtin buildifier correctly

### LINKED ISSUE #894: [buildozer] the built-in formatter doesn't play well when adding a call to another rule or function
If I have some base BUILD file like so:

```
$ cat BUILD
some_macro(
    name = "example",
    srcs = glob(["**"]),
    some_attr = another_macro(
        attr1 = "foo",
        attr2 = "bar",
    ),
)
```

and I want to modify the value of `some_attr`, I can do that like so:

```
buildozer 'set some_attr yet_another_macro(attr1="baz",attr2="quux",),' //:example
```

... this results in an ugly (non-formatted) starlark:

> _Note: I'm unaware of a cleaner way to add a call to another rule or macro as the value to a property. If I'm missing something obvious, please correct me._

```
➜ cat BUILD
some_macro(
    name = "example",
    srcs = glob(["**"]),
    some_attr = yet_another_macro(attr1="baz",attr2="quux",),
)
```

Naturally, you might want to use `buildifier` to fix this:

```
bazel run //:buildifier # or "buildifer ..." if you have it installed and available in PATH
```

If this is being executed through Bazel, this can be costly if it hasn't been built before, such as when this entire process (running buildozer to modify some attrs) is ran in an ephemeral directory. To combat that, a user may turn to Buildozer's docs and discover that there is a built-in formatter available.

As noted in the documentation, if the value of the `-buildifier` option is empty, then the built-in formatter is used.

```
buildozer --buildifier '' 'set some_attr yet_another_macro(attr1="baz",attr2="quux",),' //:example
```

This, unfortunately, does not format the call to `yet_another_macro` correctly (it remains on one line).

vladmos:
An AST after buildozer transformations may be not correct, e.g. it may have an `Ident` node with a complex expression inside. In order to format such AST correctly it should be printed to a string, re-parsed correctly and printed again.

Fixes #894

### DIFF
--- a/buildozer/buildozer_test.sh
+++ b/buildozer/buildozer_test.sh
@@ -946,6 +946,19 @@ function test_set_if_absent_present() {
 )'
 }
 
+function test_set_custom_code() {
+  in='cc_test(name = "a")'
+
+  run "$in" 'set attr foo(a=1,b=2)' '//pkg:a'
+  assert_equals 'cc_test(
+    name = "a",
+    attr = foo(
+        a = 1,
+        b = 2,
+    ),
+)'
+}
+
 function assert_output() {
   echo "$1" > "expected"
   diff -u "expected" "$log" || fail "Output didn't match"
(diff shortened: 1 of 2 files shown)


laurentlb (approved):
lgtm

PR #901, opened 2020-09-08
Deduplicate lists that are sortable in general but not sortable because of a comment

vladmos:
Some string lists are not sorted by buildifier because of a comment (either explicit "do not sort" or just any comment inside an otherwise sortable list). Such lists should be deduplicated anyway.


laurentlb (approved):
lgtm

Nice, thanks a lot

PR #905, opened 2020-09-10
Restore removed methods

vladmos:
Some methods have been moved from `edit` to `labels` in #903. They are still being used internally, so they should be restored for backward compatibility.


laurentlb (approved):
lgtm

PR #906, opened 2020-09-10
Restore the --shorten_label flag

vladmos:
#903 removed support of the `--shorten_labels` flag, however there are still few usages of it, and it should be restored.

At the same time, `ShortenLabel` was also used to compare labels, where 1. the flag shouldn't play any role, 2. it's not necessary because `labels.Equal` can correctly compare long and short labels.


laurentlb (approved):
lgtm

PR #910, opened 2020-09-21
Update the warning message for unnamed macros

vladmos:


warn/warn_macro_test.go
```diff
@@ -215,13 +215,25 @@ def not_macro(x):
 		[]string{
 			`4: By convention "macro1" should have a keyword argument called "name".
 
-It is considered a macro because it calls a rule or another macro "abc" on line 5.`,
+It is considered a macro because it calls a rule or another macro "abc" on line 5.
```
laurentlb, line 218: Suggested text:

> 4: The macro "macro1" should have a keyword argument called "name".
>
> By convention, every public macro needs a "name" argument (even if it doesn't use it). This is important for tooling and automation.
> - If this function is a helper function that's not supposed to be used outside of this file, please make it private (e.g. rename it to "_macro1").
> - Otherwise, add a "name" argument. If possible, use that name when calling other macros/rules.

laurentlb (approved):

laurentlb:
You could also expand the documentation (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#unnamed-macro). For example, add examples.

warn/warn_macro_test.go
vladmos: Thanks for the suggestion! I kept the line explaining why the current function is a macro though, it can be quite handy in some cases.

laurentlb: Great, thanks! I didn't intend to remove it.

PR #911, opened 2020-09-22
Allow += on top level lists

vladmos:
The `redefined-variable` check disallows modifying top-level variables, however the `+=` operator on lists doesn't redefine them but just modifies them in-place, that should be allowed.


laurentlb (approved):
lgtm

Issue #852, opened 2020-05-30
buildifier support for other Starlark language features


garymm:
We use the Starlark Go interpreter for some config files (not Bazel BUILD files), and we have a few additional Starlark features enabled that buildifier doesn't support, in particular we have these settings:

```go
resolve.AllowLambda = true
resolve.AllowNestedDef = true
resolve.AllowFloat = true
resolve.AllowSet = true
```

This is a request to support these features in buildifier.

vladmos:
Do you need these features in the formatter?

They are not supported by Bazel, but at least some of them should be supported by Buildifier, e.g. there's a test for nested def statements: https://github.com/bazelbuild/buildtools/blob/master/build/testdata/051.in

garymm:
I'm fairly sure I tried it yesterday and buildifer failed to format files with lambdas and floats.

vladmos:
I've just checked that all the features mentioned above except floats are supported.

garymm:
It doesn't support a lambda with **kwargs:

```
➜  echo 'x = lambda **kwargs: kwargs.get("foo")' | buildifier
<stdin>:1:14: syntax error near **
<stdin> # reformat
```

Starlark Go does support this.
Shall I file a separate bug about this, or re-open this one?

vladmos:
Thanks, it's fine to track it here.

PR #924, opened 2020-11-18
buildifier2: support lambda expressions

fenollp:
Note that in my own usage of buildifier2 the example I provided gets `build.FormatString`-ed onto just one line:
```starlark
TriggerActionAfterProbe(probe = ("http", "response"), predicate = lambda State, response: all([response["request"]["method"] == "GET", response["request"]["url"].find("/posts/") != -1, response["request"]["url"].endswith("/comments"), response["status_code"] in range(200, 299)]), action = lambda State, response: None)
```
I wonder what's different with this test suite. And now writing all this I'm guessing either
* my code isn't being exercised by the tests, or
* the tests use different formatting settings
I'll look into it (any ideas though?)

Anyway, buildifier2 would crash without this clause.

### DIFF
--- a/build/testdata/006.build.golden
+++ b/build/testdata/006.build.golden
@@ -22,3 +22,14 @@ l = lambda x, y, *args, **kwargs: f(
     *args,
     **kwargs
 )
+
+TriggerActionAfterProbe(
+    action = lambda State, response: None,
+    predicate = lambda State, response: all([
+        response["request"]["method"] == "GET",
+        response["request"]["url"].find("/posts/") != -1,
+        response["request"]["url"].endswith("/comments"),
+        response["status_code"] in range(200, 299),
+    ]),
+    probe = ("http", "response"),
+)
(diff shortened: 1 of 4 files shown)


pmbethe09:
AFAIK starlark does not support lambda expressions

https://docs.bazel.build/versions/master/skylark/language.html#differences-with-python

fenollp:
True, Bazel's Starlark does not support lambdas, however they are defined in [this spec](https://github.com/google/starlark-go/blob/e55f603d8c797b182566012861c5784481d00df4/doc/spec.md#lambda-expressions) and implemented in the Go version of the interpreter.
Also, you've reviewed PRs relating to lambda expressions (#860) so I'm not sure why you mention this?

cc #852

vladmos:
Buildifier now supports lambda expressions, but as buildifier2 is experimental and not being maintained I didn't fix lambda expressions there.

vladmos (approved):
Thanks!

PR #925, opened 2020-11-30
Allow multiple command files to be passed into buildozer

jroylance:
This updates buildozer to accept multiple command files, reading and executing the instructions in each of them. An example usage would be `buildozer -f file_1 -f file_2`. This should be backwards compatible with existing usages. 

It looks like `flag` doesn't support multiple instances of the same flag, so I've implemented a custom solution in `buildozer/main.go` [taken from StackOverflow](https://stackoverflow.com/questions/28322997/how-to-get-a-list-of-values-into-a-flag-in-golang).

### DIFF
--- a/buildozer/main.go
+++ b/buildozer/main.go
@@ -29,16 +29,25 @@ import (
 	"github.com/bazelbuild/buildtools/tables"
 )
 
+type flagArray []string
+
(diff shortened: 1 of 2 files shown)


vladmos (approved):
Thanks. Could you please sync the branch with master?

jroylance:
@vladmos I've synced the branch with master

PR #927, opened 2020-12-03
Support Kotlin Rules

jongerrish:
I've added better support for jdeps in the Bazel Kotlin rules

see: https://github.com/bazelbuild/rules_kotlin/pull/415

Namely we now specify a `--direct_dependencies` flag to the KotlinBuilder to match how Bazel's JavaBuilder works. This is basically the API that unused_deps uses to understand a targets direct dependencies.

This changes the unused_deps tool to also include `kt_*` rules for processing and also needs to check `KotlinCompile` mnemonics since Kotlin targets with no Java sources will not invoke a `Javac` action.

### DIFF
--- a/unused_deps/unused_deps.go
+++ b/unused_deps/unused_deps.go
@@ -57,7 +57,7 @@ var (
 def _javac_params(target, ctx):
     params = []
     for action in target.actions:
-        if not action.mnemonic == "Javac":
+        if not action.mnemonic == "Javac" and not action.mnemonic == "KotlinCompile":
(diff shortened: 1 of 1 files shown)


jongerrish:
I was hoping there was a better way rather than hard coding the mnemonic and rules prefixes so unused deps wouldn't have knowledge of the underlying rules details, not sure if that bothers you so much but if there's a better way I'm happy to make suggested changes.

vladmos (approved):
Thanks!

PR #929, opened 2020-12-07
feat(buildifier): add buildifier_test

### LINKED ISSUE #928: Buildifier should have a test rule
A common implementation of tools like Buildifier are at the CI step, rejecting a submission if files are not formatted with the tool. Currently, consumers must perform their own logic to execute a "check" to ensure files are not out of policy. It would be convenient to have a test rule that executes Buildifier over a set of inputs such that it can be picked up by something like `bazel test //...` and not require any custom CI step.

sudoforge:
```
This patch adds a `buildifier_test` rule, as well as a factory-based
workflow for generating both rules to reduce the burden of maintenance.

closes #928
```
---

This looks like a lot, but there are a few things to keep in mind:

* The `buildifier()` implementation has not been changed at all
* The factory-based approach taken here _really_ simplifies the burden of maintenance, as both the executable and test rules share basically, well, everything -- and without the factory-based approach, there'd be a lot of manual duplication.

---

Usage looks like this:

**`//foo:BUILD.bazel`**

```
STARLARK_FILE_GLOB = [
    "**/*.BUILD",
    "**/*.bzl",
    "**/*.sky",
    "**/BUILD",
    "**/BUILD.*.bazel",
    "**/BUILD.*.oss",
    "**/BUILD.bazel",
    "**/WORKSPACE",
    "**/WORKSPACE.*.bazel",
    "**/WORKSPACE.*.oss",
    "**/WORKSPACE.bazel",
    "*.BUILD",
    "*.bzl",
    "*.sky",
    "BUILD",
    "BUILD.*.bazel",
    "BUILD.*.oss",
    "BUILD.bazel",
    "WORKSPACE",
    "WORKSPACE.*.bazel",
    "WORKSPACE.*.oss",
    "WORKSPACE.bazel",
]

filegroup(
    name = "starlark_files",
    srcs = glob(STARLARK_FILE_GLOB),
)
```

**`//:BUILD.bazel`**

```
load("@com_github_bazelbuild_buildtools//buildifier:def.bzl", "buildifier_test")

buildifier_test(
    name = "check_starlark_format",
    size = "small",
    timeout = "short",
    srcs = [
        "//foo:starlark_files",
    ],
    mode = "check",
)
```

Then run the test: `bazel test //:check_starlark_format`.

---

Because the `glob` can't cross package boundaries the simplest approach is to define that same `filegroup` in each package. To make this easy and as simple as possible, I've created a convenience macro for my own use that I'd be happy to add into this repository. 

> I wasn't able to find a way to make the test break out of the sandbox in a similar way to how we use `$BUILD_WORKSPACE_DIRECTORY` with the executable rule, but suggestions for a consistent approach are welcome

It's pretty straightforward (in this case, my test lives at `//tools/buildifier:check_starlark_files`):

```
load("//lib/bazel:constants.bzl", "STARLARK_FILE_GLOB")

def buildifier_filegroup(name, visibility = []):
    """
    Convenience macro to a filegroup of starlark files for use with
    //tools/buildifier:check_starlark_format

    Args:
      name:        The name of the target. Must be "starlark_files".
      visibility:  The visibility of the generated target(s). Should not include //tools/buildifier:__pkg__.
    """

    # Require the user to define the name for easier scripting (especially via buildozer)
    if name != "starlark_files":
        fail("Attribute 'name' must be equal to: starlark_files")

    if "//tools/buildifier:__pkg" in visibility:
        fail("Attribute 'visibility' should not contain value: //tools/buildifier:__pkg__")

    native.filegroup(
        name = name,
        srcs = native.glob(STARLARK_FILE_GLOB),
        visibility = [
            "//tools/buildifier:__pkg__",
        ] + visibility,
    )
```

This simplifies the definition needed in each package to a measly:

```
buildifier_filegroup(name = "starlark_files")
```

---

I haven't yet added documentation for using this; want to get feedback and align on the implementation before writing the docs. **Please do not merge this before documentation has been added.**

### DIFF
--- a/buildifier/buildifier.bzl
+++ b/buildifier/buildifier.bzl
@@ -1,0 +1,41 @@
+"""
+The module defines buildifier as a Bazel rule.
+"""
+
+load(
+    "//buildifier/internal:factory.bzl",
+    "buildifier_attr_factory",
+    "buildifier_impl_factory",
+)
+
+def _buildifier_impl(ctx):
+    return [buildifier_impl_factory(ctx)]
+
+_buildifier = rule(
+    implementation = _buildifier_impl,
+    attrs = buildifier_attr_factory(),
+    executable = True,
+)
+
+def buildifier(**kwargs):
+    """
+    Wrapper for the _buildifier rule. Adds 'manual' to the tags.
+
+    Args:
+      **kwargs: all parameters for _buildifier
+    """
+
+    tags = kwargs.get("tags", [])
+    if "manual" not in tags:
+        tags.append("manual")
+        kwargs["tags"] = tags
+    _buildifier(**kwargs)
+
+def _buildifier_test_impl(ctx):
+    return [buildifier_impl_factory(ctx, test_rule = True)]
+
+buildifier_test = rule(
+    implementation = _buildifier_test_impl,
+    attrs = buildifier_attr_factory(True),
+    test = True,
+)
--- a/buildifier/def.bzl
+++ b/buildifier/def.bzl
@@ -1,112 +1,12 @@
 """
-The module defines buildifier as a Bazel rule.
+This module is the public interface to buildifier rules
 """
 
-load("@bazel_skylib//lib:shell.bzl", "shell")
-
-def _buildifier_impl(ctx):
-    # Do not depend on defaults encoded in the binary.  Always use defaults set
-    # on attributes of the rule.
-    args = [
-        "-mode=%s" % ctx.attr.mode,
-        "-v=%s" % str(ctx.attr.verbose).lower(),
-    ]
-
-    if ctx.attr.lint_mode:
-        args.append("-lint=%s" % ctx.attr.lint_mode)
-    if ctx.attr.lint_warnings:
-        if not ctx.attr.lint_mode:
-            fail("Cannot pass 'lint_warnings' without a 'lint_mode'")
-        args.append("--warnings={}".format(",".join(ctx.attr.lint_warnings)))
-    if ctx.attr.multi_diff:
-        args.append("-multi_diff")
-    if ctx.attr.diff_command:
-        args.append("-diff_command=%s" % ctx.attr.diff_command)
-    if ctx.attr.add_tables:
-        args.append("-add_tables=%s" % ctx.file.add_tables.path)
-
-    exclude_patterns_str = ""
-    if ctx.attr.exclude_patterns:
-        exclude_patterns = ["\\! -path %s" % shell.quote(pattern) for pattern in ctx.attr.exclude_patterns]
-        exclude_patterns_str = " ".join(exclude_patterns)
-
-    out_file = ctx.actions.declare_file(ctx.label.name + ".bash")
-    substitutions = {
-        "@@ARGS@@": shell.array_literal(args),
-        "@@BUILDIFIER_SHORT_PATH@@": shell.quote(ctx.executable._buildifier.short_path),
-        "@@EXCLUDE_PATTERNS@@": exclude_patterns_str,
-    }
-    ctx.actions.expand_template(
-        template = ctx.file._runner,
-        output = out_file,
-        substitutions = substitutions,
-        is_executable = True,
-    )
-    runfiles = ctx.runfiles(files = [ctx.executable._buildifier])
-    return [DefaultInfo(
-        files = depset([out_file]),
-        runfiles = runfiles,
-        executable = out_file,
-    )]
-
-_buildifier = rule(
-    implementation = _buildifier_impl,
-    attrs = {
-        "verbose": attr.bool(
-            doc = "Print verbose information on standard error",
-        ),
-        "mode": attr.string(
-            default = "fix",
-            doc = "Formatting mode",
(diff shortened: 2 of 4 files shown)


sudoforge:
IMHO, we can remove modes like `check` and `diff` from the non-test rule since users can use the `buildifier_test` rule to perform those actions and validate the state of the workspace.

With that in mind, I deprecated a few attrs and values on the non-test rule, and limited the values for attrs on the test rule. I'm currently notifying the user using `print()` which fails the CI check. What should I do about these? Do you agree that the deprecated values and attrs are a step in the right direction? If so, should I just "yank the bandaid" off? I avoided that and introduced the deprecation so that users wouldn't need to immediately set up their `buildifier_test` rules, but am perfectly fine with removing them in the maintainers agree.

sudoforge:
I'm also interested in exploring integrating with `apple/rules_lint`.

sudoforge:
I'm also exploring what integration with `apple/rules_lint` looks like.

vladmos (approved):
Thanks!

sudoforge:
Hey @vladmos, thanks for merging this in! A couple questions I'd like to get your thoughts on:

* Are the deprecation messages I've added in both sufficient enough, and does it look like a good plan to move the non-formatting modes to the test rule?

* I use the convenience macro I mentioned in my original post, `buildifier_filegroup`, to simplify filegroup creation. Should we add this to the `@com_github_bazelbuild_buildtools` workspace, or at least add instructional documentation in a "tips and tricks" manner?

* I also use an executable rule (see below) for updating the `buildifier_test` target's `srcs` attribute to simplify the maintenance of ensuring the filegroup targets are added to the `buildifier_test` target. My usage is contrived, but I think it would be helpful to add or document this process for other users -- do you think this would be useful?

**`BUILD`** file

```
sh_binary(
    name = "update_check_test_srcs",
    srcs = [":update_check_test_srcs.bash"],
```

**`update_check_test_srcs.bash`**

```
#!/usr/bin/env bash

_run_query() {
  bazel query \
    --output label \
    'attr(name, "^starlark_files$", ...)' 2>/dev/null 
}

_parse_query_results() {
  sed -e 's/\(.*\)/add srcs \1|/' | tr -d '\n'
  printf '//tools/buildifier:check_test\n'
}

if cmdfile=$(mktemp); then
  cd "$BUILD_WORKSPACE_DIRECTORY"

  _run_query | _parse_query_results > "$cmdfile"

  bazel run //tools/buildozer -- -f "$cmdfile"
else
  printf "[SCRIPT ERROR] Unable to create temporary file for generated commands"
fi
```

Kernald:
I think breaking the workspace-checking capacity of the previous `buildifier` rule is a burden - previously, it was a set-and-forget, no per-package declaration was needed. Once set up, it was impossible to make a mistake. Pushing to `buildifier_test` is definitely not equivalent. While it has some clear advantages (not rechecking every single file in the repo every time it runs, for one), it's also easy to make a mistake every time a new package is created. If there was an easy way to set up a repo-wide check to enforce this, it would be fine-ish, but you always end up needing to break out of the sandbox to do this (because of how the package boundaries and `glob` work).

While I like the intention and would ideally like to make such a move, I'm really not sure deprecating `buildifier`'s attributes in favour of `buildifier_test` without a clear path on how to have an enforceable equivalent is a good thing.

sudoforge:
> I think breaking the workspace-checking capacity of the previous `buildifier` rule is a burden - previously, it was a set-and-forget, no per-package declaration was needed. Once set up, it was impossible to make a mistake. Pushing to `buildifier_test` is definitely not equivalent.

Where was this "set-and-forget"? You had to remember to run `bazel run //:buildifier` (or run the binary directly, because often times the build target would become outdated and need to be rebuilt) to format/check your code. Maybe this was added to a VCS hook or otherwise executed before commit in a similar manner; I'd argue that my last comment displays a method to enable the same set-and-forget behavior.

This also addresses the comment about it being easy to make a mistake every time a new package is created -- while this is true (and I'd love to be able to break out of the sandbox for the test rule), it is easy enough to define the `regen_test_srcs` target I left in my previous comment within your workspace, and run that (and additionally perhaps the `buildifier` target) pre-commit in order to achieve the same set-and-forget workflow.

Can you expand on why this doesn't work for you?

Kernald:
My use-case is to have indeed a call to `bazel run //tools:buildifier` to check everything, and the CI job to fail if there's anything wrong here.

The `update_check_test_srcs.bash` script you posted solves part of the issue, but not everything - if I understand it correctly, it will update any target named `starlark_files` properly, but it won't add it (nor report anything) if it's missing entirely. And it still requires parsing pretty much everything anyway.

Unless I'm missing something (I'm not 100% sure my understanding of this script is correct, sorry if it's not), this seems to be doing pretty much the same thing but with extra steps and extra potential for mistakes (unless you add yet another check to be sure every single `BUILD` file in the repo contains a `starlark_files` group), as well as cluttering every single `BUILD` file (I know it's just a load and a call, but it's two statements that were not there before, in every file - I'm just comparing both solutions and I really struggle to see the value in this).

PR #930, opened 2020-12-08
Fix the uninitialized warning

vladmos:
Don't warn if a variable is initialized everywhere but in if-branches that end with `break` or `continue`.

### DIFF
--- a/warn/warn_control_flow_test.go
+++ b/warn/warn_control_flow_test.go
@@ -582,6 +582,21 @@ def foo(x):
 	checkFindings(t, "uninitialized", `
 def foo(x):
   if bar:
+    t = 1
+  else:
+    for y in maybe_empty:  
+      return
+
+  print(t)
+`,
+		[]string{
+			":8: Variable \"t\" may not have been initialized.",
+		},
+		scopeEverywhere)
+
+	checkFindings(t, "uninitialized", `
+def foo(x):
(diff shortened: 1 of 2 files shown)


warn/warn_control_flow_test.go
```diff
@@ -795,4 +810,22 @@ def f():
 			":7: Variable \"x\" may not have been initialized.",
 		},
 		scopeEverywhere)
+
+	checkFindings(t, "uninitialized", `
+def foo(x):
+  for y in x:
+    if foo:
+      break
```
meisterT, line 818: should we split the `break` and `continue` into two separate tests?

vladmos: The test checks that neither `break` nor `continue` break the check and act like `return` inside for-loops, I don't think separate tests are necessary.

If a variable is not defined in a branch but the branch terminates with `break` or `continue`, it shouldn't be taken into account. That's similar to `return` statements, even that `return` terminates completely, from the inside of a for-loop they all mean "terminate the cycle", from outside of a for-loop they don't mean much because it's not guaranteed that the loop is not empty, so the following situations are equivalent for this warning:

    for x in y:
        z = 1
        return
    print(z)

and

    for x in y:
        z = 1
        break
    print(z)

meisterT (approved):

PR #931, opened 2020-12-10
fix(buildifier): ignore exclude_patterns in test mode

sudoforge:


sudoforge:
Oops. Not quite sure how I missed this; I had targets defined for both executable and test modes.

sudoforge:
I can add an e2e test for this, but don't want to hold up this patch for that since `//buildifier` is currently broken (this fixes it). I'll send a separate PR with an e2e test for the rules in `//buildifier` later today if I have time, or some time over the weekend.

sudoforge:
Hey @vladmos, @meisterT, @pmbethe09  -- I don't mean to be a bother, but HEAD is broken without this patch.

vladmos (approved):
Thanks! I'll merge it now as it fixes HEAD, but please also write a test,

sudoforge:
Thanks. Yeah, I'll have some cycles to write an "e2e" test over the next few days but might not get a submission in until early next week. By "e2e" I mean a test that sets up a workspace using both of these rules and invokes them. AFAICT it'll be the first of its kind for buildifier (the integration test that exists today just tests the formatting through the tool itself, not through the rules), so I'll lean on what was done for buildozer to take the same modular approach.

PR #932, opened 2020-12-11
Make fileModified local

vladmos:
Fix bug where "fileModified" is leaking across calls to edit.Buildozer.

This would affect all users of buildozer as a library.

### DIFF
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -1232,8 +1232,8 @@ func Buildozer(opts *Options, args []string) int {
 	}
 	close(data)
 	records := []*apipb.Output_Record{}
-	hasErrors := false
-	fileModified := false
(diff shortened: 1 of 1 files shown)


edit/buildozer.go
```diff
@@ -1236,12 +1233,14 @@ func Buildozer(opts *Options, args []string) int {
 	close(data)
 	records := []*apipb.Output_Record{}
 	hasErrors := false
+	fileModified := false
```
pmbethe09, line 1236: for vars with a default value, 

var fileModified bool

is the internal convention

pmbethe09 (approved):
LGTM

edit/buildozer.go
vladmos: Done

PR #935, opened 2020-12-22
Updated the buildozer usage documentation

ssiano:
The label-list description was changed to indicate that the label-list
is space-separated instead of comma-separated.


vladmos (approved):
Thanks!

Issue #937, opened 2020-12-29
splitOptionsWithSpaces breaks $(location <target>) references


ericastor:
Running the splitOptionsWithSpaces fix can break lines that include a location reference; e.g.,
```
  linkopts = [
    "-Wl,--version-script=$(location script.lds)",
  ],
```
becomes
```
  linkopts = [
    "-Wl,--version-script=$(location",
    "script.lds)",
  ],
```
Any affected build breaks with an unterminated variable reference error, since the template variable is not terminated in the same string.

We either need to avoid breaking lines that include a location reference, or make sure we don't split on these.

ericastor:
This was fixed at the time, but not updated.

PR #938, opened 2020-12-29
Avoid splitting lines with location references

ericastor:
Avoid splitting lines including `"$(location <target>)"` references; this can break variable resolution.

Addresses issue #937.


vladmos (approved):
Thanks

PR #945, opened 2021-01-28
Autogenerate warnings documentation

vladmos:
This commit makes it possible to automatically generate WARNINGS.md based on the data provided in a textproto file. This approach has multiple advantages:

   * It's possible to easily modify the documentation for all warnings at once, e.g. mention how to disable each warning
   * It's possible to extract some information from the code, such as which warnings are disabled by default and which are not supported by buildifier (if a warning is deleted the documentation still stays for some time because existing buildifier binaries may point to it)
   * It's possible to generate the documentation differently for different purposes, e.g. for external and internal builds of buildifier


warn/docs/docs.go
```diff
@@ -0,0 +1,199 @@
+/*
+Copyright 2016 Google LLC
```
meisterT, line 2: nit: 2021

meisterT (approved):

PR #946, opened 2021-01-29
Support type annotations

### LINKED ISSUE #900: Support type annotations
The code:

```
def foo(x: 1):
    pass
```

Currently gives:

```
$ buildifier foo.bzl
foo.bzl:1:11: syntax error near :
```

That code is accepted by `python3`. I think it would be useful to support in `buildifer` in particular because types are an extension that some Starlark implementations support, much the same way that nested-`def` is a Starlark extension that is supported by buildifier.

If the above is accepted, I'd be happy to do a PR to enable it.

vladmos:
The commit allows using Python3-like type annotations for function parameters and return values:

    def f(x: int, y: str) -> List[bool]:
        pass

Fixes #900

### DIFF
--- a/build/BUILD.bazel
+++ b/build/BUILD.bazel
@@ -17,6 +17,7 @@ go_library(
         "rewrite.go",
         "rule.go",
         "syntax.go",
+        "utils.go",
         "walk.go",
     ],
     importpath = "github.com/bazelbuild/buildtools/build",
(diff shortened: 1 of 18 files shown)


meisterT (approved):

sluongng:
Is this syntax compatible with Bazel?

I take it that here Type annotation can be parsed, but is not strongly enforced/checked right?

vladmos:
Correct, Bazel currently doesn't support type annotations, and Buildifier doesn't enforce them in any way, this is just for compatibility with other Starlark dialects that use Python3-like type annotations.

PR #948, opened 2021-01-30
Do not warn about repeated loads in WORKSPACE file

chsigg:
WORKSPACE files allow load() statements that are not at the top, and the imports can shadow previous ones.

This change restricts the 'symbol already loaded' warning to non-WORKSPACE files.

### DIFF
--- a/warn/warn_control_flow.go
+++ b/warn/warn_control_flow.go
@@ -332,10 +332,13 @@ func unusedLoadWarning(f *build.File) []*LinterFinding {
 					load.To = append(load.To[:i], load.To[i+1:]...)
 					load.From = append(load.From[:i], load.From[i+1:]...)
 					i--
+                                	loadFindings = append(loadFindings, makeLinterFinding(to,
+                                        	fmt.Sprintf("Symbol %q has already been loaded on line %d. Please remove it.", to.Name, origin.line)))
+	                                continue
 				}
 
 				loadFindings = append(loadFindings, makeLinterFinding(to,
-					fmt.Sprintf("Symbol %q has already been loaded on line %d. Please remove it.", to.Name, origin.line)))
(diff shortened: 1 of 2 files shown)


laurentlb:
Sure, symbol redefinition is allowed in workspace files, but that doesn't mean it's a good idea. I would personally discourage this pattern.

chsigg:
> Sure, symbol redefinition is allowed in workspace files, but that doesn't mean it's a good idea. I would personally discourage this pattern.

I see your point that redefinition might be considered bad style, and buildifier does not fix it incorrectly but only warns about it.

How about adding a different warning for redefinitions? E.g. "Symbol %q has already been defined on line %d. Please use a different name."

vladmos:
There is a warning for redefined variables: https://github.com/bazelbuild/buildtools/blob/master/warn/warn_control_flow_test.go#L352

Is that what you mean?

Regarding this PR, I agree that the warning makes sense even if shadowing is allowed. In most cases it will confuse the reader if the symbols you need happen to have the same name you can disambiguate by using an alias:

    load("...", "name1")
    load("...", name2 = "name1)

Or, if you need to shadow, suppress the warning with a comment: https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#suppress

chsigg:
I'm suggesting to have two different warnings when load() shadows a previous symbol:

```
load("foo", "foo")
load("foo", "foo")  # warning: symbol 'foo' has already been loaded on line 1. Please remove it.
load("bar", "foo")  # warning: a different symbol 'foo' has already been loaded on line 1. Please use a different local name.
```

Currently, both cases will report the first warning, but it's not very accurate for the second case.

Note: I'm no longer suggesting to handle WORKSPACE files differently. You convinced me that emitting a warning is appropriate.

vladmos (approved):

PR #952, opened 2021-02-02
Disable warnigns about native rules

### LINKED ISSUE #923: Buildifier warns about lack of rules_cc; rules_cc says you don't have to use it
> native-cc: Function "cc_library" is not global anymore and needs to be loaded from "@rules_cc//cc:defs.bzl". (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#native-cc)

> There is no need to use rules from this repository just yet. If you want to use rules_cc anyway, add the following to your WORKSPACE file
https://github.com/bazelbuild/rules_cc#getting-started

These are clearly not compatible. In addition, the instruction to add rules_cc to your WORKSPACE file seem to be not accurate based on https://github.com/bazelbuild/bazel/issues/8743#issuecomment-704209848.

Please make the tools in the ecosystem and their documentation consistent.

Cross-posted as https://github.com/bazelbuild/rules_cc/issues/92

vladmos:
It's not clear whether it will be forced to manually load Starlark rules instead of using native rules for android, cc, java, proto, and python. The corresponding warnings are therefore not enabled by default anymore. It's still possible to use them explicitly in buildifier by specifying the required warning categories with the `--warnings` flag, e.g. `--warnings=native-cc,native-java` or `--warnings=+native-android,+native-py` to use all default warnings and some of non-default.

Fixes #923.

### DIFF
--- a/WARNINGS.md
+++ b/WARNINGS.md
@@ -605,7 +605,7 @@ UPPER_SNAKE_CASE, and providers should be UpperCamelCase ending with `Info`.
 
 The Android build rules should be loaded from Starlark.
 
-Update: the plan for disabling native rules
+Update: the plans for disabling native rules
 [have been postponed](https://groups.google.com/g/bazel-discuss/c/XNvpWcge4AE/m/aJ-aQzszAwAJ),
 at the moment it's not required to load Starlark rules.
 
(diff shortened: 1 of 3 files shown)


meisterT:
cc @comius

meisterT (approved):

PR #959, opened 2021-02-11
Feat: use go modules

Loyalsoldier:


pmbethe09:
So the idea of Go modules is when you make a breaking API change, you increase the major version.
At this time, we have not made any breaking changes that I am aware of, so even though we are version 4.0.0 it isn't really a Go 'v4' implying a 3rd breaking API change.

So I think all we need here is a go.mod and go.sum file to make this work?

PR #969, opened 2021-03-13
When buildifier_test rule is ran with `bazel run`, it will run on the WORKSPACE files rather than on the provided srcs.

### LINKED ISSUE #968: The introduction of buildifier_test removed the option to run check/lint on all BUILD files.
I used to have a `buildifier` rule that ran in a CI, that just warned on format/lint warnings for all files.

With the introduction of buildifier_test (in #929), this now became "deprecated".
It seems like I'm suppose to use `buildifier_test` for checks/lint, but then I have to have all my BUILD (and .bzl) files in the `srcs` fields, which is not an easy task in many projects :(

Even when running the test using `bazel run` (instead of `bazel test`), the `TEST_WORKSPACE` env var is defined, and thus only the runfiles-collected files are checked.

Suggestion for solution:
It seems like the `BUILD_WORKSPACE_DIRECTORY` env var is only available when running with `bazel run` (https://docs.bazel.build/versions/master/user-manual.html#run). We can check `TEST_WORKSPACE && !BUILD_WORKSPACE_DIRECTORY` (pseudo code) to decide if to run on real code or not.

mishas:
This closes #968 .

### DIFF
--- a/buildifier/runner.bash.template
+++ b/buildifier/runner.bash.template
@@ -7,7 +7,7 @@ ARGS=@@ARGS@@
 buildifier_short_path=$(readlink "$BUILDIFIER_SHORT_PATH")
 
 # Use TEST_WORKSPACE to determine if the script is being ran under a test
-if [ ! -z "${TEST_WORKSPACE+x}" ]; then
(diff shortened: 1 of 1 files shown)


mishas:
@meisterT , @pmbethe09 , @vladmos ,
Any chance you can take a look at this please?

Also, CC: @sudoforge

vladmos (approved):

PR #972, opened 2021-04-26
fix(buildifier): default buildifier_test to "diff" mode

### LINKED ISSUE #971: Buildifier_test output irrelevant error message
Hi folks, I have a target setup like this:

```
load("@com_github_bazelbuild_buildtools//buildifier:def.bzl", "buildifier_test")

buildifier_test(
    name = "check_starlark_format",
    size = "small",
    timeout = "short",
    srcs = ["//:starlark_files"],
)

filegroup(
    name = "starlark_files",
    srcs = glob(
        include = [
            "**/*.BUILD",
            "**/*.bzl",
            "**/*.sky",
            "**/BUILD",
            "**/BUILD.*.bazel",
            "**/BUILD.*.oss",
            "**/BUILD.bazel",
            "**/WORKSPACE",
            "**/WORKSPACE.*.bazel",
            "**/WORKSPACE.*.oss",
            "**/WORKSPACE.bazel",
            "*.BUILD",
            "*.bzl",
            "*.sky",
            "BUILD.*.bazel",
            "BUILD.*.oss",
            "WORKSPACE.*.bazel",
            "WORKSPACE.*.oss",
        ],
        exclude = [
            "bazel-*/**",
            ".git/**",
        ],
    ),
)
```

It works all and well, when a file is miss-formatted, the test will fail as expected.

But the error message was not very clear on why the test target failed.

```
==================== Test output for //:check_starlark_format:
buildifier: open ./WORKSPACE: read-only file system
================================================================================
```

Perhaps we should / could improve this error message a bit by telling people that this was mis-formated?

And give an option to output the diff on error?

sudoforge:
This commit changes the default value passed to the buildifier tool via
the `-mode` argument from `fix` to `diff` when defining a
`buildifier_test` target. This is necessary, as `fix` attempts to format
the files, which fails due to attempting to open and operator on files
in a read-only sandbox.

Given that `fix` is not an available value for the `mode` attribute to
the `buildifier_test` rule when manually specified, it makes sense to
remove it.

closes #971

### DIFF
--- a/buildifier/internal/factory.bzl
+++ b/buildifier/internal/factory.bzl
@@ -42,7 +42,7 @@ def buildifier_attr_factory(test_rule = False):
             doc = "Print verbose information on standard error",
         ),
         "mode": attr.string(
-            default = "fix",
+            default = "fix" if not test_rule else "diff",
             doc = "Formatting mode",
             values = ["check", "diff", "print_if_changed"] + ["fix"] if not test_rule else [],
         ),


sudoforge:
It would also be feasible to default to `check`, however, I think the additional context provided by `diff` makes it a much more functional and lucrative default value.

vladmos (approved):

PR #977, opened 2021-05-26
Allow FormatString() to work on CommentBlock

vladmos:
CommentBlock nodes are no longer special cases for the printer and don't need to be handled separately.


vladmos:
The existing tests should cover it. The behavior is not changed, just if you want to use buildifier as a library and call FormatString manually it won't crash on CommentBlocks anymore.

Issue #770, opened 2020-01-27
Running Buildifier in Windows using Bazel


amit-traiana:
I'm trying to run Buildifier in Windows. I'm using Bazel instead of the Go package:

`bazel run buildifier` 

I'm getting the following error:

```
FATAL: ExecuteProgram(C:\users\automation\_bazel_automation\iesztftu\execroot\test\bazel-out\x64_windows-fastbuild\bin\buildifier.bash) failed: ERROR: src/main/native/windows/process.cc(199): CreateProcessW("C:\users\automation\_bazel_automation\iesztftu\execroot\depo\bazel-out\x64_windows-fastbuild\bin\buildifier.bash"): %1 is not a valid Win32 application.
 (error: 193)
```

It looks like a `.bash` file is trying to get executed in a Windows shell?

amit-traiana:
Did more research - I can confirm the problem is with running a `.bash` file in native Windows shell. A way around it is to use `----run_under='bash'`, which starts a Bash shell (I'm using MSYS2 as recommended by Bazel). But that won't work perfectly with the current bash template, because `find` for example is a command that exists in both Windows and MSYS2. The Windows `find` command is earlier on PATH (it's in Windows Folder and you can't control it). So the script fails because it's using the 'Wrong' find parameters (because it's running Windows find). I was able to run the script by stating the specific `find` folder in MSYS2 and fixing other quirks, but it's clear there's a need for a `bat` script for Windows. 

When there is one, it's just a matter of using `select` based on the OS when selecting the template. I might be able to write the `bat` file, just let me know if the above works for you.

limdor:
To me the issue looks related or the same like https://github.com/bazelbuild/buildtools/issues/346

limdor:
@vladmos with the same argument as https://github.com/bazelbuild/buildtools/issues/346#issuecomment-2569162781 I would just close it as fixed

Issue #346, opened 2018-08-02
Buildifier usage via Bazel does not work on Windows


filipesilva:
When following https://github.com/bazelbuild/buildtools/blob/master/buildifier/README.md#setup-and-usage-via-bazel on windows, the following output is observed:

```
$ bazel run //:buildifier
INFO: Analysed target //:buildifier (1 packages loaded).
INFO: Found 1 target...
Target //:buildifier up-to-date:
  C:/users/kamik/_bazel_kamik/gq7cl633/execroot/angular_devkit/bazel-out/x64_windows-fastbuild/bin/buildifier.bash
INFO: Elapsed time: 2.531s, Critical Path: 0.01s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
C:/users/kamik/_bazel_kamik/gq7cl633/execroot/angular_devkit/bazel-out/x64_windows-fastbuild/bin/buildifier.bash: line 8: : command not found
```

(Note: `--direct_run` is not used because it shows a `WARNING: Option 'direct_run' is deprecated` message).

The `buildifier.bash` file is such:
```
#!/usr/bin/env bash

BUILDIFIER_SHORT_PATH='../com_github_bazelbuild_buildtools/buildifier/windows_amd64_stripped/buildifier.exe'
ARGS=('-mode=fix' '-v=false' '-showlog=false')

buildifier_short_path=$(readlink "$BUILDIFIER_SHORT_PATH")
cd "$BUILD_WORKSPACE_DIRECTORY"
"$buildifier_short_path" "${ARGS[@]}" $(find . -type f  \( -name '*.bzl' -or -name 'BUILD.bazel' -or -name 'BUILD' -or -name 'WORKSPACE' \))
```

Debugging this generated file, I find two things:
- `$BUILDIFIER_SHORT_PATH` evaluates to `../com_github_bazelbuild_buildtools/buildifier/windows_amd64_stripped/buildifier.exe`
- `$buildifier_short_path` evaluates to an empty string.

Indeed the path listed by `$BUILDIFIER_SHORT_PATH` does not actually exist. That executable can be found at `../external/com_github_bazelbuild_buildtools/buildifier/windows_amd64_stripped/buildifier.exe` instead though. Changing this variable should help on Windows, but I am unsure of the effect in other platforms.

Another problem is that the path listed is not a symlink on windows, as using the `--verbose` flag on `readlink` highlights:
```
[...]

PR #978, opened 2021-05-30
Clarify buildifier integration support

limdor:
Add a note that the integration of buildifier as a rule is only supported on Linux.
More information https://github.com/bazelbuild/buildtools/issues/346 and https://github.com/bazelbuild/buildtools/issues/770


buildifier/README.md
```diff
@@ -80,7 +80,7 @@ See also the [full list](../WARNINGS.md) or the supported warnings.
 
 ## Setup and usage via Bazel
 
-You can also invoke buildifier via the Bazel rule.
+On Linux you can also invoke buildifier via the Bazel rule.
```
keith, line 83: This rule also supports macOS so can you add that here I guess? Or just add `(not supported on Windows)` instead?

limdor: @keith fixed

limdor:
@keith any chance that we can get this merged? Is something missing?

vladmos (approved):
Thanks, sorry for the late review.

PR #980, opened 2021-06-09
Add !package keyword for new rule targeting

wcn3:
This new keyword helps target placement of new rules near the package()
declaration in a file. This is useful for adding new load() and
license() statements since it keeps them near the top since they are
globally significant rules.


aiuto:
This sounds useful, but I can't figure out how to use it from the test case. Can you add a sample usage to the PR description?

vladmos (approved):
Thanks

laurentlb:
@vladmos Can you confirm this PR is really needed?

vladmos:
I don't know, thought it would not hurt to merge it but can roll it back if you think I should.

laurentlb:
My understanding from the description is that it tries to reimplement the same thing as the `__pkg__` target, but in a more brittle way.

wcn3:
The __pkg__ keyword does not work as a relative rule for before/after placement. Perhaps it would be better to rework the PR to use '__pkg__' instead of '!package' to specify the location, but something is needed to achieve this effect.

laurentlb:
Would it make sense to reuse the term `__pkg__` here? I think that would be consistent and intuitive.

wcn3:
I think that makes more sense than inventing the new syntax I originally proposed. I can't add more to this PR, so I created https://github.com/bazelbuild/buildtools/pull/996 to capture this work.

PR #985, opened 2021-06-23
WIP : Support Tabular Data formatting in Buildifier

SKashyap:
## What?
This change adds to the Buildifier the ability to format tabular data using a set of new tags as follows : 

Format tabular data in proper visual alignment
	`#buildifier: format table`
Sort tabular data (by column) alongside the allignment
	`#buildifier: format table sort <column-num>`

## Why?

The need for the change can be seen with the following example : 
Use an input file : 
```
tableEx = [
    ("col1",      "col2",     "col3"),
    ("col1111", "col222", "col3333")    
]
```

Run it through buildifier to see the following output : 
```
tableEx = [
    ("col1","col2","col3"),
    ("col1111","col222","col3333")    
]
```
#### Detailing the problem statement : 
- Tabular Data can be either "List of Tuples"/ "List of Lists". (See ListExpr , TupleExpr)
- The buildifier's parser stage freezes how an `Expr` will be printed using flags file `ForceCompaction` for Tuples , and `ForceMultiline` for Lists. 
- These fixed notions of handling the print logic make it difficult to operate with tabular data as shown in the above example. 
- Even if the user formats the table himself, the buildfier cannot be indicated "#buildifer : Don't mess with my table". It ruins the user-provided format too. #leaveAlone does not work either. 
- We came across this problem when we were building large RPMs with 200 endstates, each having a location of install/permissions, etc as table data. 

#### Design Description : 
- Add support for 2 new tags to be parsed with `rewrite.go` :
`#buildifier: format table`
`#buildifier: format table sort <column-num>`
- Existing architecture of buildifier allows rewriting the AST within `rewrite.go` based on parsed buildifier tags.
- Utilize the same pipeline to "mark" the nodes to form the TableRoot and the TableRows based on the presence of the new tags. 
- Within `print.go`, flex the logic to allow printing a tuple /List in a tabular form when they are identified a TableRoot/TableRow node. 

#### Printing of Tabular data :
- We choose an existing golang package called `TabWriter` to jump-start us into this feature without writing custom logic to format rows/column sizes out the output. 
- `TabWriter` implements "Elastic Tabstops algorithm "

#### Other approaches evaluated : 
- We also evaluated an approach of creating a new Expr class called `TableListExpr`, `TableRowExpr` - It did not provide any advantage over the existing design given the current problem. 
  We can revisit this later if the use cases demand it. 

#### Sorting of Tabular data : 
- `Rewrite.go` logic where we detect the new tags also sort the row entries based on the index provided in the tag.

#### Future implementations: 
- This changeset only supports a "List of Tuples" as a table. A similar extension can be made for "List of List" as well. 

## Testing?
- 15 testcases added to support the use-cases 
- bazel test ... : PASSED. 

## Any other info?
There are a couple of known drawbacks due to the `TabWriter` behavior : 
- Especially in the case of "Before Comments" in tableRows. We will be curating them into a document for reviewers soon.


rtabassum:
@googlebot I fixed it.

SKashyap:
@googlebot I fixed it.

SKashyap:
@googlebot I fixed it.

SKashyap:
@googlebot I fixed it.

Gormo:
@SKashyap looks like a really nice improvement! What's the status of this?

systemlogic:
@SKashyap have you given up on PR ?

SKashyap:
> @SKashyap have you given up on PR ?

I had stopped working on this for a while. But have made some recent changes to this in our internal repository. Will collate those and wrap this asap.

PR #990, opened 2021-08-02
Add safebuildifier packge.

eljobe:
This allows client code that wants to be sure not to spawn
subprocesses to import the safebuilder package which doesn't call
os.exec.


edit/BUILD.bazel
```diff
@@ -34,3 +35,14 @@ go_test(
     embed = [":go_default_library"],
     deps = ["//build:go_default_library"],
 )
+
+go_library(
+    name = "safebuildifier",
```
vladmos, line 40: Having multiple go-libraries works well with Bazel, however not compatible if one uses standard go tooling (e.g. `go build`). Could you please move `safebuildifier.go` and this `go_library` definition to another directory (can be a subdirectory, e.g. `edit/safe`?

edit/buildifier.go
```diff
@@ -0,0 +1,44 @@
+package edit
```
vladmos, line 1: I suggest renaming the file to `default_buildifier.go`, otherwise it's ambiguous, people may think it's the canonical file for the buildifier binary.

edit/BUILD.bazel
eljobe: Done.

edit/buildifier.go
eljobe: Done.

vladmos:
Please format the `BUILD.bazel` files (likely it's just about the order of arguments): `buildifier edit/BUILD.bazel edit/safe/BUILD.bazel`

eljobe:
Okay. Done. I think I must have been running an older version of buildifier before.

vladmos (approved):

Issue #357, opened 2018-08-19
Buildozer ignores command line command/label-list when -f is provided


endobson:
Buildozer documents its usage as
```
buildozer [OPTIONS] ['command args' | -f FILE ] label-list
```

Thus `command args` and `-f FILE` are supposed to be mutually exclusive, but this isn't enforced. In addition `label-list` is ignored if `-f FILE` is provided.

I believe the intended usage is
```
buildozer [OPTIONS] ['command args' label-list | -f FILE ]
```

Example command which does nothing right now, but should likely do something different (either error or apply the command).
```
buildozer -f /dev/null 'add dep //foo:bar' //libraries:io
```

aiuto:
+100  The current behavior is almost entirely useless. 
You should be able to have a list of commands in the command file - with NO target label, and apply them
to all the labels specified on the command line.

aiuto:
#991 solves the problem for me.  It needs tests and a doc fix, but I am using it for now to complete a large change.

PR #991, opened 2021-08-02
Allow mixing command files with label lists

aiuto:
Fixes for #357.

If a line in the command file includes a '*' as the label, that will
expand to the full list of labels from the command line.

For example with the file `cmd.dozer`
```
remove compatible_with //buildenv/target:foo|*
remove compatible_with //buildenv/target:bar|//other_pkg:*
```
then this invocation
```
buildozer -f cmd.dozer //my_pkg/foo/... //biz/bang:__pkg__
```
would be equivalent to
```
'remove compatible_with //buildenv/target:foo' //my_pkg/foo/... //biz/bang:__pkg__
'remove compatible_with //buildenv/target:bar' //other_pkg:*
```

### DIFF
--- a/buildozer/README.md
+++ b/buildozer/README.md
@@ -19,9 +19,12 @@ buildozer [OPTIONS] ['command args' | -f FILE ] label-list
 ```
 
 Here, `label-list` is a space-separated list of Bazel labels, for example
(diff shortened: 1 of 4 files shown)


vladmos (approved):
Thanks

PR #992, opened 2021-08-04
Bug fix: add size check on split of "cmd|*" syntax.

aiuto:
Fixes an index error when detecting the buildozer command file syntax `command|*`.
This can occur when a command file is used with no label-list.


vladmos (approved):

edit/buildozer.go
```diff
@@ -1145,7 +1145,7 @@ func appendCommandsFromReader(opts *Options, reader io.Reader, commandsByFile ma
 			continue
 		}
 		args := strings.Split(line, "|")
-		if args[1] == "*" {
```
laurentlb, line 1148: What does the error message look like in this case?

PR #993, opened 2021-08-19
Fall back to x86_64 binary when run tools via rules_nodejs on Apple silicon

thii:


thii:
cc @alexeagle

keith:
Can we just build a arm64 binary instead? we should be unblocked to do that at this point right?

thii:
Does rules_go support that yet?

thii:
Looks like the latest of it does.

thii:
https://github.com/bazelbuild/buildtools/pull/994

PR #994, opened 2021-08-19
Add release binary for darwin-arm64 architecture

thii:
And update rules_go since it was needed to build darwin-arm64 binary.


keith (approved):
nice

vladmos (approved):
Thanks!

PR #995, opened 2021-08-20
Add darwin_arm64 to supported archs

thii:
Also downgrade Node.js version to 16.6.2 since it's the latest supported
version in rules_nodejs 0.38.0
(https://github.com/bazelbuild/rules_nodejs/blob/3.8.0/internal/node/node_versions.bzl#L1415)
to fix the npm integration tests and enable them on CI.


WORKSPACE
```diff
@@ -79,7 +79,7 @@ http_archive(
 
 load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories", "npm_install")
 
-node_repositories(node_version = "16.7.0")
+node_repositories(node_version = "16.6.2")
```
vladmos, line 82: Why is the version lower than before?

thii: I took the latest supported version from the master branch of rules_nodejs, but later realized that we are using rules_nodejs 0.38.0, in which node 16.6.2 is the latest supported one.

vladmos: Have you tried updaing rules_nodejs instead? If that can be done without much hassle I'd do that, older versions tend to become not supported/compatible with the latest Bazel.

PR #996, opened 2021-08-30
Update package placement keyword and documentation

wcn3:
Update the package placement keyword to be `__pkg__` to be consistent
with other usages in Bazel. Expanded the documentation to note this
use case.


wcn3:
This is followup work from the discussion in https://github.com/bazelbuild/buildtools/pull/980

edit/edit.go
```diff
@@ -265,7 +265,7 @@ func IndexOfRuleByName(f *build.File, name string) (int, *build.Rule) {
 
 		// Allow for precisely targeting the package declaration. This
 		// helps adding new load() and license() rules
-		if name == "!package" {
+		if name == "__pkg__" {
 			if rule, ok := ExprToRule(stmt, "package"); ok {
```
laurentlb, line 269: What happens if there's no call to `package` in the file?
Could you call `PackageDeclaration()` instead?

edit/edit_test.go
```diff
@@ -775,7 +775,7 @@ func TestIndexOfRuleByName(t *testing.T) {
 	}{
 		{"first", 3},
 		{"%8", 4},
-		{"!package", 2},
```
laurentlb, line 778: An integration test would be nice in https://github.com/bazelbuild/buildtools/blob/master/buildozer/buildozer_test.sh#L1170

laurentlb (commented):
Thanks!

edit/edit.go
wcn3: Calling PackageDeclaration() would create the rule if it doesn't exist, so that wouldn't be the function to use for targeted placement like what is being done here. It and its counterpart ExistingPackageDeclaration are working on the build.File level to return a rule, but I'm trying to get a line position. This ExprToRule usage is exactly what ExistingPackageDeclaration is doing under the hood.

If there's no call to package in the file, the insertions work exactly like they would if the package() call existed in the preferred location as described by the documentation (after loads, variables, etc. but before rules)

edit/edit_test.go
wcn3: I didn't see that test previously. That explains why there wasn't a unit test for this. Should I go ahead and revert this unit test since it's not adding any real value above the integration test?

laurentlb: Yes, sounds good.

wcn3: PTAL

laurentlb (approved):
thanks!

PR #1000, opened 2021-09-08
Make buildifier sort named arguments of all calls.

fourdman:
Currently buildifer sorts named arguments only when the called function is a literal. This change allows the called function to be any expression, e.g. `foo.bar(a=1,b=2)` or `foo().bar(a=1,b=2)`.

### DIFF
--- a/build/rewrite.go
+++ b/build/rewrite.go
@@ -267,13 +267,10 @@ func fixLabels(f *File) {
 }
 
 // callName returns the name of the rule being called by call.
-// If the call is not to a literal rule name, callName returns "".
+// If the call is not to a literal rule name or a dot expression, callName
+// returns "".
 func callName(call *CallExpr) string {
-	rule, ok := call.X.(*Ident)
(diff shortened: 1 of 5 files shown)


build/rewrite.go
```diff
@@ -288,7 +288,7 @@ func sortCallArgs(f *File) {
 		}
 		rule := callName(call)
 		if rule == "" {
-			return
+			rule = "<complex rule kind>"
```
vladmos, line 291: I'd also extract the names from DotExpr nodes (such as `foo.bar(...)`). This way we can have custom sorting rules for them in `tables.NamePriority` too.

fourdman: Done (updated callName()).

vladmos (approved):
Thanks. I'll be on vacation next week, will merge and import the change after that.

laurentlb:
fyi - this change should be mentioned in the release notes.

PR #1008, opened 2021-09-18
feat(buildozer): add 'path' attr to print command

mattem:
Adds a `path` attribute to the existing `print` command, allow printing of the absolute path the BUILD file that contains the current rule. 

This can be useful when using buildozer's `print` feature to get a lot of data programmatically from BUILD files and using a glob pattern (eg `buildozer 'print srcs deps' //foo:%some_rule`) and it would be useful to get the corresponding file path back to the BUILD file without parsing and calculating it from the label, and taking into account `BUILD.bazel` vs `BUILD` etc.

### DIFF
--- a/buildozer/README.md
+++ b/buildozer/README.md
@@ -216,6 +216,7 @@ There are some special attributes in the `print` command:
   * `rule`: the entire rule definition
   * `startline`: the line number on which the rule begins in the BUILD file
   * `endline`: the line number on which the rule ends in the BUILD file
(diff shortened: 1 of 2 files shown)


alexeagle:
hey @vladmos have a minute for this one?

vladmos:
LGTM, I'll merge it a bit later, hopefully this week. There's an ongoing incompatible change that hasn't been merged internally yet, I'd like to isolate it from other changes.

vladmos (approved):

PR #1010, opened 2021-09-27
Add mailmap to correct author names and email addresses

sudoforge:
```
This commit adds a .mailmap at the root of the repository in order to
map author names and email addresses to a canonical format (e.g.
combining duplicates, removing non-name strings). When different
email identities existed for the same author, the information from the
most recent commit made by the author was selected.

This plays nicely with `git-shortlog`, and `git-log` with the
`--use-mailmap` option (alternatively, specify `log.mailmap = True` in
your git configuration) [0].

[0]: https://git-scm.com/docs/git-config#Documentation/git-config.txt-logmailmap
```

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,20 @@
+<adamjernst@me.com> <adamjernst@fb.com>
+<alexeagle@angular.io> <eagle@post.harvard.edu>
+<bucjac@gmail.com> <buchgr@google.com>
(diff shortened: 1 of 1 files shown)


sudoforge:
I should note that all mappings other than my own are done on a best-guess basis, e.g. mapping commits with just a first name specified to the full name (eg in @kchodorow's case).

vladmos (approved):
Thanks!

Issue #801, opened 2020-03-06
[buildifier] Add ability to ignore paths when searching recursively for files


joeljeske:
The `-r` option is very nice and helpful. Thanks!

In my workspace, I have a bazel managed directory `node_modules` that contains some .bzl files. (This directory contains 3rd party artifacts from the `npm` registry managed by rules_nodejs.) The issue is that when buildifier is run recursively from the root, it finds files in this directory and changes them. This results in cache misses due to the files being changed. 

Possible Solutions: 

1. Add an `-exclude` flag that can be passed for each path to ignore
2. Add a `.buildifierignore` file that can be checked in to the repo to ignore paths when finding files

Workarounds: 

1. Manually specify all paths and directories to search through *except* my managed directory 
2. Go back to using `find`, which negates the value of `-r` and makes it not portable again. 

cc @alexeagle

rockwotj:
Maybe just respecting `.bazelignore` would work too?

drewmacrae:
I hit a similar issue and found when I added buildifier to my project's WORKSPACE and BUILD file I could specify something like:
`buildifier(
    name = "buildifier",
    exclude_patterns = ["node_modules"],
)`
In the BUILD file. Like they do [here](https://github.com/bazelruby/rules_ruby/blob/c79462151d284620679a05ea7b7c0c1839383277/BUILD.bazel).

This is like workaround 2, but at least it's automated and managed as a part of buildifier in buildtools.

six8:
https://github.com/bazelbuild/buildtools/issues/623 requests support for respecting bazelignore.

oreflow:
+1 on that this should be supported through bazelignore, closing this as duplicate of #623

PR #1011, opened 2021-10-01
Add windows support

PatriceVignola:
This PR adds Windows support to call `bazel run //:buildifier` without needing msys2 workarounds by adding a native `.bat` file. This should address issues that have been opened a while ago: https://github.com/bazelbuild/buildtools/issues/770 and https://github.com/bazelbuild/buildtools/issues/346


buildifier/runner.bat.template
```diff
@@ -0,0 +1,45 @@
...
+set stripped_args=%ARGS:~1,-1%
+
+rem Unquote the arguments
+set stripped_args=%stripped_args:'=%
+
+rem Get the absolute path to the buildifier executable
+for /f "tokens=2" %%i in ('findstr /r "\<buildifier\.exe\>" MANIFEST') do (set buildifier_abs_path=%%i)
+
+powershell ^
+function Buildify($Root)^
```
vladmos, line 15: Buildifier has the `-r` flag to recursively find all relevant files within a directory, maybe it makes sense to reuse it? The built-in recursive function can ignore some paths (currently just `.git`, but may support `.buildifierignore` in the future, see #801), I think it's better to have such logic defined once.

PatriceVignola: But the Linux version does it recursively by default, without needing the `-r` flag. I think the Windows and Linux versions should have the same behavior.

vladmos: Maybe it had been implemented this way before the `-r` flag was introduced. I can fix it later, but won't be able to fix the windows version easily because don't have a windows machine around. So I suggest using `-r` from the beginng.

PatriceVignola: How do I use the `-r` flag with `bazel run //:buildifier`? I'm able to use it with the standalone executable but not as a bazel rule.

craigmparis:
This would be very useful for me. Is there any progress on this?

buildifier/runner.bat.template
cha5on: I noted on https://github.com/bazelbuild/buildtools/issues/1136 that the `-r` flag also does not match the same set of files.  The set used by the runner template is more complete.  Ideally, those should probably be aligned, but I'd be in favor of merging this in for running on Windows first.

limdor:
What is the status of this PR? Any news?

vladmos:
I agree for merging it now and fixing the behavior of recursive flags later, sorry for the delayed review.

vladmos (approved):

limdor:
Cool thanks. Any estimations for when is planned the next release @vladmos?

limdor:
I'm not sure if I'm the only one but for me it does not seem to work the autodetection of being on windows
https://ci.appveyor.com/project/limdor/bazel-examples/builds/49536214/job/q3jx26k4anwxip9u
I need to investigate a bit more what the problem might be.

limdor:
@PatriceVignola @vladmos I'm not sure how this was validated but to me it does not work unless I remove the declaration of the bash file in Windows
https://github.com/bazelbuild/buildtools/pull/1262

vladmos:
@limdor Sorry, the previous internal import with incompatible formatting rules took longer than expected. I'll cut a new release next Monday.

Also I've just merged #1262, could you please validate the issue is resolved for you? I don't have a Windows machine at the moment.

limdor:
With #1262 it works for me.

PR #1012, opened 2021-10-06
Deny sorting for cc_embed_data.srcs

### LINKED ISSUE #1004: cc_embed_data.srcs should be listed as no-sort
The order of elements in the sources impacts the order of embedded data, so we should respect the user order.

vladmos:
Fixes #1004


laurentlb:
I've left my comments on #1004

PR #1028, opened 2022-01-18
Use repo-relative labels instead of hardcoded repo name

fmeum:
By using repo-relative labels, this repository can be loaded under a different repository name, e.g. just "buildtools". This will also make it possible to publish it to the Bazel Central Registry as a module named "buildtools".


laurentlb (approved):
lgtm

PR #1032, opened 2022-01-25
Explicitly describe relative rules in README

algorowara:
Make the meaning of "relative rule" explicit.

Figuring this out with trial and error took me more time and effort than I'm proud of, and having had the option to Ctrl+F the documentation for the word "relative" would have saved me a lot of time. I imagine it will also be useful for others.

Please contact me via algorowara@google.com with any questions. I have not signed the Google Individual Contributor License Agreement because I am a current Google employee, and I believe that would be redundant and/or legally confusing, as Google owns the copyright to my changes already.

### DIFF
--- a/buildozer/README.md
+++ b/buildozer/README.md
@@ -45,6 +45,10 @@ macros.
    `//pkg:%123`.
   * Use `-` for the package name if you want to process standard input stream
    instead of a file: `-:all_tests`.
+  * Some commands take a _relative rule_, which is simply a rule without the
+    package. The absolute rule `//path/to/pkg1:rule1` would become just `rule1`.
+    Relative rules using the special macros listed above, such as `__pkg__` or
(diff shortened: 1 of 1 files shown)


buildozer/README.md
```diff
@@ -45,6 +45,10 @@ macros.
    `//pkg:%123`.
   * Use `-` for the package name if you want to process standard input stream
    instead of a file: `-:all_tests`.
+  * Some commands take a _relative rule_, which is simply a rule without the
+    package. The absolute rule `//path/to/pkg1:rule1` would become just `rule1`.
+    Relative rules using the special macros listed above, such as `__pkg__` or
+    `%123`, are also valid.
```
laurentlb, line 51: Which commands? I think the command argument could be clarified below.

I'd like to avoid adding this block of text here, because the section specifically refers to the command-line target.

PR #1037, opened 2022-02-01
Add support for options specified in `//:.bazelrc.user`

sudoforge:
```
Add support for options specified in `//:.bazelrc.user`

With this change, users can specify options for Bazel in a configuration
file at `//:.bazelrc.user`. The options specified in this file will
be merged with options specified in `//:.bazelrc`. Because this
`//:.bazelrc.user` file is loaded at the end of `//:.bazelrc`, options
specified in the former will override those set in the latter.
```

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,1 +1,10 @@
 build:release -c opt --stamp --workspace_status_command="$PWD/status.py"
+
+################################################################################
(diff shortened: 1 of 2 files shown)


vladmos (approved):
Thanks

PR #1042, opened 2022-02-08
Fix 'unitialized' warnings for nested functions and comprehensions

### LINKED ISSUE #926: Buildifier confused about return values in nested methods
eg:
```
def outer():
  def inner():
    return 0

  i = inner()
```

Buildifier throws the following error on that code snippet:
return-value: Some but not all execution paths of "outer" return a value

But what executions of outer() have a return value? I would suspect that that's safe/valid starlark, but buildifier thinks otherwise?

vladmos:
Fixes #926

### DIFF
--- a/warn/warn_control_flow.go
+++ b/warn/warn_control_flow.go
@@ -421,6 +421,7 @@ func unusedVariableCheck(f *build.File, root build.Expr) (map[string]bool, []*Li
 		case *build.LoadStmt:
 			// LoadStmt nodes store the loaded symbols as idents, even though in the
 			// source code they are strings. These idents may confuse the check,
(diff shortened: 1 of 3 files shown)


warn/warn_control_flow.go
```diff
@@ -421,6 +421,7 @@ func unusedVariableCheck(f *build.File, root build.Expr) (map[string]bool, []*Li
 		case *build.LoadStmt:
 			// LoadStmt nodes store the loaded symbols as idents, even though in the
 			// source code they are strings. These idents may confuse the check,
+			// source code they are strings. These idents may confuse the check,
```
meisterT, line 424: accidental copy?

meisterT (approved):

PR #1045, opened 2022-02-10
Fix typo in README: dict_delete should be dict_remove.

nathdobson:
"dict_delete" never existed, so this should just say "dict_remove". I guess no one has fixed this yet because it's so rarely needed.


vladmos (approved):
Thanks!

PR #1047, opened 2022-02-10
Suppress the unused variable warnings on underscored variables in speсial cases

### LINKED ISSUE #1044: Buildifier: Variables prefixed with `_` should be ignored by `unused-variable` checks
I and seemingly other contributors to `rules_rust` (and surely other rules) have had the expectation that variables prefixed with `_` were unused. Needing to additionally annotate variables with a comment feels unnecessary and can increase the burden of reviewers where a user may call some native or external macro which returns multiple values but without a `_` prefix it's unclear which is actually unused.

```python
# @unused
compilation_ctx, cc_compilation_outputs = cc_common.compile(...)
```
vs
```python
compilation_ctx, _cc_compilation_outputs = cc_common.compile(...)
```

The latter makes it clear that a specific result of this external function is not used and allows the warning to catch cases where the second variable becomes unused, this informing users the entire snippet can be removed.

vladmos:
Fixes #1044 

Buildifier won't warn about unused variable that prefixed with `_` and:

  * it's a function argument
  * it's defined in an assign statement toghether with a non-underscored variable:  `x, _y = z`
  * It's defined as a for-loop iterator: `for _x, _y in iterable:`


UebelAndre (approved):
This seems perfect! Thank you so much!

keith:
It would be great to merge this and cut a new release as many projects using bazel CI are broken and this is a nice workaround for some cases

warn/warn_control_flow_test.go
```diff
@@ -592,6 +592,34 @@ foo()
...
+  a, b, _c = 1, 2, 3  # ok to not use _c
+  print(a)
+
+  _d, _e = 4, 5  # all are underscored
+  print(_d)
+
+  for f, g, _h, _ in my_iterable:  # ok to not use any underscored
+    print(f)
+
+  for _i, (_j, _k) in another_iterable:  # ok to not use any of them
```
meisterT, line 608: where is `another_iterable` defined?

vladmos: These code snippets are not being executed, so it doesn't matter, buildifier only cares that it's a syntactically valid code that can be parsed.

It could be a global variable, or a variable from some prelude file, or an error. If a variable is not defined, users are likely to detect it anyway when they execute the code, buildifier doesn't check for completely undefined variables (although it does for partially undefined, e.g. if it's defined inside `if` but not inside `else`). Even if it did check for such cases, it would be a separate warning category, out of the scope of this test case anyway.

meisterT (approved):

PR #1048, opened 2022-02-14
Fix potential NPE

vladmos:
Custom linter fixes may introduce nil function parameters. That was not handeled correctly by other linter checks and the AST printer and could make buildifier crash with NPE.


meisterT (commented):
Is there an easy way to add a test for it?

vladmos:
It's not trivial, the NPR only occurs if a linter check modifies the AST by adding nil-arguments to a function call which is only the case for some internal checks. I can try adding an integration test reproducing it when I import the change.

PR #1049, opened 2022-02-16
Buildifer 2 support byte string and load comments

emcfarlane:
Adds support in buildifer2 for comments on `load` statements and support for byte strings.

Bytes was added to starlark-go here: https://github.com/google/starlark-go/pull/330


emcfarlane:
What's the correct way to get the hash of a golang module dependency?

Issue #916, opened 2020-10-14
function-docstring-args regex is too strict


binoche9:
If I have the following:

```
def foo(files):
    """
    Foo function

    Args:
        files (List<Files>): List of files.
    """
```

Then buildifier complains because its [regex](https://github.com/bazelbuild/buildtools/blob/e6efbf6df90bec363c3cbd564b72be6c8a309f14/warn/warn_docstring.go#L106) is too strict. The regex to match within the parenthesis is `[\w\ ,]+`, which doesn't match the angled brackets that I'd like to use. I think it'd be good if the regex was loosened to match some more special characters.

PR #1056, opened 2022-03-21
Allow brackets to describe function arguments in docstrings

binoche9:
Allow the usage of brackets (both square and angled) inside the parentheses used to describe function arguments in docstrings. This allows concise descriptions like

```py
def foo(ctx, files):
    """
    Args:
        ctx (Context): the Starlark context
        files (List<File>): a list of files used to do (...)
        mapping (Map[string, int]): a mapping for the (...)
    """
    pass
```

I opened https://github.com/bazelbuild/buildtools/issues/916 but no one responded there so I'm just going to open this PR to address that issue and hope I have better luck here

Fixes: https://github.com/bazelbuild/buildtools/issues/916

### DIFF
--- a/warn/warn_docstring.go
+++ b/warn/warn_docstring.go
@@ -307,7 +307,7 @@ func functionDocstringArgsWarning(f *build.File) []*LinterFinding {
 			if name == "" {
 				continue
 			}
-			name = op + name // *args or **kwargs
+			name = op + name  // *args or **kwargs
 			paramNames[name] = true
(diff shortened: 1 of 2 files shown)


vladmos (approved):
Thanks! Could you please add "Fixes: #916" to the PR description so that the issue is automatically closed on merge?

binoche9:
Done! Can you just double check that I did it right before I merge?

vladmos:
It looks correct

PR #1061, opened 2022-04-12
Fix symbols considered unused when referenced as types

chermehdi:
This fix is trying to address the issue of symbols considered unused when imported and used as types only

Example:

```
load(
  ":f.bzl",
  "s1",
  "s2",
)
def test(x: s1) -> List[s2]:
  pass
```

The PR removes the need for using the `# @unused` directive.


vladmos (approved):
Thanks!

PR #1066, opened 2022-05-24
Add a native.sh_binary wrapping around the buildifier rule

luna-duclos:
This adds windows support in an incredibly minimal and non intrusive way with zero additional maintenance burden.


vladmos:
Hi, sorry, the PR seems to have been lost, are you still intending to submit it? Could you please clarify how it adds windows compatibility?

PR #1074, opened 2022-07-13
Adding rewriter functionality to rewrite to allow rewrite customization

rgwgoogle:
Functionality of current rewrite is preserved by recreating original rewrite parameters from the table currently used. The new functionality will utilize the rewrite method with a receiver.


build/print.go
```diff
@@ -130,10 +136,10 @@ func (p *printer) newline() {
 // If softNewline is called several times, just one newline is printed.
 // Usecase: if there are several nested blocks ending at the same time, for instance
 //
-//     if True:
-//         for a in b:
-//             pass
-//     foo()
+//	if True:
```
rgwgoogle, line 139: psa - My auto linter does these edits

build/rewrite.go
```diff
@@ -54,12 +54,59 @@ func allowedSort(name string) bool {
...
+			if !disabled(r.name) {
+				if f.Type&r.scope != 0 {
+					r.fn(w, f)
+				}
+			}
+		}
+	} else {
+		for _, r := range rewrites {
+			if !disabled(r.name) {
+				for _, n := range w.Rewrites.Names {
```
vladmos, line 105: Do you need the outer for-loop here? Why not just iterate `w.Rewrites.Names`?

build/rewrite.go
```diff
@@ -54,12 +54,59 @@ func allowedSort(name string) bool {
...
+		StripLabelLeadingSlashes:        tables.StripLabelLeadingSlashes,
+		ShortenAbsoluteLabelsToRelative: tables.ShortenAbsoluteLabelsToRelative,
+	}
+	rewriter.Rewrite(f)
+}
+
+func (w *Rewriter) Rewrite(f *File) {
+	if w.Rewrites == nil {
+		for _, r := range rewrites {
+			if !disabled(r.name) {
```
vladmos, line 96: Please reduce nesting:

    if disabled(r.name) {
        continue
    }
    ...

build/rewrite.go
vladmos: Also you have two chunks of code that are almost duplicates, maybe it makes sense to create a local variable that's either equal to `w.Rewrites.Names` or `rewrites` and iterate it just once?

WORKSPACE
```diff
@@ -37,6 +37,20 @@ go_repository(
     version = "v0.0.0-20200203144150-6677ee5c7211",
 )
 
+go_repository(
+    name = "testifiy_assert",
```
vladmos, line 41: This is a new third-party dependency that makes it harder to import the code to the internal repository, is it possible to avoid using it here? Just use manual checks with `t.Errorf()` with explicit error messages.

build/rewrite.go
```diff
@@ -146,8 +193,7 @@ func keepSorted(x Expr) bool {
 // Second, it removes redundant target qualifiers, turning labels like
 // "//third_party/m4:m4" into "//third_party/m4" as well as ones like
 // "@foo//:foo" into "@foo".
-//
-func fixLabels(f *File) {
+func fixLabels(w *Rewriter, f *File) {
```
vladmos, line 196: The *Rewriter object is not used for almost all the functions, please move it to the last place of the arguments list and rename to `_` for the functions where it's not used.

build/rewrite.go
```diff
@@ -54,12 +54,59 @@ func allowedSort(name string) bool {
...
+type Rewriter struct {
+	Rewrites                        *RewriteSet
+	LabelArg                        map[string]bool
+	Denylist                        map[string]bool
+	SortableListArg                 map[string]bool
+	SortDenyList                    map[string]bool
+	SortAllowList                   map[string]bool
+	NamePriority                    map[string]int
+	StripLabelLeadingSlashes        bool
+	ShortenAbsoluteLabelsToRelative bool
```
vladmos, line 76: I only see the usage of `Rewrites` and `NamePriority` in the code, do the other attributes affect anything?

WORKSPACE
rgwgoogle: Removed assert

build/rewrite.go
rgwgoogle: Keeping it as the intended use is to use rewriter which is passed in, by default it will contain table json properties

vladmos: How does it prevent moving it to the last place and renaming to `_` in the functions that don't use it?

build/rewrite.go
rgwgoogle: Made code smaller and more legible

build/rewrite.go
```diff
@@ -32,7 +32,10 @@ import (
 var DisableRewrites []string
 
 // disabled reports whether the named rewrite is disabled.
-func disabled(name string) bool {
+func disabled(w *Rewriter, name string) bool {
+	if w.DisableRewrites != nil {
+		DisableRewrites = *w.DisableRewrites
```
gtong, line 37: This shouldn't overwrite the package variable. Should do something like:
```
var disableRewrites := DisableRewrites
if w.DisableRewrites != nil {
  disableRewrites = w.DisableRewrites
}
for _, x := range disableRewrites {
```

build/rewrite.go
```diff
@@ -54,17 +57,73 @@ func allowedSort(name string) bool {
...
-				r.fn(f)
-			}
+		if !disabled(w, r.name) && (f.Type&r.scope != 0) || (w.DisableRewrites != nil) {
+			r.fn(f, w)
 		}
 	}
 }
 
+func initRewriter(w *Rewriter) {
+	if w.LabelArg == nil {
```
gtong, line 101: Doing this has the side effect of changing the passed-in Rewriter object, which would be an unexpected side-effect. I would suggest passing along the nil values as they are now, and creating local variables where the values are used.

build/rewrite.go
```diff
@@ -32,7 +32,10 @@ import (
 var DisableRewrites []string
 
 // disabled reports whether the named rewrite is disabled.
-func disabled(name string) bool {
+func disabled(w *Rewriter, name string) bool {
+	if w.DisableRewrites != nil {
```
gtong, line 36: This shouldn't overwrite the package variable. Should do something like:
```
var disableRewrites := DisableRewrites
if w.DisableRewrites != nil {
  disableRewrites = w.DisableRewrites
}
for _, x := range disableRewrites {
```

build/rewrite.go
gtong, line 101: Doing this has the side effect of changing the passed-in Rewriter object, which would be an unexpected side-effect. I would suggest passing along the nil values as they are now, and creating local variables where the values are used.

build/rewrite.go
```diff
@@ -237,7 +294,9 @@ func fixLabels(f *File) {
 					continue
 				}
 				key, ok := as.LHS.(*Ident)
-				if !ok || !tables.IsLabelArg[key.Name] || tables.LabelDenylist[callName(v)+"."+key.Name] {
+				var isLabelArg map[string]bool = *w.LabelArg
```
gtong, line 297: So this would become something like:
```
var isLabelArg map[string]bool = tables.IsLabelArg
if w.LabelArg != nil {
  isLabelArg = w.LabelArg
}
```

build/rewrite.go
```diff
@@ -54,17 +57,73 @@ func allowedSort(name string) bool {
...
+	SortAllowList                   *map[string]bool
+	NamePriority                    *map[string]int
+	StripLabelLeadingSlashes        *bool
+	ShortenAbsoluteLabelsToRelative *bool
+}
+
 func Rewrite(f *File) {
+	var rewriter = &Rewriter{
+		LabelArg:                        &tables.IsLabelArg,
+		Denylist:                        &tables.LabelDenylist,
```
gtong, line 80: Why the renames? I'd expect the fields in the Rewriter struct to correspond 1:1 to the variables in the tables package.

build/rewrite.go
```diff
@@ -54,17 +57,73 @@ func allowedSort(name string) bool {
...
+}
+
+func (w *Rewriter) Rewrite(f *File) {
+	initRewriter(w)
 	for _, r := range rewrites {
-		if !disabled(r.name) {
-			if f.Type&r.scope != 0 {
-				r.fn(f)
-			}
+		if !disabled(w, r.name) && (f.Type&r.scope != 0) || (w.DisableRewrites != nil) {
```
gtong, line 94: Should be able to simplify this if clause now.

go.sum
```diff
@@ -5,6 +5,9 @@ github.com/chzyer/logex v1.1.10/go.mod h1:+Ywpsq7O8HXn0nuIou7OrIPyXbp3wmkHB+jjWR
 github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e/go.mod h1:nSuG5e5PlCu98SY8svDHJxuZscDgtXS6KTTbou5AhLI=
 github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1/go.mod h1:Q3SI9o4m/ZMnBNeIyt5eFwwo7qiLfzFZmjNmxjkiQlU=
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
+github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
```
gtong, line 8: Wonder why these were added.

gtong (changes requested):

build/rewrite.go
```diff
@@ -54,17 +57,73 @@ func allowedSort(name string) bool {
...
+	SortableListArg                 *map[string]bool
+	SortDenyList                    *map[string]bool
+	SortAllowList                   *map[string]bool
+	NamePriority                    *map[string]int
+	StripLabelLeadingSlashes        *bool
+	ShortenAbsoluteLabelsToRelative *bool
+}
+
 func Rewrite(f *File) {
+	var rewriter = &Rewriter{
```
gtong, line 78: This can now just be:
```
var rewriter := &Rewriter{}
rewriter.Rewrite(f)
```

build/rewrite.go
```diff
@@ -54,13 +58,33 @@ func allowedSort(name string) bool {
...
+}
+
+func (w *Rewriter) Rewrite(f *File) {
 	for _, r := range rewrites {
-		if !disabled(r.name) {
-			if f.Type&r.scope != 0 {
-				r.fn(f)
-			}
+		// f.Type&r.scope is a bitwise comparison
+		if !disabled(w, r.name) && (f.Type&r.scope != 0) || (w.DisableRewrites != nil && !disabled(w, r.name)) {
```
rgwgoogle, line 86: Explanation of this if statement - We require the conditional or here because (f.Type&r.scope != 0) is intended ot check the scope. For starlark files, the scope tells us not to do call sort which is why we need to have our own check on the right side.

build/rewrite.go
kleerwater: I think the code would be a lot clearer if the Rewriter didn't have pointers for all of its members and the Rewrite function initialized the Rewriter with the values from the tables. Then the rest of the code can just use the value from the Rewriter instance directly.

vladmos:
@rgwgoogle Could you please not overwrite the existing commit, but add other commits instead? It's hard to track what has been changed since the last time I reviewed the PR.

build/rewrite.go
```diff
@@ -54,13 +58,33 @@ func allowedSort(name string) bool {
...
 }
 
-// Rewrite applies the high-level Buildifier rewrites to f, modifying it in place.
+// Rewrites controls the rewrites to be applied.
+//
+// If non-nil, the rewrites with the specified names will be run. If
+// nil, a default set of rewrites will be used that is determined by
+// the type (BUILD vs default starlark) of the file being rewritten.
+type Rewriter struct {
+	DisableRewrites                 *[]string
```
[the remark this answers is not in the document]
gtong: Our thinking was to mimic the DisableRewrites package-level variable, but making this explicit is fine too.

build/rewrite.go
gtong: Fair enough. Would add that as a comment in the code.

build/rewrite.go
rgwgoogle: We ended up going with this decision because we did not want to modify the Rewrite object when its passed through

kleerwater: This would not require modifying any objects that are passed in. This just would change it so that when the Rewriter object is created in the Rewrite function it gets initialized with the tables values. The rest of the code doesn't need to modify anything, it can just rely on the values in the Rewriter instance being authoritative.

build/rewrite.go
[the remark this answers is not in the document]
kleerwater: The DisableRewrites package-level variable is used by a command-line flag for disabling rewrites on a particular invocation, I think we should avoid conflating that use case with this one.

build/rewrite.go
```diff
@@ -54,17 +54,55 @@ func allowedSort(name string) bool {
...
+	rewriter.Rewrite(f)
+}
+
+func (w *Rewriter) Rewrite(f *File) {
 	for _, r := range rewrites {
-		if !disabled(r.name) {
-			if f.Type&r.scope != 0 {
-				r.fn(f)
-			}
+		// f.Type&r.scope is a bitwise comparison. Because starlark files result in a scope that will
```
kleerwater, line 90: This is going to result in incorrect behavior where if a file type matches the rewrite scope then the rewrite gets applied even if the rewriter specifies the rewrite set and doesn't include the rewrite.

It should be something like:
```
if disabled(r.name) {
  continue
}
if w.RewriteSet != nil {
  if !rewriteSetContains(w, r.name) {
    continue
  }
} else {
  if f.Type&r.scope == 0 {
    continue
  }
}
r.fn(f, w)
```

build/rewrite.go
```diff
@@ -389,6 +427,18 @@ func sortStringLists(f *File) {
 				return
 			}
 			rule := callName(v)
+			var sortDenyList map[string]bool = tables.SortableDenylist
```
kleerwater, line 430: This should be updated to not use tables. The only place in the file that should be referring to the tables package is the initialization of the Rewriter instance in Rewrite()

build/rewrite.go
rgwgoogle: Good catch! I will add another check on the left side stating that w.Rewrites is empty

build/rewrite.go
```diff
@@ -64,31 +60,49 @@ func allowedSort(name string) bool {
...
-	DisableRewrites                 *[]string
-	IsLabelArg                      *map[string]bool
-	LabelDenyList                   *map[string]bool
-	IsSortableListArg               *map[string]bool
-	SortableDenylist                *map[string]bool
-	SortableAllowlist               *map[string]bool
-	NamePriority                    *map[string]int
-	StripLabelLeadingSlashes        *bool
-	ShortenAbsoluteLabelsToRelative *bool
+	RewriteSet                      *map[string]struct{}
```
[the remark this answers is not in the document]
rgwgoogle: Changing to get rid of pointer. I'm curious as to if changing it to a slice would change it besides adding in a loop to go through the set?

build/rewrite.go
```diff
@@ -32,7 +32,7 @@ import (
 var DisableRewrites []string
 
 // disabled reports whether the named rewrite is disabled.
-func disabled(name string) bool {
+func disabled(w *Rewriter, name string) bool {
```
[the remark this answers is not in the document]
rgwgoogle: Good catch!

build/rewrite.go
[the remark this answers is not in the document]
kleerwater: Well, it makes declaring it more straightforward, you just create a slice with the names of the rewrites instead of having to set entries in a map where the value is unused. If someone that's more knowledgeable about go says that using a map as a set like this is idiomatic code, then feel free to stick with it.

build/rewrite.go
kleerwater: You shouldn't check for empty rewrites. If the rewrites is non-nil but empty, I would say that should just not perform any rewrites, which is what will happen with my suggested code.

kleerwater: nm, I see what you meant

build/rewrite.go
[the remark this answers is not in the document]
rgwgoogle: Gotcha. I'm not entirely convinced it's more straightforward to initialize it as a slice vs map. Both seem viable here. I would argue that it's more intuitive for checking if something exists by checking for a key rather than adding a for loop through the slice. I'm going to leave it as is unless someone has a strong opinion on this.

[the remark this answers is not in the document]
kleerwater: I definitely think it's more straightforward to say rewriteSet = []string{"callsort", "loadsort"} than having to initialize a map with values that are unused, but if sticking with the map, can you make the value type interface{} and then the value can just be nil. There should also be a comment indicating that the values are ignored.

build/rewrite.go
rgwgoogle: Gotcha fixed

build/rewrite.go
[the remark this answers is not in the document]
rgwgoogle: Okay, I ended up changing it back to a slice :)

vladmos (approved):

PR #1072, opened 2022-06-30
Remove deprecation notices for modes: check, diff, print_if_changed

sudoforge:
```
Remove deprecation notices for modes: check, diff, print_if_changed

Deprecation of these values for the \`-mode\` flag is not planned.

Closes: #1005
```

PR #1076, opened 2022-07-14
Remove deprecation notices for modes: check, diff, print_if_changed

sudoforge:
```
Deprecation of these values for the `-mode` flag is not planned.

Closes: #1005
```

Replaces #1072


sudoforge:
ignore the branch deletion; forgot to filter that out in a script that was deleting all non-head branches in my forks

PR #1078, opened 2022-07-31
Buildozer new command: copy rule

### LINKED ISSUE #134: Buildozer new command: copy rule
I'd like a buildozer command that copies a rule. Right now buildozer has `copy <attr> <from_rule>` to copy attribute values, and `new <rule_kind> <rule_name>` to make a new rule, but I would like:

* a new command `copy_rule <from_rule> <rule_name>`
* or perhaps a new arg on `new`: `new <rule_kind> <rule_name> [based_on <from_rule>]`

This can only be achieved today using scary usage of things like `print rule` or `print startline` and `print endline` combined with `sed` or something... :fearful: some other tricky scripting.

jmcrawford45:
Fixes #134


dws:
Can we get this landed?

PR #1083, opened 2022-08-18
Add AttrMap() function for getting rule attributes

aqua4:


build/rule.go
```diff
@@ -212,6 +212,15 @@ func (r *Rule) AttrKeys() []string {
 	return keys
 }
 
+// AttrMap returns the map of attribute keys to values.
+func (r *Rule) AttrMap() map[string]Expr {
+	m := make(map[string]Expr)
+	for _, key := range r.AttrKeys() {
+		m[key] = r.Attr(key)
```
laurentlb, line 219: The `Attr` function iterates over the arguments. So your function will have complexity O(n^2).

You should iterate explicitly over r.Call.List instead of calling AttrKeys.

PR #1091, opened 2022-09-30
Consolidate File.Rules() and RulesAt()

cushon:
and allow RulesAt to find targets nested in list comprehensions, matching the behaviour added to Rules() in
82b21607e00913b16fe1c51bec80232d9d6de31c.


vladmos (approved):
Thanks!

PR #1096, opened 2022-11-04
Support deleting wildcard attrs

illicitonion:
When e.g. changing a rule type from http_archive to git_repository, it can be useful to remove all attributes before populating new ones.

Effectively this supports "delete and replace", without needing to work out where the replacement rule needs inserting into the file.


vladmos (approved):
Thanks!

PR #1097, opened 2022-11-07
recognize various test_* arg conventions

sxlijin:
test_data, test_deps, test_srcs, and test_tags are all commonly used conventions for rules. Buildifier should recognize these conventions and sort them; c.f. examples in [android](https://cs.github.com/JetBrains/android/blob/b373163869f676145341d60985cdbdaca3565c0b/project-system/BUILD#L9), [kythe](https://cs.github.com/cpehle/kythe/blob/ca97abf654f3ec7a099ff9ce93a0e041005b842d/kythe/go/serving/xrefs/BUILD#L6), [tensorflow](https://cs.github.com/418sec/tensorflow/blob/2fb5d641d852ff40df2dd85f461115ab9583d851/tensorflow/lite/testing/BUILD#L67), and other repos (e.g. [bbai](https://cs.github.com/rnburn/bbai-lib/blob/3a985793c34857ede615092e1efc7934d6d49a5b/example/median/BUILD#L17)).

You should also be able to confirm this in google3 with something like `csearch -l 'test_(data|deps|srcs|tags) = \[' | wc -l`.


vladmos:
I think it makes sense to recognize the attributes, however it will also make a lot of existing files misformatted (if the attributes are not sorted, and if that's not been enforced by buildifier that's probably the case). They need to be formatted before these changes go live. I can try to do that, but not in the next few weeks, perhaps in December.

sxlijin:
Ack, thanks for letting me know.

I'm guessing the buildifier presubmit is now auto-released, which is why you can't decouple this merge and the release?

vladmos:
I can in theory disable this internally, but then the change will be a no-op there.

sxlijin:
I'm happy to let this brew for a month or two while you get the LSC together. If it hasn't landed by, say, January, then I'd appreciate it if you would consider disabling this inside google3.

sxlijin:
Hey, just checking in: how's the LSC going?

sxlijin:
Bump :)

sxlijin:
Checking in again - any progress on the LSC?

vladmos:
Hi, is there a chance you can help me with the LSC? This way we can submit it faster because I'll be able to global-approve the changes myself.

sxlijin:
I haven't worked at Google since 2021 :( otherwise I would

vladmos (approved):

PR #1099, opened 2022-11-12
Add remove_if_equal command to buildozer

### LINKED ISSUE #105: Buildozer doesn't have a way to remove an attribute if it has a specific value
I want to clear attributes that are set to a specific value. The use case is that an option was provided that had one default, and the default has switched. I now want to clear out all the settings where the new default is explicitly set.

dfreese:
I've left open the structure to removing more complex expressions, but trigger an error in all but the more basic cases.  I wasn't sure what sort of behavior this type of command should have in more complex cases.

I generate a comparison function up-front, which does a lot of virtual switching on types as there didn't seem to be a way to genericly compare expressions.  I've used getAttrValueExpr, however, this does make for a little weirdness in trying to handle the case of strings versus identifiers.  The test case with `"True"` versus `True` tries to demonstrate why identifiers are compared to both string and identifiers.

I assume this will seem incomplete in some respects, but I was putting this up for initial feedback on the approach as well as what would be expected in the test coverage.

Fixes #105

### DIFF
--- a/buildozer/buildozer_test.sh
+++ b/buildozer/buildozer_test.sh
@@ -447,6 +447,42 @@ function test_remove_package_attribute() {
   [ $(wc -c < "$root/pkg/BUILD") -eq 0 ] || fail "Expected empty file"
 }
 
+function test_remove_if_equal_label() {
+  in='go_library(
+    name = "edit",
+    shared_library = ":local",  # Suffix comment.
+)'
+  run "$in" 'remove_if_equal shared_library :local' '//pkg:edit'
+  assert_equals 'go_library(name = "edit")'
+}
+
+function test_remove_if_equal_ident() {
+  in='go_library(
+    name = "edit",
+    flag = True,
+)'
+  run "$in" 'remove_if_equal flag True' '//pkg:edit'
+  assert_equals 'go_library(name = "edit")'
+}
+
+function test_remove_if_equal_string() {
+  in='go_library(
+    name = "edit",
+    flag = "True",
+)'
+  run "$in" 'remove_if_equal flag True' '//pkg:edit'
+  assert_equals 'go_library(name = "edit")'
+}
+
+function test_remove_if_equal_string_attr_string() {
+  in='go_library(
+    name = "edit",
+    toolchain = "something",
+)'
+  run "$in" 'remove_if_equal toolchain something' '//pkg:edit'
+  assert_equals 'go_library(name = "edit")'
+}
+
 function test_move_last_dep() {
   run "$one_dep" 'move deps runtime_deps //buildifier:build' '//pkg:edit'
   assert_equals 'go_library(
--- a/buildozer/buildozer_test.sh
+++ b/buildozer/buildozer_test.sh
@@ -456,6 +456,27 @@ function test_remove_if_equal_label() {
   assert_equals 'go_library(name = "edit")'
 }
 
+function test_remove_if_equal_label_does_not_match() {
+  in='go_library(
+    name = "edit",
+    shared_library = ":local",  # Suffix comment.
+)'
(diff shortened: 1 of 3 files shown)


buildozer/buildozer_test.sh
```diff
@@ -447,6 +447,42 @@ function test_remove_package_attribute() {
...
+
+function test_remove_if_equal_string_attr_string() {
+  in='go_library(
+    name = "edit",
+    toolchain = "something",
+)'
+  run "$in" 'remove_if_equal toolchain something' '//pkg:edit'
+  assert_equals 'go_library(name = "edit")'
+}
+
```
laurentlb, line 485: You should add a negative test as well (where the value is not equal)

edit/buildozer.go
```diff
@@ -417,6 +417,46 @@ func cmdRemove(opts *Options, env CmdEnvironment) (*build.File, error) {
 	return nil, nil
 }
 
+func cmdRemoveIfEqual(opts *Options, env CmdEnvironment) (*build.File, error) {
+	attr := env.Args[0]
+	args := env.Args[1:]
+	expr := getAttrValueExpr(attr, args, env)
+	cmp, err := func() (func(build.Expr) bool, error) {
```
laurentlb, line 424: Why is it in a function? Might be more readable to remove this function which is executed immediately.

edit/buildozer.go
```diff
@@ -417,6 +417,46 @@ func cmdRemove(opts *Options, env CmdEnvironment) (*build.File, error) {
...
+				case *build.StringExpr:
+					return labels.Equal(val.Value, input.Name, env.Pkg)
+				case *build.Ident:
+					return val.Name == input.Name
+				default:
+					return false
+				}
+			}, nil
+		default:
+			return nil, fmt.Errorf("unsupported expression type %T", input)
```
laurentlb, line 445: Do we need to show an error here? I'd assume that the condition is false, so we can silently do nothing.

edit/buildozer.go
```diff
@@ -417,6 +417,46 @@ func cmdRemove(opts *Options, env CmdEnvironment) (*build.File, error) {
 	return nil, nil
 }
 
+func cmdRemoveIfEqual(opts *Options, env CmdEnvironment) (*build.File, error) {
+	attr := env.Args[0]
+	args := env.Args[1:]
+	expr := getAttrValueExpr(attr, args, env)
```
laurentlb, line 423: getAttrValueExpr is using some heuristics to guess the expected type of an attribute. It is a bit brittle and I think we don't need this function here.

e.g. If you call `env.Rule.Attr` and get a string expression, you can just compare the content of the string with `attr`. No need to build an expr.

laurentlb (changes requested):

buildozer/buildozer_test.sh
dfreese: done

edit/buildozer.go
dfreese: done

edit/buildozer.go
dfreese: I tend to find immediately invoked functions, and a single assignment more clear.  I tend to write more C++ where that allows the variable to be const, but that isn't an advantage here.  Thanks for pointing that out; removed.

edit/buildozer.go
dfreese: done

dfreese (commented):
Thanks for the review, and sorry for the delay in my response.  Personal matters came up.

vladmos (changes requested):
Looks good, but please also document the new command in buildozer/README.md

dfreese:
> Looks good, but please also document the new command in buildozer/README.md

Thanks.  Done.

vladmos (approved):
Thanks!

laurentlb (approved):

PR #1100, opened 2022-11-15
fix fix.go to stop creating empty package() decls.

bimargulies-google:
Several places in fix.go called 'PackageDeclaration' instead of 'ExistingPackageDeclaration'. This would add an empty package declaration if none was present.


laurentlb:
The assumption was that we would call `RemoveEmptyPackage` afterwards.

The change looks good to me, but I'll let @vladmos decide if the new dependency on `com_github_google_go_cmp` is worth adding.

PR #1104, opened 2022-11-25
Add a new warning for named args in dict methods

laurentlb:
For reference: https://github.com/bazelbuild/starlark/blob/master/spec.md#dict%C2%B7get


vladmos (approved):
Thanks!

PR #1105, opened 2022-11-26
new_git_repository deprecation.

aiuto:
- Stop advising people to use `new_git_repository`
- Fix workspace tests to account for both being equivalent now.

This is only part of the deprecation.  We need an entirely new feature to rewrite new_git_repository as git_repository. I'm leaving that for someone else.  It's not obvious that we should just do it, because then buildifier would only work for bazel >= 6.x. So what we really need is for buildifier to understand bazel versions, or be tracked by LTS.  For now, I am happy to stop advising people to use the deprecated version.

https://github.com/bazelbuild/bazel/issues/16823


vladmos (approved):
Thanks!

PR #1107, opened 2022-12-02
Add loads after workspace() declarations

illicitonion:
Bazel requires that workspace() declarations come before any loads.

Otherwise buildozer produces WORKSPACE files which fail bazel with the following error:
```
Error in workspace: workspace() function should be used only at the top of the WORKSPACE file
```


vladmos (approved):
Thanks!

PR #1108, opened 2022-12-09
Support repo-absolute file paths in buildozer

### LINKED ISSUE #314: allow buildozer to operate on WORKSPACE files
WORKSPACE files are written in the same format as BUILD files and can be formatted by buildifier. I would like to use buildozer in a script that queries the contents of a WORKSPACE file, uses that output to check for newer upstream versions of the packages, then writes out the updates to the WORKSPACE as buildozer commands.

If I rename my WORKSPACE to BUILD, it works just fine. For example:
```
$ bazel run //tools:buildozer -- 'print label urls strip_prefix' //:%http_archive
INFO: Analysed target //tools:buildozer (0 packages loaded).
INFO: Found 1 target...
Target //tools:buildozer up-to-date:
  bazel-bin/tools/buildozer
INFO: Elapsed time: 0.460s, Critical Path: 0.00s
INFO: Build completed successfully, 1 total action
rule "//:gmaven_rules" has no attribute "urls"
//:io_grpc_grpc_java [https://github.com/grpc/grpc-java/archive/v1.12.0.tar.gz] grpc-java-1.12.0
//:gmaven_rules (missing) "gmaven_rules-%s" % GMAVEN_TAG
//:com_google_protobuf [https://github.com/google/protobuf/archive/3.5.1.1.zip] protobuf-3.5.1.1
//:com_google_googletest [https://github.com/google/googletest/archive/08d5b1f33af8c18785fb8ca02792b5fac81e248f.zip] googletest-08d5b1f33af8c18785fb8ca02792b5fac81e248f
//:com_github_google_benchmark [https://github.com/google/benchmark/archive/6d74c0625b8e88c1afce72b4f383c91b9a99dbe6.zip] benchmark-6d74c0625b8e88c1afce72b4f383c91b9a99dbe6
//:com_google_absl [https://github.com/abseil/abseil-cpp/archive/59ae4d5a0e833bedd9d7cc059ac15a9dc130e3f7.zip] abseil-cpp-59ae4d5a0e833bedd9d7cc059ac15a9dc130e3f7
```
Another way to achieve this would be to allow the user to pass in the path to a single BUILD file, though I'm not sure whether buildozer needs to open other BUILD files relative to the first.

fmeum:
In addition to the pseudo-label syntax for relative file paths (e.g. `WORKSPACE` or `toolchains/BUILD.tpl`), also support absolute labels (e.g. `//WORKSPACE` or `//toolchains/BUILD.tpl`).

Also documents the syntax for files in the README.

Fixes #314

### DIFF
--- a/edit/edit.go
+++ b/edit/edit.go
@@ -73,13 +73,20 @@ func InterpretLabelForWorkspaceLocation(root string, target string) (buildFile s
 
 	defaultBuildFileName := "BUILD"
 	if strings.HasPrefix(target, "//") {
+		pkgPath := filepath.Join(rootDir, pkg)
+		if isFile(pkgPath) {
+			// allow operation on other files like WORKSPACE
+			buildFile = pkgPath
+			pkg = filepath.Dir(pkg)
+			return
+		}
 		for _, buildFileName := range BuildFileNames {
-			buildFile = filepath.Join(rootDir, pkg, buildFileName)
+			buildFile = filepath.Join(pkgPath, buildFileName)
 			if isFile(buildFile) {
 				return
 			}
 		}
-		buildFile = filepath.Join(rootDir, pkg, defaultBuildFileName)
+		buildFile = filepath.Join(pkgPath, defaultBuildFileName)
 		return
 	}
 	if isFile(pkg) {
--- a/edit/edit.go
+++ b/edit/edit.go
@@ -21,6 +21,7 @@ package edit
 import (
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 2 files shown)


edit/edit.go
```diff
@@ -73,13 +73,20 @@ func InterpretLabelForWorkspaceLocation(root string, target string) (buildFile s
 
 	defaultBuildFileName := "BUILD"
 	if strings.HasPrefix(target, "//") {
+		pkgPath := filepath.Join(rootDir, pkg)
+		if isFile(pkgPath) {
+			// allow operation on other files like WORKSPACE
+			buildFile = pkgPath
+			pkg = filepath.Dir(pkg)
```
vladmos, line 80: Shouldn't it be `path.Dir` instead? `filepath` is OS-specific and works correctly with local paths, but packages defined by labels always have forward slashes, `path.Dir` seems to be a better solution here (even if `filepath.Dir` parses paths with forward slashes too).

fmeum: I copied this from below, but pushed a new commit that uses `path.Dir` and also explicitly uses `filepath.FromSlash` to convert logical to platform paths.

PR #1113, opened 2022-12-28
Sort and deduplicate `use_repo` positionals 

fmeum:
After refactoring out existing logic to sort string expressions in mixed lists, this logic is reused to sort the positional arguments in `use_repo` calls in chunks delimited by comments. Keyword arguments were already sorted due in the generic call argument sorting pass.

PR #1114, opened 2022-12-28
Sort `version` right after `name` in `MODULE.bazel`

fmeum:
Since name and version are what uniquely determines a Bazel module, these attributes should be right at the top of both `module` and `bazel_dep` calls. This can also simplify the job of external dependency management tooling (potentially based on regular expressions).

### DIFF
--- a/buildifier/integration_test.sh
+++ b/buildifier/integration_test.sh
@@ -49,9 +49,9 @@ mkdir test_dir/workspace  # name of a starlark file, but a directory
 mkdir test_dir/.git  # contents should be ignored
 echo -e "a+b" > test_dir/.git/git.bzl
 cat > test_dir/MODULE.bazel <<'EOF'
-module(name='my-module',version='1.0')
+module(name='my-module',version='1.0',compatibility_level=1)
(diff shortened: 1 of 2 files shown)


fmeum:
@meteorcloudy

fmeum:
Stacked on #1113

meteorcloudy (approved):
LGTM for the third commit

vladmos:
Does the description belong to some other CL? I don't see any instance of "version" in the diff (except for the tests).

fmeum:
> Does the description belong to some other CL? I don't see any instance of "version" in the diff (except for the tests).

The third commit, https://github.com/bazelbuild/buildtools/pull/1114/commits/33a1af313a3a722378c0092de2b886ca9b4de9df, which is the only one in this PR that isn't part of another PR, adds two `version` attributes to the priority list.

vladmos:
I see, let me merge the other PR first then.

fmeum:
@vladmos Unstacked

vladmos (approved):
Thanks!

PR #1116, opened 2023-01-04
Format module overrides

fmeum:
Module overrides are grouped together with the `bazel_dep` they override and have their `module_name` argument sorted first.


fmeum:
@vladmos @meteorcloudy Could you take a look? This is another (probably the last) `MODULE.bazel` improvement that prevents unnatural formatting.

vladmos (approved):

fmeum:
@vladmos Thanks for the reviews! Is the next release already planned? It would be great to get one with all the `MODULE.bazel` fixes.

PR #1121, opened 2023-01-05
Fix handling of synthetic package declarations

vladmos:
It's not correct to make exceptions for empty package declarations because they may become non-empty by a subsequent buildozer command. Instead, they need to be inserted at the correct position (namely after `workspace()` calls) so that the order of the statements is always correct (docstring -> `workspace()` -> `load()` -> `package()`).

### DIFF
--- a/edit/edit.go
+++ b/edit/edit.go
@@ -151,26 +151,32 @@ func PackageDeclaration(f *build.File) *build.Rule {
 	if pkg := ExistingPackageDeclaration(f); pkg != nil {
 		return pkg
 	}
-	all := []build.Expr{}
-	added := false
+	insertAfter := -1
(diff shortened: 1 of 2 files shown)


edit/edit.go
```diff
@@ -151,26 +151,32 @@ func PackageDeclaration(f *build.File) *build.Rule {
...
 			}
+		default:
 		}
-		all = append(all, stmt)
-	}
-	if !added { // In case the file is empty.
-		all = append(all, call)
+		break
 	}
+	all := []build.Expr{}
```
pmbethe09, line 174: nit: use 
var all []build.Expr

pmbethe09 (approved):
LGTM

PR #1127, opened 2023-01-23
Add `File#RuleNamed` Method

jkjk822:
Add method to retrieve rule by name from a file pointer, similar to edit#FindRuleByName. This allows access from the file object itself instead of needing an extra package.


vladmos (approved):
Thanks!

PR #1092, opened 2022-10-10
Allow buildifier_test to escape the sandbox

milesdai:
This PR addresses #1075 by adding a `no_sandbox` attribute and a WORKSPACE path attribute to`buildifier_test`. Currently, users must explicitly specify every BUILD file they would like to lint in the `srcs` attribute of `buildifier_test`. This PR allows the user to create a test to lint all BUILD files as follows:

```
buildifier_test(
    name = "buildifier_check",
    diff_command = "diff -u",
    mode = "diff",
    no_sandbox = True,
    workspace = "//:WORKSPACE",
)
```

PR #1128, opened 2023-01-24
Improves portability of buildifier_test

jszumski:
Follow up to https://github.com/bazelbuild/buildtools/pull/1092, which doesn't run on macOS 12.x due to missing `realpath`.  This change falls back to `readlink` in that case and errors if neither is available.

Note that this codepath is only used when the `workspace` argument is provided to break out of the test sandbox:
```
buildifier_test(
    name = "buildifier-check",
    mode = "check",
    no_sandbox = True,
    workspace = "//:WORKSPACE.bazel",
)
```


vladmos (approved):
Thanks!

PR #1133, opened 2023-01-29
devcontainer.json

sri-ka1kiZ:
templatre

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,5 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/universal:2",
+  "features": {
+  }
+}


meisterT:
I have never used a devcontainer - can you describe what the easiest way it is to try and what the container provides? Is Bazel installed in the container?

vladmos:
I too have never used devcontainers, but if that's something that is specific to your development process, consider adding `.devcontainer` to the `.gitignore` file, this way you'll be able to keep your local development config without having it interfere with git.

sluongng:
I am a big fan of .devcontainer and remote workspace (such as Github Codespaces)... but this PR is a spam and I think we should close it.

Bazel remote workspace require some elaboration to setup that at minimum, something like https://github.com/google/gvisor/blob/master/.devcontainer.json

PR #1137, opened 2023-02-13
Preserve loads if they're different

illicitonion:
WORKSPACE files currently allow shadowing loads (see https://github.com/bazelbuild/bazel/issues/17480), i.e. the following is valid:

```starlark
load("//:one.bzl", "fn")

fn()

load("//:two.bzl", "fn")

fn()
```

Currently buildozer removes the second load because the symbol was already loaded. This is incorrect - removing the second load changes the behaviour of this file.

Instead, only delete a load of the last time it was loaded was from the same place.


vladmos (approved):
Thanks, sorry for the delay!

PR #1143, opened 2023-03-15
Automatically sort string lists within complex statements

vladmos:
Buildifier currently sorts some string lists (such as `srcs` or `deps` arguments) automatically, but it doesn't sort anything if the value is not a list but a complex expression that contains valid sortable lists inside (such as list + constant, or a select statement with inner lists). This commit fixes that behaviour.

The change doesn't introduce any new sortable parameters, it only touches the parameters that would have been sorted anyway if they were plain string lists.

The change is not backwards compatible and will require a global cleanup on the existing files.


build/rewrite.go
```diff
@@ -448,12 +448,12 @@ func sortStringLists(f *File, w *Rewriter) {
 					continue
 				}
 				if w.IsSortableListArg[key.Name] ||
-					w.SortableAllowlist[context] ||
-					(!disabled("unsafesort") && allowedSort(context)) {
+						w.SortableAllowlist[context] ||
```
meisterT, line 451: nit: indentation looks broken

build/rewrite.go
```diff
@@ -560,6 +560,37 @@ func SortStringList(x Expr) {
 	list.List = sortStringExprs(list.List)
 }
 
+// FindAndSortStringLists finds and sorts string lists recursively withing
```
meisterT, line 563: nit: within

build/rewrite.go
vladmos: Fixed with gofmt

build/rewrite.go
vladmos: Done.

meisterT (approved):

laurentlb (approved):
lgtm

PR #1146, opened 2023-03-15
Shorten labels within complex expressions

vladmos:
If buildifier shortens labels within a single list of labels, it should also short them if the value of the argument is a complex statement such as a concatenation of lists, select statements and other nodes.

This is similar to #1143 but focuses on shortening and joining labels rather than sorting them.

### DIFF
--- a/build/testdata/070.build.golden
+++ b/build/testdata/070.build.golden
@@ -4,8 +4,8 @@ foo(
         "aaa",
         "bbb",
     ] + CONSTANT + [
-        "ccc",
-        "ddd",
+        "//ccc",
+        "//ddd",
     ],
     data = [
         "iii",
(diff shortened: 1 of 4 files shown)


build/testdata/070.build.golden
```diff
@@ -4,8 +4,8 @@ foo(
         "aaa",
         "bbb",
     ] + CONSTANT + [
-        "ccc",
-        "ddd",
+        "//ccc",
+        "//ddd",
```
laurentlb, line 8: why? This looks like a bug?

vladmos: You mean that `srcs` is interpreted as labels?

laurentlb: oops, I missed the change in the .in file
lgtm

laurentlb (approved):

meisterT (approved):

Issue #1149, opened 2023-03-17
Proposal: Add command that manages `use_repo` calls


fmeum:
Due to Bzlmod's [strict deps](https://docs.bazel.build/versions/5.0.0/bzlmod.html#strict-deps) feature, it is necessary to declare all repositories generated by a module extension that a particular module wants to use via a `use_repo` call. With module extensions that potentially generate a large number of repositories (~1000) such as gazelle's `go_deps`, this can be tedious and error-prone. 

Buildozer would be the ideal tool for automatically manipulating `use_repo` calls. Bazel itself as well as individual module extensions could then generate buildozer commands to fixup or add `use_repo` calls as needed.

## Proposal

Add the following buildozer command:

```
a) buildozer 'use_repo_add [dev] <absolute .bzl file label> <extension name> <repository name>...' //MODULE.bazel:all
b) buildozer 'use_repo_remove [dev] <absolute .bzl file label> <extension name> <repository name>...' //MODULE.bazel:all
```

The command would operate as follows (*statement* is always meant to refer to a top-level statement only):

1. Parse all `X = use_extension(Y, Z, kwargs)` statements where X is an identifier and Y and Z are string literals and where `kwargs` contains `dev_dependency = True` if and only if the optional `dev` argument is given.
   Store the list of all `X` for which `Y` and `Z` are the first and second argument of the `use_repo` command. This comparison should take into account that extensions in the current module can be referred to via relative labels: If the first argument has a repository name that equals the `repo_name` (or `name`, if absent) of the `module` call, it should also be considered a match.
2. Parse all `use_repo(X, ...)` calls where `X` is in the list obtained in step 2.
   (If no such call exists, add one right after the last assignment statement `X = ...` for an `X` in the list obtained in step 2.)
3. Depending on whether `use_repo_add` or `use_repo_remove` is used:
  a) Add all repository name arguments that do not exist as positional arguments or keyword argument string values on any such `use_repo` call as positional string arguments to the first such call.
  b) Remove all positional string arguments that match a repository name argument, as well as all keyword arguments whose string value does so.
5. Remove all statements of the form `use_repo(X)` (i.e., `use_repo` calls that list no repositories) as no-ops.

## Implementation

A prototype based on the `build` package exists at https://github.com/fmeum/auto_use_repo/blob/main/update/update.go, but should be improved and integrated into buildozer.

## Example

```
[...]

PR #1150, opened 2023-03-18
Fix empty `package()` calls appearing in non-BUILD files

fmeum:
Caused by https://github.com/bazelbuild/buildtools/commit/9c3c1fc0106ed2174f1a7c4a322a08f0a553bc1e, but only relevant if users use the pseudo-target `__pkg__` with non-BUILD files as recommended for file-level operations.

Work towards #1149 
Work towards https://github.com/bazelbuild/bazel/issues/17908


fmeum:
@laurentlb I observed this small regression caused by my last commit when I tried to use `//MODULE.bazel:__pkg__` to skip per-target fixes.

I will try to follow up with a change that provides a more direct way to achieve the same goal.

Wyverald:
is there precedent for editing other non-BUILD files? I find the syntax `//MODULE.bazel:__pkg__` very unappealing. I'm guessing that `//:MODULE.bazel` is off the table since it would try to edit the `MODULE.bazel` target in the `BUILD` file?

fmeum:
`//MODULE.bazel:all` also works and is what I am currently using. Maybe we could make `//MODULE.bazel` work? It *could* also be referring to a `bazel_dep(name = "MODULE.bazel")` though.

Wyverald:
> Maybe we could make `//MODULE.bazel` work? It _could_ also be referring to a `bazel_dep(name = "MODULE.bazel")` though.

Did you maybe mean `java_library(name = "MODULE.bazel")` or something? I didn't know that each individual `bazel_dep` could be referred to.

fmeum:
> Did you maybe mean `java_library(name = "MODULE.bazel")` or something? I didn't know that each individual `bazel_dep` could be referred to.

From the point of view of buildozer, all rules are the same. As long as your tags have a `name` attribute, you can even operate on them. We are using this to emit migration commands in Gazelle that look like this:

```
new go_deps.gazelle_override gopkg.in/yaml.v3|//MODULE.bazel:all
add directives gazelle:go_naming_convention\ go_default_library|//MODULE.bazel:gopkg.in/yaml.v3
rename name path|//MODULE.bazel:gopkg.in/yaml.v3
remove build_naming_convention|//MODULE.bazel:%go_deps.module
remove build_file_proto_mode|//MODULE.bazel:%go_deps.module
```

But having anything in `MODULE.bazel` with `name = "MODULE.bazel"` would still be very weird.

Wyverald:
Ah I see. I guess my main gripe is that `//MODULE.bazel:foo` changes the `foo` "target" in the file `/MODULE.bazel`, instead of the `foo` target in the file `/MODULE.bazel/BUILD`.

I think I'd prefer a special syntax, something like `//:__module__`, for the MODULE.bazel file. That means you'd need difference syntax to operate on the stuff in there, but we're already kind of doing that (with `use_repo_add`, for example).

fmeum:
@vladmos Could you take a look at this fix? I would submit a separate PR that proposes special syntax to identify the `MODULE.bazel` file.

PR #1151, opened 2023-03-27
Add external repo support to unused_deps.

brown:
External repo targets can be passed as arguments to unused_deps and unused external repo dependencies are detected.

When an external repo BUILD file needs to be edited to remove an unused dependency, the generated buildozer command will need some postprocessing.  The unused_deps utility does not know the file system location of the BUILD file, just its external repo label.

### DIFF
--- a/edit/edit.go
+++ b/edit/edit.go
@@ -41,9 +41,9 @@ var (
 // InterpretLabelForWorkspaceLocation returns the name of the BUILD file to
 // edit, the full package name, and the rule. It takes a workspace-rooted
 // directory to use.
-func InterpretLabelForWorkspaceLocation(root string, target string) (buildFile string, pkg string, rule string) {
+func InterpretLabelForWorkspaceLocation(root, target string) (buildFile, repo, pkg, rule string) {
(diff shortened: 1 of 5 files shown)


edit/edit.go
```diff
@@ -41,17 +41,17 @@ var (
 // InterpretLabelForWorkspaceLocation returns the name of the BUILD file to
 // edit, the full package name, and the rule. It takes a workspace-rooted
 // directory to use.
-func InterpretLabelForWorkspaceLocation(root string, target string) (buildFile string, pkg string, rule string) {
+func InterpretLabelForWorkspaceLocation(root string, target string) (buildFile string, repo string, pkg string, rule string) {
```
pmbethe09, line 44: nit: while you are here can group inputs, outputs

(root, target string) (buildFile, repo, pkg, rule string)

pmbethe09 (approved):
LGTM

edit/edit.go
brown: I updated the change.  Thanks for reviewing.

PR #1152, opened 2023-03-30
Add use_repo_add and use_repo_remove commands

### LINKED ISSUE #1149: Proposal: Add command that manages `use_repo` calls

fmeum:
These commands can be used to programmatically update `use_repo` calls in MODULE.bazel files. In the future, Bazel will emit buildozer fixups relying on this command when it detects that such calls are out-of-date.

Fixes #1149 
Work towards https://github.com/bazelbuild/bazel/issues/17908


fmeum:
@Wyverald Could you review for the Bzlmod aspects?

edit/bzlmod/bzlmod.go
```diff
@@ -0,0 +1,320 @@
...
+	"strings"
+
+	"github.com/bazelbuild/buildtools/build"
+	"github.com/bazelbuild/buildtools/labels"
+)
+
+// Proxies returns the names of extension proxies (i.e. the names of variables to which the result
+// of a use_extension call is assigned) for the given extension with the given value of the
+// dev_dependency attribute.
+func Proxies(f *build.File, extBzlFile labels.Label, extName string, dev bool) []string {
```
Wyverald, line 31: shouldn't this return just one string? especially given https://github.com/bazelbuild/bazel/pull/17920

fmeum: Clarified offline that:
```
ext1 = use_extension("//:foo.bzl", "foo")
ext1.tag()
ext2 = use_extension("//:foo.bzl", "foo")
ext2.tag()
```
is supported as Bazel returns the same proxy object in both cases.

fmeum:
This is soft-blocked on https://github.com/bazelbuild/buildtools/pull/1150 as otherwise the `//MODULE.bazel:__pkg__` target syntax is broken.

fmeum:
@Wyverald I updated the signature and name of the label normalization function and added a reference to https://github.com/bazelbuild/bazel/blob/dd822392db96bb7bccdb673414a20c4b91e3dbc1/src/main/java/com/google/devtools/build/lib/bazel/bzlmod/ModuleFileGlobals.java#L416.

fmeum:
@vladmos Are you available to review this PR?

vladmos:
Sorry, I was conducting a large scale change because of some incompatible changes of formatting rules and temporarily put all other PRs on hold. LGTM

vladmos (approved):

Issue #1098, opened 2022-11-08
released versions on proxy.golang.org`


goddenrich:
I noticed the tool has been released as v5.1.0 https://github.com/bazelbuild/buildtools/releases/tag/5.1.0 but v0.0.0-xxx is the latest release on https://pkg.go.dev/github.com/bazelbuild/buildtools@v0.0.0-20221004120235-7186f635531b. Are you planning to release the versioned releases here and so they appear at eg. https://proxy.golang.org/github.com/bazelbuild/buildtools/@v/v5.1.0.info?

tyler-french:
Still noticing this: https://pkg.go.dev/github.com/bazelbuild/buildtools?tab=versions
This may be due to the missing `v` prefix from the `tags`.

fmeum:
@vladmos

tyler-french:
It would be great to adjust this to make it easier to update: 
```
go install github.com/bazelbuild/buildtools@v6.0.1
```
This is helpful when we build binaries from source locally for security/environment reasons.

vladmos:
How do these versions get released on proxy.golang.org? I only push releases here on Github and to NPM.

fmeum:
@vladmos They are picked up automatically as new tags are added and requested by users. That is also why the format of the tag (`v1.2.3` instead of `1.2.3`) is important.

vladmos:
I've just released a new technical version tagged `v6.1.2`. Should it eventually appear on proxy.golang.org or does anything need to be done manually?

fmeum:
@vladmos It should, but I missed an important point: Go modules tend to not increase their major semver version that often and when they do, their import path needs to be modified as well (in this case, to include the `/v6` suffix). The "version with Bazel" approach unfortunately appears to be at odds with Go's module versioning scheme. 

Assuming backwards compatibility is ensures, would it be possible to move the Bazel version into the minor version?

tyler-french:
@fmeum should we use the "incompatible" tag since the major version is >2?

https://github.com/golang/go/wiki/Modules#can-a-module-consume-a-v2-package-that-has-not-opted-into-modules-what-does-incompatible-mean

fmeum:
@tyler-french I have never used it myself, how would this look like?

I just noticed that Bazel CI doesn't seem to be prepared to handle the semantic versioning format: https://buildkite.com/bazel/stardoc/builds/866#0187c31b-4c54-45ac-9ea2-3513e38731ec

[...]

PR #1162, opened 2023-04-27
Prepare release scripts for the new tag format

vladmos:
Versions in the future will be tagged with `v1.2.3` instead of `1.2.3`, the release scripts should handle them correctly (see #1098).

### DIFF
--- a/status.py
+++ b/status.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+ #!/usr/bin/env python
 
 from __future__ import print_function, unicode_literals
 from subprocess import Popen, PIPE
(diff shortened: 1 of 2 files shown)


status.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+ #!/usr/bin/env python
```
meisterT, line 1: nit: did you add a space here?

meisterT (approved):

status.py
vladmos: It added itself

meisterT (approved):

PR #1164, opened 2023-04-28
Add the ability to escape `|` in command files to Buildozer.

Zilulil:
`|` in command files can now be escaped like `\|` to prevent Buildozer from splitting the command on them. This allows the use of some more complex regex in `substitute` as well as handling targets and commands that need to contain `|` for other reasons.

As a result of this change double null bytes (`\x00\x00`) is no longer legal in a command string however.


vladmos (approved):
Thanks!

PR #1169, opened 2023-05-25
unused_deps: Add exported dependencies before removing

vj-sc:
A dependency that is marked for removal might contain [exports](https://bazel.build/reference/be/java#java_library.exports). Since exports are treated as if the parent explicitly depended on them, removal of such a dependency can cause issues since the exported dependencies will no longer be available to the parent. 
This PR proposes to fix this by first adding the exports explicitly to the parent's deps before removing the dependency.

### DIFF
--- a/unused_deps/unused_deps.go
+++ b/unused_deps/unused_deps.go
@@ -291,6 +291,8 @@ func printCommands(label string, deps map[string]bool) (anyCommandPrinted bool) 
 				if hasRuntimeComment(str) {
 					fmt.Printf("buildozer 'move deps runtime_deps %s' %s\n", str.Value, label)
 				} else {
+					// add dep's exported dependencies to label before removing dep
+					fmt.Printf("buildozer \"add deps $(%s query 'labels(exports, %s)' | tr '\\n' ' ')\" %s\n", *buildTool, str.Value, label)
 					fmt.Printf("buildozer 'remove deps %s' %s\n", str.Value, label)
 				}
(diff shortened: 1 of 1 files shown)


unused_deps/unused_deps.go
```diff
@@ -291,6 +291,8 @@ func printCommands(label string, deps map[string]bool) (anyCommandPrinted bool)
 				if hasRuntimeComment(str) {
 					fmt.Printf("buildozer 'move deps runtime_deps %s' %s\n", str.Value, label)
 				} else {
+					// add dep's exported dependencies to label before removing dep
+					fmt.Printf("buildozer \"add deps $(%s query 'labels(exports, %s)' | tr '\\n' ' ')\" %s\n", *buildTool, str.Value, label)
```
vj-sc, line 295: I tried to locate a utility function for obtaining a label's attribute but was unable to find one, so I used a bazel query invocation here. It'd be better to avoid the query. I'd appreciate any pointers on this in case I missed something.

vladmos:
Do I understand correctly that you're adding deps unconditionally even if it doesn't contain exports? Or will the command be no-op if there are no exports?

0vj00:
> Do I understand correctly that you're adding deps unconditionally even if it doesn't contain exports? Or will the command be no-op if there are no exports?

In the case where there are no exports, the command becomes `buildozer "add deps " <label>` and since there are no arguments to the `add deps` part, buildozer doesn't modify the label.

PR #1173, opened 2023-06-14
Also generate `import`-style targets with Gazelle

fmeum:
Since the repo uses checked-in build files, it should support both the `import` and `go_default_library` naming conventions supported by Gazelle to ensure that it works with all possible values of the `gazelle:go_naming_convention_external` directive.

This is achieved by using `gazelle:go_naming_convention import_alias`.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -21,21 +21,22 @@ test_suite(
     name = "tests",
     tests = [
         "//api_proto:api.gen.pb.go_checkshtest",
-        "//build:go_default_test",
+        "//build:build_test",
         "//build_proto:build.gen.pb.go_checkshtest",
         "//buildifier:buildifier_integration_test",
         "//deps_proto:deps.gen.pb.go_checkshtest",
-        "//edit:go_default_test",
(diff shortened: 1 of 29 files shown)


fmeum:
CC @cgrindel

vladmos:
Former 'go_default_test` rules don't have aliases, is that expected? Is that not part of the convention anymore?

fmeum:
I'm not entirely sure whether this changed at some point, but as far as I understood the convention the aliases exist just to ensure references from other repos to this repo aren't broken. Other repos would not reference tests, so aliases for them shouldn't be needed.

fmeum:
@vladmos Friendly ping on this, it would definitely make downstream users' lives easier. :⁠-⁠)

vladmos (approved):
Thanks!

vladmos:
@fmeum Do you need a new release with these changes or just having them in the repo enough for now?

fmeum:
@vladmos This particular change is fine without a release - in fact, Go doesn't see the releases since their major versions don't match the import path.

PR #1176, opened 2023-07-27
Warning against overriding some global attributes in rules.

aiuto:
Add lint warnings against naming attributes 'licenses', 'applicable_licenses', or 'package_metadata'.  Using any of those can cause unexpected behaviors when generating SBOMs.


aiuto:
cc: @teeler  This would have caught the problem from yesterday.

aiuto:
Friendly ping

aiuto:
Friendly ping

aiuto:
Another week, another ping.

laurentlb (approved):

vladmos:
@aiuto Sorry for the delay, the PR was lost in my backlog.

@laurentlb it will be enabled internally by default (however that can be overridden). I'd keep them enabled unless there's a good reason for using these attributes internally.

aiuto:
We want this check enabled internally. That was the motivation. A rule declaration with an attribute named `licenses` was causes SBOM generation to break for targets which transitively used that rule.

PR #1177, opened 2023-07-28
Fix `use_repo_add` with no existing `use_repo`

fmeum:
`use_repo_add` now inserts a `use_repo` call for the correct proxy when none exists. Previously, it added such a call for the last proxy in the module file, whatever that might be.

Originally observed in https://bazelbuild.slack.com/archives/CDBP88Z0D/p1690549451155769?thread_ts=1690507292.465539&cid=CDBP88Z0D


fmeum:
CC @Wyverald

fmeum:
@vladmos Friendly ping

Issue #1006, opened 2021-09-16
load-on-top should put the load at the right location


laurentlb:
When a `load` statement is put to the top with `buildifier --lint=fix --warnings=load-on-top`, the statement is added to the end of the loads. This means the statement is not placed correctly according to `--warnings=out-of-order-load`.

vladmos:
Should `load-on-top` also sort the existing load statements that are already on top if they are not sorted? Should it do anything if all load statements are already on top but not sorted?

laurentlb:
A simple solution is to assume that the statements are sorted and try to find the correct place to insert. This guarantees we're not regressing.

I don't know if we should sort the existing statements. It increases the diff, but maybe it's okay.

My end goal is to have out-of-order-load on by default in the formatter, but this will require updating lots of existing files. I think it would improve the user experience (they don't have to think where to put a load), and it's done by formatters in other languages.

PR #1183, opened 2023-08-21
Enable out-of-order-load by default

keith:
I couldn't find any info to why this one wasn't enabled by default already, it seems like a nice organizational rule that can help reduce conflicts when adding and removing loads in high traffic bzl files

### DIFF
--- a/WARNINGS.md
+++ b/WARNINGS.md
@@ -765,7 +765,6 @@ The statement has no effect. Consider removing it or storing its result in a var
 
   * Category name: `out-of-order-load`
   * Automatic fix: yes
-  * [Disabled by default](buildifier/README.md#linter)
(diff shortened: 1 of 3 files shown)


laurentlb:
+1 I'd love to see this enabled by default (related discussion in #1006).

I'll let @vladmos decide how to move forward with this, as it will update the formatting of many existing files.

vladmos:
I think it makes sense to enable the warning by default even internally (the warnings are non-blocking by design, are only shown for changed lines and there's a fix attached to them).

I'm also considering moving this functionality from the linter to the formatter (meaning that files with load statements in the middle of a file or load statements in the wrong order will be considered not formatted), but it's not that trivial because I'd need to re-format the existing files first.

vladmos (approved):

PR #1185, opened 2023-08-23
Enable unsorted-dict-items by default

keith:


vladmos:
I'm not sure if this PR is safe; there are dicts that have different ordering on purpose, disabling the warning for such cases may cause a lot of noise (but I don't know how to estimate how many false positive such a warning can lead to).

keith:
Do you think this will ever make sense to do?

vladmos:
Hard to say but I remember when this warning was implemented for the first time too many users complained about it being being too noisy, so I added it to non-default warnings (probably even implemented non-default warnings specifically for this one).

What's your rationale for making it default? Do you have a repository for which you'd like it to be applied? If so you can also change the CI configuration (i.e. call buildifier with `--warnings=+unsorted-dict-items`)

keith:
I think in general it's a nice cleanliness win for most cases. And it's one of the few things that's not formatted. I kinda view buildifier like `go fmt` where non-default configuration is so non-standard that if it's not enabled by default it doesn't exist.

PR #1186, opened 2023-08-24
Add `format` command to force unconditional formatting

fmeum:
This allows users of buildozer to format files without also having buildifier installed. This was possible previously by making a trivial edit and reverting it, but a dedicated command is cleaner.

Will be used as part of the upcoming `bazel mod tidy` command.


fmeum:
@vladmos Could you review? We need this for `bazel mod tidy` as it should format unconditionally and we don't also want to require users to install `buildifier` in addition to `buildozer`.

vladmos (approved):

PR #1193, opened 2023-09-02
Add Java options to api.proto

guw:
Put into proper Java package `com.google.devtools.buildozer` and use `BuildozerProtos` as class name


pmbethe09 (commented):
This is a breaking change for anyone already using Java.

PR #1195, opened 2023-09-04
Fix check for WORKSPACE setting for buildifier, allow WORKSPACE.bzlmod

avdv:
- buildifier: Fix check if `WORKSPACE` is non-empty

This fixes the `buildifier_test` always passing the test, since the bash script does not find any files in sandbox mode.

- buildifier: Allow `WORKSPACE.bzlmod`


buildifier/runner.bash.template
```diff
@@ -11,7 +11,7 @@ buildifier_short_path=$(readlink "$BUILDIFIER_SHORT_PATH")
 if [[ ! -z "${TEST_WORKSPACE+x}" && -z "${BUILD_WORKSPACE_DIRECTORY+x}" ]]; then
   FIND_FILE_TYPE="l"
   # If WORKSPACE was provided, then the script is being run under a test in no_sandbox mode
-  if [[ ! -z "${WORKSPACE+x}" ]]; then
+  if [[ ! -z "${WORKSPACE:+x}" ]]; then
```
vladmos, line 14: What does this change do?

avdv: The `:+` expansion also checks whether the variable is non-empty, not only if it is not set.
```bash
function f() {
  if [ ! -z "${WORKSPACE+x}" ]; then
    echo variable is set
  fi

  if [ ! -z "${WORKSPACE:+x}" ]; then
    echo variable is set and not empty
  fi
}

WORKSPACE=""

f

WORKSPACE="other"

f
```
```console
variable is set
variable is set
variable is set and not empty
```

vladmos (approved):
Thanks!

PR #1196, opened 2023-09-05
Return error `MISSING` when `print_comment` can't find attribute.

danielmirchandani:
Right now, `print_comment` is inconsistent with `print`: `print` returns `MISSING` when it can't find an attribute, but `print_comment` puts nothing in the output field, so callers would have to rely on parsing STDERR to know when this happened.

While I'm here, I'm fixing "a break statement without a target is redundant at the end of a case" warnings from my IDE.


vladmos (approved):
Thanks!

PR #1197, opened 2023-09-06
Updated the bazel query syntax to escape for regex

GSarathChandra:
The bazel query generated by the existing code doesn't escape the quotes correctly for the regex.

Current generated bazel query = 
`bazel query --tool_tag=unused_deps --keep_going --color=yes --curses=yes kind((kt|java|android)_*, target:label)`

When this is run directly in a terminal, it is failing with 
```bash: syntax error near unexpected token `(' ```

The bazel query generated after the fix = 
`bazel query --tool_tag=unused_deps --keep_going --color=yes --curses=yes 'kind("(kt|java|android)_*", target:label)'`

And this is working as expected.
Updated the quotations syntax following the example at https://bazel.build/query/guide#tests-beneath-foo


GSarathChandra:
After some testing, I've identified that this change is not necessary and that the tool is working as intended. Hence, closing the PR.

PR #1200, opened 2023-09-28
Make package on top apply fix

AttilaTheFun:
Currently the package-on-top buildifier lint rule only shows a warning but is unable to offer a fix.
This PR addresses this issue by essentially cloning the behavior of the load-on-top rule, but placing the package statement after the load statements (as well as any docstrings or comments).
I updated the test to use checkFindingsAndFix as well and it passes.


warn/warn_cosmetic.go
```diff
@@ -102,30 +102,75 @@ func sameOriginLoadWarning(f *build.File) []*LinterFinding {
...
-	seenRule := false
-	for _, stmt := range f.Stmt {
+	if f.Type == build.TypeWorkspace || f.Type == build.TypeModule {
+		// Not applicable to WORKSPACE or MODULE files
+		return nil
+	}
+
+	// Find the misplaced load statements
+	misplacedPackages := make(map[int]*build.CallExpr)
+	firstStmtIndex := -1 // index of the first seen non string, comment or lead statement
```
vladmos, line 117: typo: load

warn/warn_cosmetic.go
```diff
@@ -102,30 +102,75 @@ func sameOriginLoadWarning(f *build.File) []*LinterFinding {
...
+			continue
+		}
+		misplacedPackages[i] = rule.Call
+	}
+
+	// Calculate a fix
+	if firstStmtIndex == -1 {
+		firstStmtIndex = 0
+	}
+	offset := len(misplacedPackages)
```
vladmos, line 143: Nit: add `if offset == 0 { return nil }` for better readability?

vladmos (approved):

AttilaTheFun:
Thanks @vladmos ! I incorporated your suggestions and pushed the commit. I don't think I have permission to merge it, so if you could do so whenever you have a chance that would be great.

vladmos:
I'll merge the pull request in a few days, I want to do it separately from other pull requests to make sure it doesn't break any internal projects.

Issue #1201, opened 2023-09-28
buildifier: uninitialized check fails with list unpack assignment if the same variable is used later


Gandem:
With the following example in `test.bzl`:

```py
def foo():
  [x, y] = [1, 2]
  x = 3
  print(x)
```

Linting this file returns the following:
```bash
$ buildifier -lint=warn -warnings uninitialized test.bzl
test.bzl:2: uninitialized: Variable "x" may not have been initialized. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#uninitialized)
```

However, the code above is supposed to be valid per the starlark spec on assignments (https://github.com/bazelbuild/starlark/blob/master/spec.md#assignments), so the linter should not be returning any error.

I've submitted a pull request with a tentative fix: https://github.com/bazelbuild/buildtools/pull/1202

Please let me know if there's anything that needs to be added/modified in this change, thanks! 🙏

PR #1202, opened 2023-09-28
buildifier: fix uninitialised check when using list unpack assignment

Gandem:
Fixes https://github.com/bazelbuild/buildtools/issues/1201

As the test added in this commit shows, prior to this commit, the uninitialized check would incorrectly fail when using list unpack assignment, if the same variable is reused later.

This modifies `CollectLValues`, so that it supports list unpack assignments (in addition to tuple, and normal assignments).

Please let me know if there's anything that needs to be added/modified in this change, thanks! 🙏


vladmos (approved):
Thanks!

vladmos:
Could you please update the branch? Usually I can do this myself just before merging, but now I likely don't have permissions to do so.

Gandem:
@vladmos Done, thanks 🙇 !

PR #1207, opened 2023-10-10
chore: update to latest rules_nodejs

alexeagle:
Simplify by writing the buildozer program in JavaScript rather than TypeScript so we don't need any devDependencies


vladmos:
There are build failures for the PR, could you please take a look at them?

alexeagle:
Green now. I also added an optional pre-commit setup so CI doesn't yell at me about buildifier. (It doesn't even print which files had format/lint warnings...)

vladmos:
Thanks!

Do I understand correctly that the precommit hook only those who explicitly installed `pre-commit` on their development machines?

vladmos (approved):

PR #1209, opened 2023-10-12
Add AttrMap() function for getting rule attributes

aqua4:


oreflow:
Closing this due to being stale for a long time.
If reopening It seems like this adds a method which is not used and not tested, this would need some further motivation / clarification for why it is necessary

PR #1215, opened 2023-11-08
No format for tabular data

systemlogic:
Not in all cases formatting of build data / code is required. Code can be written in a way which looks beautiful and easy to spot problem when maintained in tabular format maintain certain spacing between columns. Current implementation of buildfier results loosing well formatted code( manually maintained) which has been supported so long before porting to bazel.


oreflow:
Closing this PR since it's been stale for a long period and still in draft state. Feel free to reopen if it's still needed

PR #1216, opened 2023-11-08
No format for tabular data

systemlogic:
Not in all cases formatting of build data / code is required. Code can be written in a way which looks beautiful and easy to spot problem when maintained in tabular format using certain spacing between columns. Current implementation of buildfier results in loosing well formatted code( manually maintained) which has been needed in tabular code.


aiuto:
This seems to change the whitespace of nearly every line. Did your editor save tabs as spaces?

systemlogic:
> This seems to change the whitespace of nearly every line. Did your editor save tabs as spaces?

No, such settings in my vim compiler, which save tabs as spaces. 

However our own internal parser, works with precision of spaces and all such tabular code need to be separated with spaces.

systemlogic:
@aiuto ping for review request.

mzeren-vmw:
FWIW, we are using an internal version of #985. That patch needs to be rebased and reworked. Importantly for us it supports sorting by a particular column.

tonyaiuto:
Is there anyone in the bazelbuild org willing to move this forward.
cc/ @meteorcloudy

meteorcloudy:
@vladmos @oreflow are now owning this repo, so they'll need to decide

systemlogic:
@vladmos @oreflow  Any thought ?

Issue #1227, opened 2024-01-02
new_load to= aliases broken


mering:
According to the [docs](https://github.com/bazelbuild/buildtools/blob/master/buildozer/README.md#edit-commands) there should be a `new_load <path> <[to=]from(s)>` statement. Unfortunately, the `to=``part currently seems to be ignored (at least when another load with the `from` name is already available.

I created a minimum reproduction example at #1228.

mering:
Can't reproduce.

PR #1228, opened 2024-01-02
Add test case for demonstration purposes

mering:
This test case should work according to the docs. See #1227.


mering:
Can't reproduce.

PR #1222, opened 2023-12-13
Add missing repo boundary files

fmeum:

PR #1236, opened 2024-01-25
MODULE.bazel should be a marker of the root of the repo

charlesoconor:
When converting over to bzlmod and not having a `WORKSPACE` file in the root of the repo anymore

```
buildifier 'print label' //python/...:%py_library
```

Will return only the label name. e.g `//:some_target` instead of `//nested:some_target` or where ever it should be relative to root.

### DIFF
--- a/wspace/workspace.go
+++ b/wspace/workspace.go
@@ -55,6 +55,7 @@ func isExecutable(fi os.FileInfo) bool {
 var repoRootFiles = map[string]func(os.FileInfo) bool{
 	workspaceFile:            isFile,
 	workspaceFile + ".bazel": isFile,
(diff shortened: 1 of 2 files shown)


charlesoconor:
It looks like I mucked up the integration test with this since there is a `test_dir/MODULE.bazel`, so it's changing the labels to be relative to that. e.g `//to_fix.bzl` instead of `//test_dir/to_fix.bzl`. 

I might be missing an understanding of how `MODULE.bazel` files would work since just starting my migration over to bzlmod.

charlesoconor:
Looks like this is the same idea. https://github.com/bazelbuild/buildtools/pull/1222

PR #1062, opened 2022-04-27
Add ReadFile variable in buildtools/file API.

mohdanishaikh:

PR #1238, opened 2024-02-06
Add WriteFile and OpenReadFile variables in buildtools/file API.

openvcdiff:
These are analogous to the ReadFile variable added in #1062.


vladmos (approved):
Thanks!

PR #1244, opened 2024-02-15
Stricter formatting of load statements

vladmos:
Remove the following linter warnings/fixes and make them part of the default formatting process:

  * `load-on-top` (moves all load statements to the top of a file)
  * `same-origin-load` (compresses load statements from the same source into one load statement)
  * `out-of-order-load` (sorts load statement lexicographically by their source but global imports have priority)

Formatted starlark files without active warnings of these categories will remain formatted. Files with the warnings will became non-formatted according to the new formatting rules.

For backward compatibility the following comments are still taken into account:

    # buildifier: disable=load-on-top
    # buildifier: disable=same-origin-load
    # buildifier: disable=out-of-order-load


WARNINGS.md
```diff
@@ -765,8 +768,11 @@ The statement has no effect. Consider removing it or storing its result in a var
 
   * Category name: `out-of-order-load`
   * Automatic fix: yes
+  * Not supported by the latest version of Buildifier
   * [Suppress the warning](#suppress): `# buildifier: disable=out-of-order-load`
 
+Obsolete; the warning has been implemented in the formatter and the fix is now automatically applied to all files.
```
lukesandberg, line 774: should this say 'unless suppressed'?

build/rewrite.go
```diff
@@ -911,6 +938,245 @@ func fixMultilinePlus(f *File, _ *Rewriter) {
...
+
+	chunks1 := strings.Split(path1, "/")
+	chunks2 := strings.Split(path2, "/")
+
+	for i, chunk1 := range chunks1 {
+		if i >= len(chunks2) {
+			return false
+		}
+		chunk2 := chunks2[i]
+		// Compare case-insensitively
```
lukesandberg, line 1066: are bzl paths not case sensitive generally?  is `build` as good as `BUILD`

build/rewrite.go
```diff
@@ -911,6 +938,245 @@ func fixMultilinePlus(f *File, _ *Rewriter) {
...
+		chunk1Lower := strings.ToLower(chunk1)
+		chunk2Lower := strings.ToLower(chunk2)
+		if chunk1Lower != chunk2Lower {
```
lukesandberg, line 1069: would it be more efficient (fewer copies) to call `strings.EqualFold` here instead of lower casing the string?

lukesandberg (approved):

build/rewrite.go
vladmos: Depends on the file system. On Linux the paths are usually case sensitive, `build` and `BUILD` can be two separate files in ext3.

build/rewrite.go
vladmos: It could be efficient but I don't only compare for equality here, I need to know which chunk comes first.

WARNINGS.md
vladmos: Done.

lukesandberg (approved):

pmbethe09 (approved):
LGTM

hunshcn:
release a new version?

PR #1245, opened 2024-02-23
chore(bazel): add MODULE.bazel files for bzlmod

mmorel-35:
Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>


fmeum:
Something to be aware of: The release versioning used by this repo is incompatible with how Go modules are versioned. It would be great if this could be resolved before this module is added to the BCR. This could otherwise result in quite a bit of pain for Go projects using the repo as a Go module and also picking it up as a transitive bazel_dep.

MODULE.bazel
```diff
@@ -1,2 +1,49 @@
...
+)
+
+bazel_dep(name = "aspect_bazel_lib", version = "1.40.0")
+bazel_dep(name = "aspect_rules_js", version = "1.38.0")
+bazel_dep(name = "bazel_features", version = "1.2.0")
+bazel_dep(name = "bazel_skylib", version = "1.5.0")
+bazel_dep(name = "gazelle", version = "0.34.0", repo_name = "bazel_gazelle")
+bazel_dep(name = "protobuf", version = "23.1", repo_name = "com_google_protobuf")
+bazel_dep(name = "rules_go", version = "0.45.1", repo_name = "io_bazel_rules_go")
+bazel_dep(name = "rules_nodejs", version = "5.8.2")
```
fmeum, line 13: Could some of these deps be dev dependencies instead?

alexeagle:
> release versioning used by this repo is incompatible with how Go modules are versioned
as one option to consider (and I wonder if we are doing it right), https://github.com/aspect-build/aspect-cli/tags publishes two tags for each release, one that looks like a bazel module version and one that looks like a Go package version (prefixed with `v`)

fmeum:
@vladmos Would that be an option for buildtools?

.gitignore
```diff
@@ -4,3 +4,4 @@
 .ijwb
 *.iml
 /.bazelrc.user
+MODULE.bazel.lock
```
vladmos, line 7: What's this file? Is it being created by your editor?

keith: bzlmod has a lockfile which afaiui the primary feature is it allows you to build offline. but it is pretty verbose, and i believe there are some cross-platform holes in it today, so it adds some maintenance burden to check in unless you really need it. ignoring it allows you the benefits locally w/o having to deal with updating it

PR #1246, opened 2024-03-01
buildozer: correctly substitute custom string attrs

thumphries-stripe:
- Fix a bug in the `substitute` buildozer command
- Add some tests for that command, since none were found

`getAttrValueExpr` runs via the [`TypeOf` map](https://github.com/bazelbuild/buildtools/blob/master/lang/tables.gen.go#L6), and encodes the value as an ident if the type is not known. This works correctly for known attributes, but produces an invalid result for anything not contained in that map.

However, we already know the attribute is a string due to the type assertion on L543, and can construct it directly.


vladmos (approved):
Thanks!

Issue #1249, opened 2024-03-11
Handle multiple glob arguments in 'set' arguments


stagnation:
Following on #100 and #99
> Ideally, we'd find a more general solution.

The special case for single argument glob works fine, but with multiple arguments it again parses the glob as individual files.

```
$ buildozer 'set srcs glob(["*.c"])' //:foobar
fixed /.../buildozer/glob-multiple-args/BUILD.bazel
$ cat BUILD.bazel
cc_library(
    name = "foobar",
    srcs = glob(["*.c"]),
)
$ buildozer 'set srcs glob(["*.c", "*.h"])' //:foobar
fixed /.../buildozer/glob-multiple-args/BUILD.bazel
$ cat BUILD.bazel
cc_library(
    name = "foobar",
    srcs = [
        "\"*.h\"])",
        "glob([\"*.c\",",
    ],
)
```

stagnation:
I'll take a look and see if I can improve the special handling of globs here.

dfreese:
Ran into this when trying to trying to set a `srcs` list attribute to a macro that wraps glob.  Perhaps having a "raw text" mode might be more a more general approach?  Such as:

```
buildozer 'set srcs r"glob(["*.c", "*.h"])"' //:foobar
```
which would mean, "set it to exactly this string."

stagnation:
Yes that sounds like a good idea. I have also been meaning to look into how to set imported special tag values to the tags attribute and found the same issue that buildozer knows that tags are string so it adds the quotes, which I don't want in that case. A raw mode would solve both of those issues.

I know there is also a type-table that could potentially be manipulated to achieve some of this, but I did not find the time to muck around with it. 

But when the developer knows the character sequence to insert it would be good to just allow that and bypass the user-friendly type information that otherwise simplifies writing the commands and shell quoting and all of that. But for maximum control I want the semantic parsing of buildozer with the file insertion control of regular sed.

PR #1250, opened 2024-03-11
buildozer: handle multiple glob tokens in set command

stagnation:
To address #1249. This continues the special case handling for globs in `set X glob(...)` commands from #100. Ideally we should have a higher parsing level to extract the inner arguments from functions so they can be reorganized like regular files are. This does not work for `add X glob(...)` which follows the status quo. At the moment we only use `set` internally but can look at `add` too on the back burner. Thanks for this wonderful tool!

### DIFF
--- a/buildozer/buildozer_test.sh
+++ b/buildozer/buildozer_test.sh
@@ -1961,6 +1961,18 @@ function test_set_glob_srcs() {
 )'
 }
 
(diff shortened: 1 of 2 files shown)


stagnation:
Just found this: https://github.com/bazelbuild/buildtools/pull/572 which also improves globs, higher up in the parser. Is that a better solution to revive and continue with?

PR #1257, opened 2024-03-21
constant-glob validates named include argument

### LINKED ISSUE #1256: constant-glob fails to lint when the first argument is named
constant-glob does not correctly lint constant globs when using a named `include` parameter.

This raises a warning:
```
❯ echo 'glob(["foo.py"])' | buildifier -lint=warn -warnings=+constant-glob -type build -
glob(["foo.py"])
<stdin>:1: constant-glob: Glob pattern "foo.py" has no wildcard ('*'). Constant patterns can be error-prone, move the file outside the glob. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#constant-glob)
```

This fails to raise a warning, even though I'd expect it to:
```
❯ echo 'glob(include = ["foo.py"])' | buildifier -lint=warn -warnings=+constant-glob -type build -
glob(include = ["foo.py"])
```

mark-thm:
Fixes #1256.

Previously: constant-glob failed to warn when using named argument `include` with a constant pattern.

Now: constant-glob warns when using named argument `include` with a constant pattern or when the first unnamed argument uses a constant pattern.

### DIFF
--- a/warn/warn_bazel.go
+++ b/warn/warn_bazel.go
@@ -45,7 +45,7 @@ func constantGlobPatternWarning(patterns *build.ListExpr) []*LinterFinding {
 			message := fmt.Sprintf(
 				`Glob pattern %q has no wildcard ('*'). Constant patterns can be error-prone, move the file outside the glob.`, str.Value)
 			findings = append(findings, makeLinterFinding(expr, message))
-			break // at most one warning per glob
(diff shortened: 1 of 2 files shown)


meteorcloudy:
@vladmos Can you take a look

vladmos (approved):
Thanks!

PR #1262, opened 2024-04-07
Do not declare bash output on windows

limdor:
Calling declare_file on windows even if later is overwritten is a problem in windows genereting the following error
```
The following files have no generating action:
buildifier.bash
```

### DIFF
--- a/buildifier/internal/factory.bzl
+++ b/buildifier/internal/factory.bzl
@@ -168,8 +168,6 @@ def buildifier_impl_factory(ctx, test_rule = False):
             fail("Cannot use 'no_sandbox' without a 'workspace'")
         workspace = ctx.file.workspace.path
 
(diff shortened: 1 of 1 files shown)


buildifier/internal/factory.bzl
```diff
@@ -168,8 +168,6 @@ def buildifier_impl_factory(ctx, test_rule = False):
             fail("Cannot use 'no_sandbox' without a 'workspace'")
         workspace = ctx.file.workspace.path
 
-    out_file = ctx.actions.declare_file(ctx.label.name + ".bash")
```
limdor, line 171: Note for the reviewer, this is already present in https://github.com/bazelbuild/buildtools/pull/1262/files#diff-c5e2ba5db338e88cf7c850afd25c044dfa97198fefa4b296370c9286ab81d415L184

vladmos (approved):
Thanks!

PR #1266, opened 2024-04-22
Add support for `*.MODULE.bazel` files

fmeum:
After https://github.com/bazelbuild/bazel/pull/22075, these are the files that can be `include()`ed into `MODULE.bazel` files and should be formatted and edited just like them.

Work towards bazelbuild/bazel#22063


fmeum:
CC @Wyverald

Wyverald (approved):

PR #1270, opened 2024-05-14
Add tables to runfiles

stevebarrau:
Add the tables label to the runfiles.

PR #1271, opened 2024-05-22
Ensure that `buildifier`'s `add_tables` attribute is in the runfiles

shs96c:
Without this, if the `add_tables` value is the output of another rule, it won't be available to `buildifier`.


stevebarrau:
@shs96c is this a dupe of https://github.com/bazelbuild/buildtools/pull/1270 ?

shs96c:
You're right! Closing :)

PR #1284, opened 2024-08-07
Replace log.Fatalf with log.Exitf to avoid spamming stderr.

paulherman:
The set of warnings is something that is changing over time. As scripts tend to hardcode these, they get out of date and we hit this unexpected warning code path. There's little benefit of having a stack trace from this issue. Thus, it's more convenient to just print the error itself and nothing else.


lysliu (approved):
LGTM

vladmos (approved):

Issue #1287, opened 2024-08-22
Support MODULE.bazel in `FindRepoBuildFiles()`


mering:
Currently, it only parses `WORKSPACE`: https://github.com/bazelbuild/buildtools/blob/19c62d9fc1d29ca99c3faa08f78cd90b36f49908/wspace/workspace.go#L123

This is used for example by `edit.InterpretLabel()`:
https://github.com/bazelbuild/buildtools/blob/19c62d9fc1d29ca99c3faa08f78cd90b36f49908/edit/edit.go#L41-L113

PR #1288, opened 2024-08-22
Return empty string when no BUILD file is available in our workspace

mering:
There are many cases where `BUILD` files are not available locally: Either they are available upstream or in BCR. Consumers of this function often ignore this return value anyways. Setting it to an empty string seems much more plausible then falling through an matching any other unsuitable case.

Related #1287


drigz:
@vladmos FYI, I'm rodrigoq@, the author of the TODO, and while very ashamed that I allowed the code to fall through in this case as it doesn't seem to make any sense, this seems like a solid improvement.

vladmos (approved):
Thanks!

PR #1291, opened 2024-09-08
Create go.yml

Punkin-Anarcrassy:


oreflow:
I'm not sure I understand the reason to add this since we're generally using Bazel to build these. Is there some specific need for building it with the go toolchain separately?

PR #1293, opened 2024-09-12
perf: Avoid recompiling regexes

dzbarsky:
We can compile these once at package scope instead. This is a slight perf improvement on a medium repo.

I removed some unused code while I was here.


DavidZbarsky-at:
Example: 
image

dzbarsky:
@vladmos can you please take a look at this one?

PR #1302, opened 2024-10-17
Enable native-* rules

keith:
With bazel 8.x these should now be loaded externally. It's currently
only a soft requirement, but at this point all users should start
migrating to this.


limdor:
It would be nice if before enabling it again, the Readme of the rules_cc is fixed.
It still says that there is no need to use rules_cc
https://github.com/bazelbuild/rules_cc?tab=readme-ov-file#getting-started
This has already being reported in rules_cc in https://github.com/bazelbuild/rules_cc/issues/273

keith:
@comius can you ptal

keith:
https://github.com/bazelbuild/rules_cc/pull/300

keith:
looks like an infra error with macOS CI here

keith:
@vladmos ptal

comius (approved):

vladmos (approved):
Thanks!

PR #1303, opened 2024-10-17
Add rules_shell fixing

keith:
Currently these are split into 3 rules because 1 rule cannot add 3 load
statements and this repo doesn't have a defs.bzl that exposes them all


keith:
i don't feel strongly about the defs.bzl part since realistically these rules shouldn't be disabled, but happy to try to go that route and add that upstream to unify this into 1 rule

keith:
@vladmos @comius another one for you, CI issue is infra

warn/warn_bazel_api.go
```diff
@@ -798,6 +798,27 @@ func nativeProtoSymbolsWarning(symbol string, bzlfile string) func(f *build.File
 	}
 }
 
+func nativeShBinaryWarning(f *build.File, fileReader *FileReader) []*LinterFinding {
+	if f.Type != build.TypeBzl && f.Type != build.TypeBuild {
+		return nil
+	}
+	return NotLoadedFunctionUsageCheck(f, fileReader, []string{"sh_binary"}, "@rules_shell//shell:sh_binary.bzl")
```
comius, line 805: Can you move hard coded  "@rules_shell//shell:" into tables. See other warnings.

This makes it a bit easier to import, because we're overriding those labels internally.

comius (changes requested):
Thanks for this change!

comius:
cc @c-mita

warn/warn_bazel_api.go
keith: ah nice, i skipped that not knowing you were replacing that internally, i should have known! Updated!

vladmos (approved):

PR #1308, opened 2024-11-18
Collapse empty call expressions

snady:
Use compact mode for call expressions if there are no parameters or line comments.

Running buildifier on
```
func(
)
```
doesn't format it because of `ForceMultiline`, but one could reasonably expect it to collapse into 
```
func()
```
This comes up sometimes when removing attrs with buildozer -- if `ForceMultiline=true` but the rule/macro no longer has any attrs, the result will look like the first example e.g
```
some_macro(
  attr_to_remove=[
    ...
  ]
)
```

```
some_macro(
)
```


vladmos (approved):
Thanks!

vladmos:
Hi, please resolve the conflicts, I'll merge the PR

vladmos (approved):
Thanks!

PR #1309, opened 2024-11-21
Fix and improve float formatting

fmeum:
`12e+34` is no longer broken into `12e + 34`.

Also adds a rewrite pass that turns e.g. `.23E+021` into `0.23e21`: 
* adds a leading 0 to floats
* removes leading `+`s and `0`s from the exponent
* lowercases `E`


build/testdata/003.golden
```diff
@@ -5,4 +5,17 @@ numbers = [
...
+    0.0,
+    -0.0,
+    1.0,
+    -1.0,
+    +1.0,
+    1e6,
+    -1e6,
+    -1.23e-45,
+    3.539537889086625e+24,
+    3.539537889086625E+24,
```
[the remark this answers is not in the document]
laurentlb: Yes, I think it makes sense to standardize this. Changing to lowercase `e` can affect existing users (so it's worth mentioning in the release notes), but it should be easy to rollout.

fmeum:
I added a rewrite pass that adds a leading `0`, removes leading `0`s and `+`s from the exponent and lowercases the `E`.

build/testdata/003.golden
```diff
@@ -3,6 +3,24 @@ numbers = [
...
+    1e6,
+    -1e6,
+    -1.23e-45,
+    3.539537889086625e24,
+    3.539537889086625e24,
+    3.539537889086625e24000,
+    3.539537889086625e24000,
+    3539537889086624823140625,
+    0x123,
+    0xE45,
```
vladmos, line 24: Shouldn't this `E` also be lowercased?

vladmos: Or this is invalid and just left as is?

fmeum: This is not a float, but a hex number. We could discuss whether hex literals should be canonicalized in some way, but I would prefer to handle that in a separate PR.

vladmos (approved):

PR #1310, opened 2024-11-27
Require proto rules to be loaded by buildifier

comius:
Because the rules are moved into the repository the loads become mandatory with Bazel@HEAD.

Add new warning for each rule and symbol that needs to be loaded. Buildifier doesn't handle well loading from different bzl files in a single warning.

Enable all proto rules warning by default.


comius:
Hey @meteorcloudy, any idea why build is failing on protobuf on this PR, but not on daily builds?

buildifier/config/config_test.go
```diff
@@ -260,9 +267,16 @@ func TestValidate(t *testing.T) {
 			"native-android",
 			"native-build",
 			"native-cc",
+			"native-cc-proto",
 			"native-java",
+			native-java-lite-proto",
```
meteorcloudy, line 272: Missing a quote here.

meteorcloudy (changes requested):

buildifier/config/config_test.go
meteorcloudy:
```
gentestmain: ParseFile("buildifier/config/config_test.go"): buildifier/config/config_test.go:272:26: string literal not terminated (and 10 more errors)
```

comius: Ahhh thanks!! I was looking at protobuf warning and didn't see the error :)

meteorcloudy (approved):

meteorcloudy:
@vladmos Please take a final look to merge it

vladmos (approved):

PR #1321, opened 2025-01-10
Use apparent repo names when fixing loads

fmeum:
Whether a load for e.g. proto rules should be from `@com_google_protobuf`, `@protobuf` or a fully custom name depends on the `repo_name` on the `bazel_dep` for the `protobuf` module in the root module's module file.

This PR scans `MODULE.bazel` and included `*.MODULE.bazel` files for `bazel_dep`s and uses the derived mapping for the suggested loads.


fmeum:
@vladmos @Wyverald Would be great to get this out into an 8.0.1, otherwise many users will pick up load fixes that don't work for them.

fmeum:
FYI @comius

PR #1329, opened 2025-01-30
Fix load placement for files with docstring and copyright header

fmeum:
Also add missing setup calls for other tests.


fmeum:
@vladmos Could you create another patch release after the merge?

vladmos (approved):

PR #1341, opened 2025-03-06
Add change command to buildozer to change rule kind in place

dfreese:
I'm actively working on a migration where I need to change the kind of a rule from one to another in place.  Prior to this step, I used move, rename, and set to make the attributes of the two rules look comparable.  I now need to switch to the new rule. Combing change with multiple other commands in a single invocation allows for doing reasonably complex changes in place.

I wasn't seeing a command to perform this operation, though I could have missed that.  "rename" is already taken for attributes.  "change" seemed the clearest after that. Open to other suggestions.

I've provided some basic test cases, and am happy to add more, but it didn't seem particulary worthwhile, given the simplicity of the command implementation.

### DIFF
--- a/buildozer/README.md
+++ b/buildozer/README.md
@@ -108,6 +108,8 @@ Buildozer supports the following commands(`'command args'`):
   * `new <rule_kind> <rule_name> [(before|after) <relative_rule_name>]`: Add a
     new rule at the end of the BUILD file (before/after `<relative_rule>`). The
     identifier `__pkg__` can be used to position rules relative to package().
+  * `change <rule_kind>`: Change a rule or rules to the new rule kind in place.
+    All attributes on the current rule are preserved.
   * `print <attr(s)>`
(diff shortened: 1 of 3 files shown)


oreflow:
Hi David,

Thank you for suggesting this, however, if I understand your use case correctly this is already supported through

`set kind <value>`: Set the target type to value.

Quickly ran your added test case, of

```
foo(
    name = "foo",
    flag = True,
    # Comment on attribute.
    label_list = [
        "//:a",
        "//:b",
        "//:c",
    ],
    value = 1,
)
```
With the following command

`$ buildozer 'set kind bar' :foo`

Results in

```
bar(
    name = "foo",
    flag = True,
    # Comment on attribute.
    label_list = [
        "//:a",
        "//:b",
        "//:c",
    ],
    value = 1,
)
```

Please let me know if this solves your use case, or if the proposed `change` does anything differently

dfreese:
Missed the `set kind` docs.  🤦  Confirmed `set kind` works in my use case.  Thanks.

Issue #623, opened 2019-04-30
Buildifier should respect .bazelignore


HackAttack:
When I run `buildifier -r`, I think it should skip the same directories that Bazel itself skips.

aherrmann:
Same goes for buildozer, e.g. `buildozer 'new foo bar' //...:__pkg__`.

amartani:
I opened a PR implementing it for buildozer - https://github.com/bazelbuild/buildtools/pull/1343

If that is accepted, I can also take a shot at doing the same for buildifier.

dmivankov:
For buildifier it also goes where it definitely shouldn't, like
```
./.git/logs/refs/heads/funny_branch_name.bzl # reformat
```
for repo having non-compacted branch named `funny_branch_name.bzl`

icaromag:
Hey @amartani thanks for that! any plans to have it implemented in the buildifier?

amartani:
I still want to do it, but I haven't had time yet. At $dayjob, we worked around it by using `buildozer format //...:__pkg__` instead of `buildifier` 🙃

PR #1343, opened 2025-03-07
buildozer: Respect .bazelignore

amartani:
Currently, when expanding `/...` targets, buildozer does not respect `.bazelignore` files, which can be surprising. This implements reading `.bazelignore` if it exists and ignoring any paths from it to match bazel's behavior.

Note that we only do prefix comparison, with no support for globs, as that is bazel's behavior. We also ignore any absolute paths, which, in bazel itself, results in an error. Reference implementation:

https://github.com/bazelbuild/bazel/blob/8dbfcfae924b014366a0b47cd26632c5076b51a2/src/main/java/com/google/devtools/build/lib/skyframe/IgnoredSubdirectoriesFunction.java#L76-L103

Part of https://github.com/bazelbuild/buildtools/issues/623 (this only applies for buildozer)


amartani:
Hi @vladmos, are you available to review this PR? Thank you!

edit/buildozer.go
```diff
@@ -1321,6 +1329,53 @@ func findBuildFiles(rootDir string) []string {
...
+func shouldIgnorePath(path string, rootDir string, ignoredPrefixes []string) bool {
+	rel, err := filepath.Rel(rootDir, path)
+	if err != nil {
+		return false
+	}
+	// Normalize path separators to forward slashes
+	rel = filepath.ToSlash(rel)
+
+	for _, prefix := range ignoredPrefixes {
+		if strings.HasPrefix(rel, prefix) {
```
vladmos, line 1372: Will `foobar` be ignored if `foo` is among ignored prefixes?

amartani: Ah yes, good call. I changed the comparison to only accept exact match or match with `prefix+"/"`, and changed `getIgnoredPaths` to `Clean` the paths so that `/` suffixes in .bazelignore are stripped from it. 111bc6e962b5115c91e5d9a0001910bdc423d6eb

vladmos:
Could you please add a boolean option to Opts (e.g. `UseBazelignore`), with a default value of true, and propagate it from `appendCommands` to `targetExpressionToBuildFiles`? And, if it's set to false, use an empty slice instead of calling `getIgnoredPrefixes`.

This is a useful feature for the open-source version of buildozer, however, not required by the internal one and may add some unnecessary latency for file operations there, it'd be nice if it can be configurable.

amartani:
> Could you please add a boolean option to Opts (e.g. `UseBazelignore`), with a default value of true, and propagate it from `appendCommands` to `targetExpressionToBuildFiles`? And, if it's set to false, use an empty slice instead of calling `getIgnoredPrefixes`.
> 
> This is a useful feature for the open-source version of buildozer, however, not required by the internal one and may add some unnecessary latency for file operations there, it'd be nice if it can be configurable.

@vladmos PTAL; I named `RespectBazelignore` since it feels clearer to me on how it is used, and it aligns with what I see in other tools (eg. [respect-gitignore on ruff](https://docs.astral.sh/ruff/settings/#respect-gitignore)) but let me know if you prefer to `UseBazelignore`. Also exposed it as a command-line flag and added a shortcut conditional on `shouldIgnorePath` to minimize any impact on the common case of not using `.bazelignore`.

vladmos (approved):
Thanks! I'll merge it within a couple of days, there's currently another chunk of changes being imported at the moment, don't want multiple imports to interfere with each other.

amartani:
@vladmos no particular hurry, but is this unblock to be merged? Thank you!

vladmos (approved):

vladmos:
Sorry for the delay, merging it.

PR #1345, opened 2025-03-21
Keep `include` calls together in `MODULE.bazel` files

fmeum:


vladmos (approved):

fmeum:
How do you obtain `buildifier`? This PR hasn't been included in any release yet, so I don't see how it could format `include`s without empty lines at this point.

psalaberria002:
> How do you obtain `buildifier`? This PR hasn't been included in any release yet, so I don't see how it could format `include`s without empty lines at this point.

We consume it as a go module with a commit reference.

fmeum:
In that case you'll have to wait for a new buildtools release, which can then be used for a new release of the `buildozer` module. Bazel uses that module for `bazel mod tidy`, so when it's out, you just need to bump a `bazel_dep` rather than update Bazel.

honnix:
> In that case you'll have to wait for a new buildtools release, which can then be used for a new release of the `buildozer` module. Bazel uses that module for `bazel mod tidy`, so when it's out, you just need to bump a `bazel_dep` rather than update Bazel.

Which module is supposed to be bumped? We use buildifier and buildozer through rules_multitool and are experiencing the same format conflict issue.

fmeum:
The `buildozer` BCR module is what Bazel uses under the hood. You can pin it with a `single_version_override`.

TroyKomodo:
@fmeum what I am gathering here is if we add 
```
bazel_dep(name = "buildozer", version = "8.2.1")
```

to the `MODULE.bazel` then this will force bazel internally to use `>= 8.2.1` of `buildozer` thus making `bazel mod tidy` include this formatting change?

If so, perhaps this can be documented somewhere outside of this PR comments. for bazel versions prior to some version `x.y.z`

PR #1346, opened 2025-03-26
Build and ship binaries for riscv64

kxxt:
This is needed to bring riscv64 support into rules_rust so a new release would be appreciated :smile:.


vladmos (approved):
Thanks!

PR #1348, opened 2025-04-03
Note that list-append incorrectly warns on select statements

pollendo:
Buildifier advises authors to use the append function to add singletons to variables.

This works when the variable is a list. If the variable is a select statement, this doesn't work. In this case, the only option is to add a singleton list to the select.

### DIFF
--- a/WARNINGS.md
+++ b/WARNINGS.md
@@ -612,6 +612,8 @@ the readability.
   * [Suppress the warning](#suppress): `# buildifier: disable=list-append`
 
 Transforming `x += [expr]` to `x.append(expr)` avoids a list allocation.
+This replacement does not apply to select statements, but the warning might trigger
+anyway. Feel free to suppress using the command above.
 
 --------------------------------------------------------------------------------
 
--- a/warn/docs/warnings.textproto
+++ b/warn/docs/warnings.textproto
@@ -394,7 +394,9 @@ warnings: {
 warnings: {
   name: "list-append"
   header: "Prefer using `.append()` to adding a single element list"
-  description: "Transforming `x += [expr]` to `x.append(expr)` avoids a list allocation."
+  description: "Transforming `x += [expr]` to `x.append(expr)` avoids a list allocation.\n"
+    "This replacement does not apply to select statements, but the warning might trigger\n"
+    "anyway. Feel free to suppress using the command above."
   autofix: true
 }
 


warn/docs/warnings.textproto
```diff
@@ -394,7 +394,9 @@ warnings: {
 warnings: {
   name: "list-append"
   header: "Prefer using `.append()` to adding a single element list"
-  description: "Transforming `x += [expr]` to `x.append(expr)` avoids a list allocation."
+  description: "Transforming `x += [expr]` to `x.append(expr)` avoids a list allocation.\n"
+    "This replacement does not apply to select statements, but the warning might trigger\n"
+    "anyway. Feel free to suppress using the command above."
```
oreflow, line 399: I'm not sure the "command above" statement really works here?

WARNINGS.md
```diff
@@ -612,6 +612,8 @@ the readability.
   * [Suppress the warning](#suppress): `# buildifier: disable=list-append`
 
 Transforming `x += [expr]` to `x.append(expr)` avoids a list allocation.
+This replacement does not apply to select statements, but the warning might trigger
+anyway. Feel free to suppress using the command above.
```
oreflow, line 616: "Feel free to" sounds a bit informal, consider something like:

"It is recommended to suppress the warning for this edge case"

warn/docs/warnings.textproto
pollendo: you could derive it from the `list-append` name in the textproto, but the text is intended for the generated .md, which contains the command to suppress it right above this message. I'd assume users are reading the .md, not the .textproto.

vladmos: Nit: command -> comment

oreflow: I see, I thought this file was used for warning output as well. Then that's indeed no issue.

WARNINGS.md
oreflow: Closing this comment since other location is more relevant

vladmos (approved):

warn/docs/warnings.textproto
```diff
@@ -394,7 +394,9 @@ warnings: {
 warnings: {
   name: "list-append"
   header: "Prefer using `.append()` to adding a single element list"
-  description: "Transforming `x += [expr]` to `x.append(expr)` avoids a list allocation."
+  description: "Transforming `x += [expr]` to `x.append(expr)` avoids a list allocation.\n"
+    "This replacement does not apply to select statements, but the warning might trigger\n"
+    "anyway. It is recommended to suppress the warning for this edge case using the comment above."
```
oreflow, line 399: I looked into this a bit more, and believe I misunderstood the phrasing about the "replacement"
Consider:

1. A more strict phrasing here
NOTE: `.append()` does not work if the target is a `select`, hence this warning can safely be ignored or suppressed.

2. Updating the actual warning checker output to direct users towards disabling the warning to also suggest the same.
https://github.com/bazelbuild/buildtools/blob/main/warn/warn_operation.go#L192

pollendo: Sorry, I missed your comment. Will follow-up with your second suggestion in another PR today!

Issue #1354, opened 2025-04-25
buildifier rewrite BUILD targets "@abc//:abc" incorrectly.


laramiel:
Our project uses the re2 library. Presently this is named "com_google_re2" in our workspace, but I'm trying to rename the repository to follow more common conventions. 

This rename gives me a BUILD.bazel file with the following:

```
cc_library(
  name = ...
  deps = [
    ...
    "@re2//:re2"
   ]
)
```

After running buildifier on it the deps are rewritten to:

```
  deps = [
  ...
   "@re2",
  ...
  ]
```

This rewrite causes bazel build errors like:

```
ERROR: git/tensorstore/tensorstore/util/BUILD?l=694:23: no such package '@@re2//': The repository '@@re2' could not be resolved: '@@re2' is not a repository rule and referenced by '//tensorstore/util:unit'
```

I believe that this is incorrect; buildifier should leave this dep as-is since in `@re2//:re2`, the repository is `re2`, the package is the empty string, and the target is `re2`.

This behavior is tested in build/testdata/019.build.golden

katre:
Agreed that rewriting `@re2//:re2` as plain `@re2` is incorrect: repositories do not have stable names depending on how they are imported in MODULE.bazel, and so assuming that there is a target with that name is incorrect.

jbms:
I'm a bit confused --- I've definitely successfully used `@repo` to mean `@repo//:repo` in the past.  As long as the `@repo` -> `@repo//:repo` transformation done by bazel is purely syntactic --- i.e. it does not take into account the canonical name of `repo`, then the behavior should be predictable for users and consistent with what buildifier was doing, and the issue of a "stable" name for the repo should not matter.

fmeum:
Reverted in https://github.com/bazelbuild/buildtools/pull/1356, so this can be closed.

PR #1355, opened 2025-04-25
Keep targets like "@foo//:foo" formatted as-is.

laramiel:
Fix incorrect rewrites of `@foo//:foo`.  Encountered when renaming `@com_google_re2//:re2` to `@re2//:re2`.

buildifier should leave these as-is since in `@re2//:re2`, the repository is `re2`, the package is the empty string, and the target is `re2`, and it is not correct to assume that repositories have stable names.

https://github.com/bazelbuild/buildtools/issues/1354


vladmos (approved):
Thanks!

Wyverald:
please revert this. this is a long-standing label syntax (and documented) that the community has come to rely on.

Issue #1330, opened 2025-01-31
buildifier false pos on >>= operator


carpenterjc:
example code 
```
def foo(num):
    num >>= 4
    return num
```

result `no-effect: Expression result is not used. (https://github.com/bazelbuild/buildtools/blob/main/WARNINGS.md#no-effect)`

These operators do have an effect, and you can clearly see the number shifting if you print num before and after.

oreflow:
I believe this was fixed by the linked PR. Feel free to reopen if the issue remains

PR #1357, opened 2025-04-26
Bug fix. Correctly identify expressions using &=, ^=, >>=, and <<= as assignment statements when parsing BUILD files.

pdoyle12:
Correctly identify expressions using &=, ^=, >>=, <<= as assignment statements when parsing BUILD files.

The Starlark grammar reference (https://github.com/bazelbuild/starlark/blob/master/spec.md#grammar-reference) identifies expressions using all of these as assignment statements, but the parser doesn't include them in the list, probably an old oversight.

This fixes issue #1330.

### DIFF
--- a/build/parse.y
+++ b/build/parse.y
@@ -1076,7 +1076,7 @@ func binary(x Expr, pos Position, op string, y Expr) Expr {
 	ystart, _ := y.Span()
 
 	switch op {
-	case "=", "+=", "-=", "*=", "/=", "//=", "%=", "|=":
+	case "=", "+=", "-=", "*=", "/=", "//=", "%=", "&=", "|=", "^=", "<<=", ">>=":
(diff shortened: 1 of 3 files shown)


vladmos (approved):
Could you please add some basic tests (to `build/testdata/052.in`/`golden` for the formatter and to `warn/warn_control_flow_test.go` for the issue mentioned in the linked bug)?

pdoyle12:
I've added tests in the warn_control_flow_test.go file. (I did validate that these correctly catch the specific problem mentioned in the linked bug.) It turns out there already are formatting tests for these cases in the golden file (052.in); if you'd like a more specific test as well, let me know.

vladmos:
You're right, I thought only some of the assignment expressions were mentioned there (as binary expressions, but there's no difference for the formatter).

PR #1363, opened 2025-05-29
Add Starlarkified Python attributes to generate_tables.go

brandjon:
This ensures backwards compatibility in the CLI for attributes like "main" and "python_version". Otherwise, they require quoting, when building buildozer against newer versions of Bazel.


oreflow:
Changes were merged in #1365 with some additional fixes

PR #1366, opened 2025-06-02
Support multi-symbol replace_load correctly

sbarfurth:
The logic for removing old loading symbols replaced by `replace_load` was flawed. It would leave behind old (redundant) loads, iff the `replace_load` was to replace two symbols from the same location. This change simplifies how loads are replaced by simply keeping track of non-replaced loads rather than trying to remove from the list with index manipulation.

I've added a few more test cases. Most of these new tests will fail on the old code.


sbarfurth:
Should I be requesting a review from someone else? @vladmos @pmbethe09 @oreflow

PR #1372, opened 2025-06-18
make backslashes visible in docs for dict_add and dict_set

dws:
In order to escape colons in the keys and values that you provide to dict_add and dict_set, you must precede them with backslashes.  The raw README.md file has the backslashes, but they are not visibile when rendered.  We here double these backslashes so that they'll be visible when rendered.


dws:
@oreflow FYI

oreflow:
Nice, thanks for fixing this :)

oreflow (approved):

PR #1373, opened 2025-06-21
Also print `fixed ...` messages with `-stdout`

fmeum:
This makes it easier for users to implement a "check only" mode that doesn't modify any files, but is able to tell which ones would have been changed by buildozer.

Work towards https://github.com/bazelbuild/bazel/issues/24263


fmeum:
@Wyverald FYI

dws:
+1 to getting this landed

fmeum:
@vladmos

buildozer/buildozer_test.sh
```diff
@@ -2393,4 +2393,32 @@ EOF
...
+
+  cat > MODULE.bazel.expected.stdout <<EOF
+module(
+    name = "foo",
+    version = "0.27.0",
+)
+EOF
+
+  $buildozer -stdout 'format' //MODULE.bazel:all > stdout 2> stderr
+  if [[ "$(cat stderr)" == "fixed */MODULE.bazel" ]]; then
```
oreflow, line 2417: I think this change looks good and makes sense overall, but could we update the stderr assertion to also use a "expected.stderr" file, and use the diff command instead of the custom if assertion?

It just seems simpler and more aligned with the rest of this test.

oreflow (approved):

fmeum:
@vladmos Would it be possible to get this released so that Bazel can start depending on this behavior?

vladmos:
@fmeum This change causes some failures internally and may need to be reverted. The feedback I got:

> This change seems on the surface desirable as it corrects the flaw that without the "modified:" argument, the consumer could not distinguish RETURN_CODE_OK_NO_CHANGES from RETURN_CODE_OK. (It was getting "NO_CHANGES" no matter what). Unfortunately it causes problems with Python API users that have "out, err = buildozer.Buildozer().invoke()" because the secondary value is now the non-empty contents of stderr which is treated as a success/fail indicator (empty string = good).

Do you have any reasonable changes in mind that would both satisfy your needs and are not breaking at the same point? Or at least easy to fix on the consumer side (without actually parsing stderr)?

fmeum:
It does seem wrong to me to treat `stderr != ""` as indicating an error. Would consumers be able to check the exit code somehow? Is the buildozer Python API public?

meteorcloudy:
@fmeum @vladmos This seems to still blocking the progress of https://github.com/bazelbuild/bazel/pull/25540?

@oreflow in case you could help

PR #1374, opened 2025-06-24
Run buildifier against all MODULE.bazel -like files

psalaberria002:
I noticed the included files in our MODULE.bazel were not being linted.

I tracked it to this runner.bash template.

https://bazel.build/rules/lib/globals/module#include


dmivankov:
So continuation of https://github.com/bazelbuild/buildtools/pull/1266 to actually run on those files automatically

PR #1377, opened 2025-06-25
Don't fail on directories named `foo.bzl`

fmeum:


fmeum:
FYI @alexeagle

alexeagle:
Oh thanks for finding that! I hadn't run into the bug yet.

fmeum:
@vladmos Could you take a look? It's a very small change

PR #1380, opened 2025-07-17
feat: autodetect buck2 file names

lf-:
This interprets BXL files as bzl, which they basically are. It also interprets BUCK files as BUILD, which they also are.

Ref: https://buck2.build/docs/bxl/

### DIFF
--- a/build/lex.go
+++ b/build/lex.go
@@ -137,12 +137,14 @@ func getFileType(filename string) FileType {
 	switch ext {
 	case ".bzl":
 		return TypeBzl
+	case ".bxl": // Buck eXtension Language
+		return TypeBzl
 	case ".sky":
 		return TypeDefault
 	}
 	base := basename[:len(basename)-len(ext)]
 	switch {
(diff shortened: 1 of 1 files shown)


lf-:
I dunno how to write a meaningful test for this change.

vladmos:
There is an existing test in `lex_test.go`, please just add the new filenames there.

Please also add non-trivial combinations such as `buck.bzl` or `BUILD.bxl` (which should resolve to TypeBzl).

vladmos:
Do you want to fix the formatting or the warnings? This change will cause all Bazel-specific warnings (that are normally applied only to BUILD/.bzl files) to be applied to the Buck files as well.

If you only want to fix the formatting, then you don't need to do anything to bxl files (all unknown files are formatted as bzl), and create another filetype for BUCK files (maybe call it `TypeDefaultBuild` instead of `TypeBuck` as it's not BUCK-specific but only suggests that the file should be formatted using BUILD file rules).

PR #1384, opened 2025-08-09
buildozer: add new_if_absent command

amartani:
This change adds a new command 'new_if_absent' to buildozer.

The 'new_if_absent' command behaves the same as the 'new' command, but it does not fail if a rule with the same name already exists. Instead, it performs a no-op.

This is useful for scripts that want to ensure a rule exists without having to check for its existence first.


amartani:
I applied gemini's suggestions and merged current main. ptal

PR #1386, opened 2025-08-25
Add a lint warning for external repository references and usage canonical repository names

nickschaap:
In Bzlmod, the usage of canonical repository names should be avoided as it is not a public API makes BUILD files less portable. Developers also should never hardcode paths inside external repos.


Silic0nS0ldier:
With the release of buildtools 8.5.1 this is now in the wild. These new lints have caught have caught several problems but they've also hit several false positives.

* The `/external/` string check includes comment strings. While these can include examples of `/external/` (e.g. describing file system layout) they can not affect build behaviour and should be excluded from linting.
  ```starlark
  def foo():
      """
      ...
          File("bazel-out/k8-fastbuild/bin/external/rules_pycross++lock_repos+pypi/_lock/boto3@1.38.27"),
                                          ^^^^^^^^^^
      ...
      """
  ```
* As above, but for `@@` in comment strings.
* Not all `/external/` substrings refer to the `external` directory (used by external repositories and may be removed in favour of a sibling layout in the future). For example, a `go_library` target under a directory called `external` ends up with an `importpath` attribute that trips this up.
* The latter is also _technically_ true for `@@`, although the sequence is rare enough to not be a problem in practise.

Comment strings should be ignored, though I'm not sure how false positives in other places can be avoided (it may well be impossible).

fmeum:
@oreflow @vladmos Could we roll this back in a new minor/patch release? The idea and intentions are great, but this applies far too widely and results in a lot of avoidable churn. I can help narrow down the scope of a reland.

PR #1387, opened 2025-08-26
enable_workspace when bazel verison > 8

fzakaria:
If you don't want workspace enabled and are on version 8, unused_deps did not work due to the lack of workspace support.

You could either fix this by adding it to your project's bazelrc however we conditionally enable it here depending on the version of bazel being executed.

This does have the side-effect that it seems to cause a rebuild due to the arguments of the CLI changing.

### DIFF
--- a/unused_deps/unused_deps.go
+++ b/unused_deps/unused_deps.go
@@ -76,6 +76,37 @@ javac_params = aspect(
 `
 )
 
+/**
+ * Parses `bazel version` output and returns the version string.
+ * Example output:
+ * Build label: 8.2.1
+ * Build target: bazel-out/darwin-fastbuild/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
+ * Build time: Wed Jan 27
+ */
+func bazelMajorVersion() (int, error) {
+	out, err := cmdWithStderr(*buildTool, "version").Output()
+	if err != nil {
+		return 0, err
+	}
+	lines := strings.Split(string(out), "\n")
+	for _, line := range lines {
+		if strings.HasPrefix(line, "Build label: ") {
+			version := strings.TrimPrefix(line, "Build label: ")
+			parts := strings.Split(version, ".")
+			if len(parts) < 1 {
+				return 0, fmt.Errorf("could not parse version from 'bazel version' output: %s", version)
+			}
+			var major int
+			_, err := fmt.Sscanf(parts[0], "%d", &major)
+			if err != nil {
+				return 0, fmt.Errorf("could not parse major version from 'bazel version' output: %s", version)
+			}
+			return major, nil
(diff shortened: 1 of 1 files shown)


meisterT:
cc @meteorcloudy - I am not sure this is the right thing to do as we are moving away from WORKSPACE

meteorcloudy:
Yes, I think we should move buildtools towards being more MODULE.bazel native.

oreflow:
I think as mentioned by others here on the thread that this change should not be submitted, based on https://bazel.build/versions/8.3.0/about/roadmap#bzlmod:-external, it seems like the `--enable_workspace` flag will be removed altogether with Bazel 9.0, so enabling the flag to be set here by default seems to be against Bazel direction, and will break with Bazel 9.

Additionally enabling it by default might steer others towards relying on the behavior

Since you suggest this can be done in the bazelrc I would recommend doing that instead as an intermediate solution and trying to migrate away from using this flag as it won't be future-safe.

```
https://bazel.build/versions/8.3.0/about/roadmap#bzlmod:-external

“With Bazel 8, we will disable WORKSPACE support by default (it will still be possible to enable it using --enable_workspace); with Bazel 9 WORKSPACE support will be removed. Starting with Bazel 7.1, you can set --noenable_workspace to opt into the new behavior.”
```

joca-bt:
The tool still has issues when running on bzlmod. Is there any alternative fix planned?

```
❯ unused_deps //modules/libs/...
2026/01/27 11:24:02 running: bazel query --tool_tag=unused_deps --keep_going --color=yes --curses=yes kind('(kt|java|android)_*', //modules/libs/...)
2026/01/27 11:24:03 running: bazel build --tool_tag=unused_deps --keep_going --color=yes --curses=yes --output_groups=+unused_deps_outputs --override_repository=unused_deps=/tmp/unused_deps552745060 --aspects=@@unused_deps//:unused_deps.bzl%javac_params //modules/libs/...
WARNING: errors encountered while analyzing target '//modules/libs/db-client:db-client', it will not be built.
error loading package 'external': the repository @@unused_deps does not exist, but has been specified as overridden with --override_repository. Use --inject_repository instead to add a new repository.
...
INFO: Found 17 targets...
ERROR: command succeeded, but not all targets were analyzed
INFO: Elapsed time: 0.463s, Critical Path: 0.00s
INFO: 7 processes: 506 action cache hit, 7 internal.
ERROR: Build did NOT complete successfully

2026/01/27 11:24:03 open /home/joguerra/.cache/bazel/_bazel_joguerra/ce80a57e1d8c15075a3f127252ef0b0f/execroot/_main/bazel-out/k8-fastbuild/bin/modules/libs/db-client/db-client.javac_params: no such file or directory
...
```

I'm on Bazel 8.5.1.

PR #1393, opened 2025-09-12
Updating positional-args check to be more comprehensive

oreflow:
- Previous implementation just assumed that macros were called at the root of a BUILD file, and that all calls were to macros/rules except for explicit allowlist. Updating the check to more precisely validate whether something is a rule/macro, and checking all call expressions (including ones nested in dicts/attrs).
- Also updating IsMacro implementation to correctly support symblic macro definitions (`my_macro = macro(...)`)


warn/warn_bazel_test.go
```diff
@@ -89,13 +89,44 @@ exports_files(["foo.txt"])
 
 func TestPositionalArguments(t *testing.T) {
 	checkFindings(t, "positional-args", `
+my_macro = macro()
```
AnnaSvalova, line 92: I'd split this test into several smaller ones. It's difficult to quickly understand what "line 23" or "line 11" is.

warn/warn_macro.go
```diff
@@ -243,12 +246,20 @@ func (ma macroAnalyzer) IsMacro(fn function) (report functionReport) {
 		return
 	}
 
-	// Check whether fn.name is a rule
-	if fileData.rules[fn.name] {
+	// Check whether fn.name is a rule or macro
+	if fileData.rulesOrMacros[fn.name] {
```
AnnaSvalova, line 250: it's weird that we collect rules and macros in this list, but set report.isMacro to _true_. The description of this function says "whether the given function is a macro". You need to update the description, the field name and the function name, or split the _rulesOrMacros_ list into 2, probably.

warn/warn.go
```diff
@@ -219,6 +217,7 @@ var MultiFileWarningMap = map[string]func(f *build.File, fileReader *FileReader)
 	"native-sh-binary":                   NativeShellRulesWarning("sh_binary"),
 	"native-sh-library":                  NativeShellRulesWarning("sh_library"),
 	"native-sh-test":                     NativeShellRulesWarning("sh_test"),
+	"positional-args":                    positionalArgumentsWarning,
```
AnnaSvalova, line 220: Out of curiosity: why is this change needed?

AnnaSvalova (changes requested):

warn/warn.go
oreflow: The function definition of the previous is designed to run for a single Expr, this one is for the entire file
- Previous one does not receive the full file as input
- Previous one can only return a single finding.

PR #1398, opened 2025-09-30
buildifier: decouple runner scripts from binaries

rdesgroppes:
This change aims at making `buildtools/buildifier` more modular and easier to consume by enabling use of prebuilt binaries without having to pull in unnecessary dependencies - while strengthening upstream as the authoritative home for runner script templates.

Today, `buildifier/BUILD.bazel` exposes two different sets of rules:
- exports for runner script templates (`runner.bash.template`, `runner.bat.template`) as well as `README.md`,
- `go` build & `sh` test rules which require external repos (`rules_go`, `rules_shell`) even for consumers who don't need to build the `buildifier` binary.

It is worth noticing there are two distinct community efforts on runner scripts, the present/upstream `bazelbuild/buildtools` and the derived `keith/buildifier-prebuilt` ([Bazel Central Registry module](https://registry.bazel.build/modules/buildifier_prebuilt)). These drift apart over time, leading to unsynchronized contents ([factory.bzl](https://github.com/keith/buildifier-prebuilt/blob/main/buildifier/factory.bzl), [runner.bash.template](https://github.com/keith/buildifier-prebuilt/blob/main/runner.bash.template)) and efforts, e.g. on Windows support:
- #1011
- keith/buildifier-prebuilt#89 (open for a year)

Intended benefits of this change:
- decoupling: by moving `go_binary` and `sh_test` targets into `buildifier/cmd`, the top-level `buildifier/BUILD.bazel` no longer requires `rules_go` nor `rules_shell`,
- flexibility: users can either build binaries from source (via `//buildifier/cmd`) or configure rules to use prebuilt binaries (e.g. via [rules_multitool](https://registry.bazel.build/modules/rules_multitool)) while still consuming the same upstream runner script templates,
- backward compatibility: aliases keep earlier `//buildifier:buildifier*` labels working by delegating to the corresponding `//buildifier/cmd:buildifier*`.


AnnaSvalova (commented):
I'm not against this change, but `cmd` sounds a bit unusual for build tools and bazel 🤔 Maybe `cli` or similar?

rdesgroppes:
> I'm not against this change, but `cmd` sounds a bit unusual for build tools and bazel 🤔 Maybe `cli` or similar?

@AnnaSvalova, I'll of course follow your call here, but I was wondering whether the Go convention should be favored in this case? The `cmd` directory seems to be the standard location for CLI applications there: https://github.com/golang-standards/project-layout?tab=readme-ov-file#cmd

I thought it might be helpful to stick with it for familiarity. But happy to change it to `cli` if you prefer:
- [ ] rename to `cli`
- [ ] stick to `cmd`

AnnaSvalova:
> > I'm not against this change, but `cmd` sounds a bit unusual for build tools and bazel 🤔 Maybe `cli` or similar?
> 
> @AnnaSvalova, I'll of course follow your call here, but I was wondering whether the Go convention should be favored in this case? The `cmd` directory seems to be the standard location for CLI applications there: https://github.com/golang-standards/project-layout?tab=readme-ov-file#cmd
> 
> I thought it might be helpful to stick with it for familiarity. But happy to change it to `cli` if you prefer:
> 
> * [ ]  rename to `cli`
> * [ ]  stick to `cmd`

Fair point. Also, I'd say that an even more common pattern is using a project name instead of "cli" or "cmd".

In any case it was a nit, LGTM.

AnnaSvalova (approved):

gunsch:
Drive-by FYI --- I think this broke the `go install` command listed in the docs.

i.e. as [documented here](https://github.com/bazelbuild/buildtools/tree/main/buildifier#setup):

```
go install github.com/bazelbuild/buildtools/buildifier@latest
```

Difference between your commit and the one prior:

```
$ go install github.com/bazelbuild/buildtools/buildifier@e879524f2986d9b75488855aa938013db2cda4ea
go: downloading github.com/bazelbuild/buildtools v0.0.0-20251107112229-e879524f2986

$ go install github.com/bazelbuild/buildtools/buildifier@9a9819efe30f
go: github.com/bazelbuild/buildtools/buildifier@9a9819efe30f: module github.com/bazelbuild/buildtools@9a9819efe30f found (v0.0.0-20251110131859-9a9819efe30f), but does not contain package github.com/bazelbuild/buildtools/buildifier
```

Edit: looks like this works now instead:

```
$ go install github.com/bazelbuild/buildtools/buildifier/cmd@9a9819efe30f
```

DavyWk:
Emphasizing @gunsch's comment above- we use the documented `go install github.com/bazelbuild/buildtools/buildifier@latest` and several of our builds broke.

rdesgroppes:
@AnnaSvalova @gunsch @DavyWk I just filed:
- #1413

thejcannon:
You can't win 'em all. We got broken too, so I fixed it to use the new location. Guess who just got broken again 😅 ?

If anything this is a reminder using `@latest` is a footgun, and I, now with two holes in my feet, have run out of feet.

PR #1399, opened 2025-10-07
buildifier: honor `exclude_patterns` on Windows too

rdesgroppes:
The `exclude_patterns` parameter passed to `buildifier` macros happens to be ignored by the Windows runner script template, which the present change proposes to address.

The change tries to adhere to existing choices, by limiting the impact on `factory.bzl` where the `find`-suitable pattern can't be used by the embedded powershell script, and also by respecting the script's coding style.


oreflow:

PR #1426, opened 2025-12-04
Add wildcard support for key in `dict_replace_if_equal`

mering:


edit/buildozer.go
```diff
@@ -767,17 +767,36 @@ func cmdDictReplaceIfEqual(opts *Options, env CmdEnvironment) (*build.File, erro
 		return env.File, nil
 	}
```
oreflow: Can we add a test case which verifies the new behaviour?

mering:
@oreflow Thanks for the review! I addressed all comments. Can you please take another look? Thanks!

oreflow (approved):

PR #1427, opened 2025-12-06
Add support for mapped repository names in use_repo_add

malt3:
Add support for importing repositories under a custom local name, and add a
new use_repo_set command.

Repositories can be specified as:
- Simple names: foo -> use_repo(ext, "foo")
- Mapped names: bar=baz -> use_repo(ext, bar = "baz"), importing the
  repository baz under the local name bar. A mapping whose local name equals
  its value (foo=foo) is simplified to foo.

use_repo_add leaves already-present arguments untouched (existing mappings are
never overwritten), consistent with the add/dict_add commands. use_repo_set
replaces any existing import that shares a local name or repository value with
one of the given repos (e.g. to rename a mapping). Both delegate to a single
internal implementation.

Buildozer does not validate that local names are valid Starlark identifiers;
invalid names are emitted verbatim.


edit/bzlmod/bzlmod.go
```diff
@@ -118,10 +120,20 @@ func NewUseRepo(f *build.File, proxies []string) (*build.File, *build.CallExpr)
 	return &build.File{Path: f.Path, Comments: f.Comments, Stmt: stmt, Type: build.TypeModule}, useRepo
 }
 
+// repoToAdd represents a repository to be added to a use_repo call.
+type repoToAdd struct {
+	key          string
+	value        string
+	isMapping    bool
+	isValidIdent bool
```
fmeum: Could these two fields become functions of `key` and `name` instead? If someone specifies `foo=foo`, that could just be simplified to a positional arg.

edit/bzlmod/bzlmod.go
```diff
@@ -432,3 +591,89 @@ func collectApparentNamesAndIncludes(f *build.File) (map[string]string, []string
...
+
+	// Check if it's a reserved keyword
+	if starlarkReservedKeywords[s] {
+		return false
+	}
+
+	for i, r := range s {
+		if i == 0 {
+			// First character must be letter or underscore
+			if !unicode.IsLetter(r) && r != '_' {
```
fmeum: At least Bazel's Starlark interpreter only allows ASCII characters in identifiers, not arbitrary Unicode letters: https://cs.opensource.google/bazel/bazel/+/d61a021aa63739f28ae1ba9f00ecf0c05e3edd09:src/main/java/net/starlark/java/syntax/Identifier.java;l=86

malt3:
@oreflow gentle ping. This could unblock a change in Bazel (see https://github.com/bazelbuild/bazel/pull/27891).

oreflow (commented):
Starting off with high level topics before going into code details.

### **kwargs support bar=baz -> use_repo(ext, bar = "baz")
This part of the change looks good overall, but there's an implicit assumption added that if calling `use_repo_add` with an existing key, the argument is replaced, which is inconsistent with other buildozer behaviors as `add` and `dict_add` both state that `"If the key was already present, it will not be overwritten."` (see difference between `dict_set` and `dict_add`).

### foo.2=foo -> use_repo(ext, **{"foo.2": "foo"})
This part of the change adds a fair bit of complexity in parsing and validation, which also in their implementation require embedding part of the Starlark language specification in this repo. Just as an example, the list of reserved Bazel keywords does not seem to match https://github.com/bazelbuild/starlark/blob/master/spec.md#built-in-constants-and-functions, and it would be infeasible for us to maintain this in the buildtools repo as the starlark language evolves. So if this is kept we should look into using some authorative source to import on this rather than hardcoding it in this codebase.

Do you have a specific use case in mind where you would need to add repos with names matching matches reserved keywords? It seems a bit like this should just be discouraged, even if it would be possible using dict unpacking it seems better if this would return an error.

Additionally behaviors of Buildozer are generally not trying to be smart, so the conditional conversion of reserved / invalid identifiers into dict unpacking might be confusing to users. It might be simpler to place this in the hands of the users, e.g. if a user really wants to add dict-unpacked-values with `use_repo_add`, it might be more intuitive to support a syntax like `use_repo_add {"foo": "bar"}`, rather than doing a conversion based on logic here.

malt3:
Thank you for taking a look!

> This part of the change looks good overall, but there's an implicit assumption added that if calling use_repo_add with an existing key, the argument is replaced, which is inconsistent with other buildozer behaviors [...]

I agree that this is inconsistent. That being said, replacing existing keys is the behavior we want from `bazel mod tidy`, which is the use case I have in mind. That leaves me with two options: either we add `use_repo_set` with the current behavior (replace keys if they already exist) and change `use_repo_add` to leave keys unchanged if they already exist, or we accept this inconsistency and document it clearly. What's your preference?
A third option would be to keep `use_repo_add` completely unchanged (no support for mapping repo names) and reserve the mapping functionality for `use_repo_set`.

> Do you have a specific use case in mind where you would need to add repos with names matching matches reserved keywords? It seems a bit like this should just be discouraged, even if it would be possible using dict unpacking it seems better if this would return an error.

I was merely trying to support the full set of possible repo names according to Bazel's documentation. I do agree that this introduces a lot of complexity.
Furthermore, I would be fine to drop support for dict unpacking and simply return an error. This slightly limits the connected Bazel feature of mapping repo names via extension metadata, but that limitation is acceptable. Does that work for you?

I'll simplify and adapt this PR based on your feedback.

oreflow:
> I agree that this is inconsistent. That being said, replacing existing keys is the behavior we want from bazel mod tidy, which is the use case I have in mind. That leaves me with two options: either we add use_repo_set with the current behavior (replace keys if they already exist) and change use_repo_add to leave keys unchanged if they already exist, or we accept this inconsistency and document it clearly. What's your preference?

I would go with `use_repo_set`, just for the consistency, I'd imagine implementation/complexity-wise it could still just go in a single internal function for `_add` and `_set` which just takes a boolean whether to replace or ignore existing values.

> A third option would be to keep use_repo_add completely unchanged (no support for mapping repo names) and reserve the mapping functionality for use_repo_set.

I believe the mapping addition to `use_repo_add` makes sense, plus it's nice for `_add` and `_set` to support the same feature set. But feel free to omit it if you feel it adds complexity.

> Furthermore, I would be fine to drop support for dict unpacking and simply return an error. This slightly limits the connected Bazel feature of mapping repo names via extension metadata, but that limitation is acceptable. Does that work for you?

I think the Error approach is good, however this would still require keeping the parsing and validation, I would even go as far as just allowing users to set mappings with invalid names, as mentioned Buildozer doesn't really deal with validation / opinionated / smart functionality. So if someone explicitly tells Buildozer to do something which will be invalid, I believe we can simply allow them to do so to keep the functionality simple and intuitive, in general Buildozer is more just for the convenience of automated edits, rather than validation.

Sidenote, If we wanted validation it might make sense to add a Buildifier warning for any re-use of reserved keywords / invalid identifiers (IIRC this doesn't exist already). That way it would also support a lot more use cases. But not really tied to this PR

Thanks for following up so actively and working on this :)

malt3:
Hi @oreflow. Apologies for dropping the ball on this.
I've simplified the code a lot and followed your suggestions.
Could you take another look? Thank you!

PR #1430, opened 2025-12-11
Add `-lint=suggest` to include replacement in `--format=json` output

Silic0nS0ldier:
This PR adds a new CLI flag value for `-lint`, `suggest`. When used in conjunction with `--format=json` the output populates the new field `.files[].warnings[].replacement`.

Note that `start` and `end` are inconsistent with the existing fields.
- `.files[].warnings[].start` and `.files[].warnings[].end`
  - `.line` is a 1-indexed line number.
  - `.column` is a 1-indexed UTF-16 code unit.
- `.files[].warnings[].replacement.start` and `.files[].warnings[].replacement.end`
  - `.` is a 0-indexed byte positions.

Replacement positions are derived from `[]byte` arrays, so conversion (to the correct encoding) would be necessary. But also;
- Correctly detecting line endings in a manner consistent with the parser.
- Correctly handling multi-byte sequences (and potentially Unicode grapheme clusters) in a manner consistent with the parser.

As a result normalisation is difficult, bug prone, and unless `[]byte` usage is eliminated close to it's source, liable to carry a (minor) performance and memory penalty.

Output sample;

```
❯ buildifier -lint=suggest -mode=check -format json -v ___/BUILD.bazel
{
    "success": false,
    "files": [
        {
            "filename": "___/BUILD.bazel",
            "formatted": false,
            "valid": true,
            "warnings": [
                {
                    "start": {
                        "line": 259,
                        "column": 1
                    },
                    "end": {
                        "line": 259,
                        "column": 10
                    },
                    "category": "native-java-test",
                    "actionable": true,
                    "autoFixable": true,
                    "message": "Function \"java_test\" is not global anymore and needs to be loaded from \"@rules_java//java:java_test.bzl\".",
                    "url": "https://github.com/bazelbuild/buildtools/blob/main/WARNINGS.md#native-java-test",
                    "replacement": {
                        "start": 57,
                        "end": 57,
                        "content": "@rules_java//java:java_test.bzl\", \"java_test\")\nload(\""
                    }
                }
            ]
        }
    ]
}
```

The use case for this change is integrating suggested fixes with an internal tool.

### DIFF
--- a/buildifier/utils/diagnostics.go
+++ b/buildifier/utils/diagnostics.go
@@ -82,7 +82,7 @@ type warning struct {
 	AutoFixable bool         `json:"autoFixable"`
 	Message     string       `json:"message"`
 	URL         string       `json:"url"`
-Replacement *replacement `json:"replacement,omitempty"` // Optional replacement
+    Replacement *replacement `json:"replacement,omitempty"` // Optional replacement
 }
 
 type position struct {
--- a/buildifier/utils/diagnostics.go
+++ b/buildifier/utils/diagnostics.go
@@ -19,9 +19,10 @@ package utils
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
+
 	"github.com/bazelbuild/buildtools/build"
 	"github.com/bazelbuild/buildtools/warn"
-	"strings"
 )
 
 // Diagnostics contains diagnostic information returned by formatter and linter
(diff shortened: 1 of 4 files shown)


Silic0nS0ldier:
Pinging @oreflow, is this change something you'd be open to accepting?

oreflow:

Silic0nS0ldier:
Rebased. @fmeum I notice you've been active as a reviewer in this repo, mind taking a look?

fmeum:
Do we really need a new flag value to gate a new JSON field? Could this be the behavior of `warn` mode with `json` format? Adding a new field seems safe.

buildifier/utils/diagnostics.go
```diff
@@ -147,3 +156,14 @@ func makePosition(p build.Position) position {
 		Column: p.LineRune,
 	}
 }
+
+func makeReplacement(r *warn.Replacement) *replacement {
+	if r == nil {
+		return nil
+	}
+	return &replacement{
+		Start:   r.Start,
```
fmeum, line 165: This is the position in the formatted output, but the command may not format. Not sure how to deal with that.

PR #1439, opened 2026-01-14
Disable bzl-visibility lint if the target file has a `visibility()` declaration

haberman:
The [official docs](https://bazel.build/concepts/visibility#bzl-visibility-buildifier-lint) for the bzl-visibility lint say:

> This lint predates the load visibility feature and is unnecessary in workspaces where .bzl files declare visibilities.

This statement is correct; [load visibility](https://bazel.build/concepts/visibility#load-visibility) is a strictly superior mechanism, because it is more explicit and more strictly enforced. All code should move towards `visibility()` and away from this legacy `private/` naming scheme.

However, this lint currently fires even in cases where visibility has been explicitly specified. This is unfortunate, because the usage may be intended, in which case the warning is pure noise. The only way to disable the warning in that case is to put the following comment on the code *performing* the import:

```python
# Currently buildifier forces me to put the following comment on the load() statement
# even if foo.bzl has added me to its visibility list:

load("//some/package/private:foo.bzl", "some_symbol") # buildifier: disable=bzl-visibility
```

If `foo.bzl` has explicitly added the caller to its `visibility()` list, this extra comment is pure noise that adds no value. Worse, it's likely to get copied and pasted around, whether or not it is really needed.

We should fix this by making the warning notice that the target file has an explicit `visibility()` definition. If so, there is no need to emit the legacy warning.

MARKDOWN=true

### DIFF
--- a/warn/warn_visibility_test.go
+++ b/warn/warn_visibility_test.go
@@ -127,3 +127,21 @@ load("//foo/private:bar.bzl", "abc")
 		[]string{},
 		scopeEverywhere)
 }
+
+// Test that the warning *is* issued if the module does *not* have a visibility statement.
+func TestBzlVisibilityWithoutVisibilityStatement(t *testing.T) {
+	defer setUpFileReader(map[string]string{
+		"foo/private/bar.bzl": `
+
+abc = 1
+`,
+	})()
+
+	checkFindings(t, "bzl-visibility", `
+load("//foo/private:bar.bzl", "abc")
+`,
+		[]string{
+			`Module "//foo/private:bar.bzl" can only be loaded from files located inside`,
+		},
+		scopeEverywhere)
+}


warn/warn_visibility_test.go
```diff
@@ -108,3 +108,22 @@ bar1()
 		[]string{},
 		scopeEverywhere)
 }
+
+// Test that the warning is not issued if the module has a visibility statement.
```
oreflow: Since this change enables the multi-file functionality, it would be great if we could add another test case which validates that the same with a .bzl file which has content but no visibility statement. Essentially I believe this test case would pass with the implementation of "if the .bzl file exists, skip the warning"

haberman:
Thanks for the review! PTAL.

Issue #1436, opened 2026-01-09
Buildozer - Allow operations across multiple files


oreflow:
Buildozer operations currently do not allow for changes across multiple files, it would be great if we could support this, either for moving rules/content between files, or things like copying attributes between files.

The team currently lacks bandwidth to expand on the functionality, but this issue is meant to be used for tracking of interest for this in case we have more bandwidth in the future.

soorya38:
Hi @maintainer,
I’d like to work on this issue. I’ve gone through the codebase and the discussion, and I believe I can provide a solid fix.
If this issue isn’t currently assigned, could you please assign it to me?
Thanks!

oreflow:
Hi @soorya38,

Thank you for volunteering to take this on :)

I ended up marking this as low priority since I believe there's a lot of hidden complexity in this where there's code paths which in its current state are making the assumption that we're operating on a single file, and may start failing.

In addition to this, having a solid definition for this might require rewriting definitions of a number of rules, and the core operation which currently accepts a single `*File`.

So that being said, and just being assumptions as I haven't dug into details of what changes would be required... If you feel like you have a good grasp and idea for this, and you are able to create a PR following [Google guidelines for change sizes](https://google.github.io/eng-practices/review/developer/small-cls.html). I'd be happy to review a PR.

Just note though that if this grows as complex as I'd expect we might not currently have capacity to review and support getting it merged. So there's a risk of it being throwaway work

PR #1441, opened 2026-01-19
buildozer: support multi-file copy operations

### LINKED ISSUE #123: Buildozer stack-overflows on Windows platforms, when not in WORKSPACE directory
It seems that on Windows platforms, Buildozer will stack-overflow, if no super-directory contains a WORKSPACE file.

Root cause is in the function "func Find(dir string) (string, error) {...}" in the file "buildtools/wspace/workspace.go". The condition:

if dir == "" || dir == "/" || dir == "."

which checks whether the top-level directory has been reached, does not seem to take into account the possibility that the top-level directory is something like "C:\". Stack-overflow seems to be the result.

soorya38:
This PR significantly enhances the `copy` (and `copy_no_overwrite`) command in **buildozer** by adding support for **cross-file operations**.

Previously, buildozer assumed that the source rule for a copy operation existed in the **same BUILD file** as the target rule. This change removes that limitation.

## Related issue: #1436 

---

## What’s Changed

When the source rule label refers to a different package/file (for example):

```bash
buildozer 'copy deps //pkg1:rule1' //pkg2:rule2
```

buildozer will now:

1. Detect that the source rule is located in a different BUILD file
2. Load and parse the source BUILD file on demand
3. Retrieve the requested attribute from the source rule
4. Apply it to the target rule in the current file

This enables copying or moving attributes across multiple BUILD files, addressing a known limitation in current behavior.

---

## Testing

* **Local Reproduction**
  Verified using a script that creates two packages (`pkg1`, `pkg2`) and successfully copies the `deps` attribute from a rule in `pkg1` to a rule in `pkg2`.

* **Existing Tests**
  Ran the following to ensure no regressions:

  ```bash
  go test ./edit/...
  ```

---

## Related Issue

* Fixes #123 *(update with the correct issue number if applicable)*

---

## Checklist

* [x] Signed the CLA (Contributor License Agreement)
* [x] Ran tests (`go test ./...`)

### DIFF
--- a/edit/buildozer.go
+++ b/edit/buildozer.go
@@ -811,7 +811,26 @@ func cmdDictListAdd(opts *Options, env CmdEnvironment) (*build.File, error) {
 }
 
 func copyAttributeBetweenRules(env CmdEnvironment, attrName string, from string) (*build.File, error) {
-	fromRule := FindRuleByName(env.File, from)
+	// Check if the source rule is in another package
+	buildFile, _, _, rule := InterpretLabelForWorkspaceLocation(NewOpts().RootDir, from)
+	var fromRule *build.Rule
+
+	if buildFile != "" && buildFile != env.File.Path {
+		// The rule is in another file, we need to load it
+		data, _, err := file.ReadFile(buildFile)
+		if err != nil {
+			return nil, fmt.Errorf("could not read file '%s': %v", buildFile, err)
+		}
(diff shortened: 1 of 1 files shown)


oreflow:
Apologies for the delay here. As mentioned in #1436 we are quite short on capacity. Closing this PR and unassigning the issue as the PR does not have tests, and Gemini comments were left unresolved.

PR #1446, opened 2026-01-29
Updating github release script to find binaries

oreflow:
- The old script relied on binaries for all targets existing in the symlinked `bazel-bin` directory. This seems to no longer be the case in recent bazel release, so this update instead uses `aquery` to find the various go_binary GoLink outputs, and uploads them from there.
- Also reduces list of targets to only exist once in the bash script

### DIFF
--- a/release/github.sh
+++ b/release/github.sh
@@ -74,7 +74,7 @@ done
 
 echo "Validating that all expected binaries are present"
 for binary_name in "${all_binary_names[@]}"; do
-  if [ ! -f ""$BIN_DIR/$binary_name"" ]; then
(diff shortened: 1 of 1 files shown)


vladmos (approved):

bewuethr:
Seeing that [release 10.0.0](https://github.com/bazel-contrib/buildtools/releases/tag/v10.0.0) doesn't have any binaries attached, it looks like this fix didn't quite work?

fmeum:
That was human error on my part, will upload them soon.

fmeum:
v10.0.1 is out with binaries

bewuethr:
@fmeum Excellent, thanks!

PR #1448, opened 2026-02-04
Allow callers to explicitly provide types for set and add commands

vladmos:
Buildozer tries to guess the format of an attribute's value before adding it to the file, sometimes it does it incorrectly. This PR provides a way to explicitly set a type of the attribute. Currently it's only supported for the `set`, `set_if_absent` and `add` commands, but can be extended further for other commands as well (likely will also be required for `dict_set` and `dict_add`).

The syntax is backward-compatible with the old command usages:

    set name bar  # sets name = "bar"
    set name:expr FOO  # sets name = FOO

The only supported type for now is `expr` for arbitrary expressions.

Attribute names can't contain colons, hence no previosly existing calls should be affected by the change. It should also be straightforward for the callers because resembles the syntax for type hints in Python (and Starlark).

### DIFF
--- a/buildozer/README.md
+++ b/buildozer/README.md
@@ -86,8 +86,9 @@ See `buildozer -help` for the full list.
 
 Buildozer supports the following commands(`'command args'`):
 
-  * `add <attr> <value(s)>`: Adds value(s) to a list attribute of a rule. If a
-    value is already present in the list, it is not added.
+  * `add <attr>[:type] <value(s)>`: Adds value(s) to a list attribute of a rule.
+    If a value is already present in the list, it is not added. See
+   [supported types](#supported-types) for the available types.
   * `new_load <path> <[to=]from(s)>`: Add a load statement for the given path,
     importing the symbols. Afterwards, consider running `buildozer 'fix unusedLoads'`.
   * `replace_load <path> <[to=]from(s)>`: Similar to `new_load`, but removes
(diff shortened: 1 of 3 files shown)


edit/buildozer.go
```diff
@@ -558,48 +600,70 @@ func cmdSubstitute(opts *Options, env CmdEnvironment) (*build.File, error) {
...
 			list = append(list, getStringExpr(arg, env.Pkg))
 		}
 		return &build.ListExpr{List: list}
+	case attrType == rawListAttr:
+		// list of arbitrary objects (idents, ints, already quoted strings, etc.)
+		var list []build.Expr
+		for _, arg := range args {
+			list = append(list, &build.Ident{Name: arg})
+		}
+		return &build.ListExpr{List: list}
```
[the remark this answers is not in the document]
vladmos: Review again

oreflow (commented):
What is the specific use case we're aiming to support with this?

## Note on the feature

Just intuitively, adding type support here adds a whole new dimension of complexity across the tools, and the implementation of only supporting it for certain commands feels a bit inconsistent.

## Implementation

I believe the current implementation, given that it adds ~20+ lines of code, with a bunch of conditional branching to each command supporting types would not scale if needing to support this across the tools, the code would just grow too complex. if we want to enable this, I believe we should break out the functionality and implement a set of generic and reusable type-supporting functions which can be reused across different commands. Just for instance, if I'm reading it correctly there's already divergences between `set` and `add` in how they deal with `raw` through `Ident` or `LiteralExpr`

I believe there's also an amount of validation logic missing which may lead to unclear errors when this is used, especially mixed with the current code base often having silent / ignored errors this might be quite tricky to use.

## Option?
Back to question 1 and what the use case is for this, but would it be possible to support the requirement by just allowing a "raw" type, which allows users to modify things in a raw format? it might be a lot simpler to scale and involve less logic.

buildozer/README.md
```diff
@@ -188,6 +194,14 @@ The following commands only apply to `MODULE.bazel` files (e.g. the target
     usages with `dev_dependency = True` will be considered instead. Extension
     usages with `isolated = True` are ignored.
 
+#### Supported types
```
oreflow, line 197: Consider adding an actual example for what this looks like, or linking to the test file here.

oreflow: And maybe also mention that this does allow invalid changes and puts the responsibility with the user.

edit/buildozer.go
```diff
@@ -83,18 +83,52 @@ type CmdEnvironment struct {
...
+const (
+	// notProvidedTypeAttr represents an attr type that's not provided explicitly and
+	// needs to be inferred by buildozer from bazel's build language proto
+	notProvidedTypeAttr = iota
+	// rawAttr is a type for idents, ints or other types that buildozer should
+	// take as is, without quoting or wrapping
+	rawAttr
+)
+
+// parseAttr parses the attr name and optional type, e.g. "foo" or "bar:string"
```
oreflow: Update example to not be "string"

edit/buildozer.go
```diff
@@ -83,18 +83,52 @@ type CmdEnvironment struct {
...
+// The type names are taken from https://bazel.build/rules/lib/toplevel/attr
```
oreflow: please update

oreflow (approved):
Nice! This looks so much simpler and cleaner 🔥

PR #1463, opened 2026-04-14
Catch more file patterns

keith:
Previously `buildifier -r .` would not pick up `foo.BUILD.bazel` files
or `foo.BUILD` files even though the `runner.bash.template` would. This
attempts to unify these.


keith:
@vladmos ptal

vladmos (approved):
Thanks!

PR #1464, opened 2026-05-08
Adding support for .bzl file operations

mydatascience:
## Buildtools PR checklist

- [X] The code in this PR is covered by unit/integration tests.
- [X] I have tested these changes and provide testing instructions below.

## Description

This PR extends edit label resolution to allow targeting .bzl files directly (for buildozer-style edits), e.g. //pkg:defs.bzl and //pkg:defs.bzl:rule. This makes it possible to apply edit transformations (like adding/removing entries from list attributes such as patches = [...]) to .bzl files, while still relying on buildifier-compatible formatting.

It also adds unit tests covering:

resolving .bzl targets via InterpretLabelForWorkspaceLocation
editing a patches list in a parsed .bzl file (add + remove) and validating formatted output

### (optional) These changes were tested using the following steps

Run unit tests for the edited package:
go test ./edit
(Optional sanity) Run the specific new tests:
go test ./edit -run 'TestInterpretLabelForWorkspaceLocation_BzlFile|TestAddAndRemovePatchesInBzlFile'

### DIFF
--- a/build/rule.go
+++ b/build/rule.go
@@ -165,24 +165,31 @@ func (f *File) implicitRuleName() string {
 // begins with a literal, if the call expression does not conform to either of these forms, an
 // empty string will be returned
 func (r *Rule) Kind() string {
-	var names []string
-	expr := r.Call.X
-	for {
-		x, ok := expr.(*DotExpr)
-		if !ok {
-			break
-		}
-		names = append(names, x.Name)
-		expr = x.X
-	}
-	x, ok := expr.(*Ident)
-	if !ok {
+	if r.Call.X == nil {
 		return ""
 	}
-	names = append(names, x.Name)
-	// Reverse the elements since the deepest expression contains the leading literal
-	for l, r := 0, len(names)-1; l < r; l, r = l+1, r-1 {
-		names[l], names[r] = names[r], names[l]
+	// The kind may be a simple identifier (e.g. `go_library`) or a dotted expression
+	// (e.g. `native.go_library`). Extract the dotted path conservatively.
+	var names []string
+	var walk func(Expr) bool
+	walk = func(e Expr) bool {
+		switch e := e.(type) {
+		case *Ident:
+			names = append(names, e.Name)
+			return true
+		case *DotExpr:
+			// DotExpr represents `X.Name`.
+			if !walk(e.X) {
+				return false
+			}
+			names = append(names, e.Name)
+			return true
+		default:
+			return false
+		}
+	}
+	if !walk(r.Call.X) {
+		return ""
 	}
 	return strings.Join(names, ".")
 }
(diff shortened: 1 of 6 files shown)


mydatascience:
/gemini review
--

edit/buildozer.go
```diff
@@ -1132,11 +1164,19 @@ type rewriteResult struct {
...
+				if _, ok := frame.(*build.DefStmt); ok {
+					return
+				}
 			}
-		}
+			if as, ok := x.(*build.AssignExpr); ok {
+				if lhs, ok := as.LHS.(*build.Ident); ok {
+					vars[lhs.Name] = as
+				}
+			}
```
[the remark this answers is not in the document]
mydatascience: Yes — that's a real limitation. Intended. Added to README

oreflow:
PTAL at the buildkite errors before we spend time reviewing this

mydatascience:
The previous fail was due to CI infra. Rerun is fine

build/rule.go
oreflow, line 1: Do you happen to have some further reasoning / requirements for the variable support?

It has been an intentional decision for the buildtools to identify targets using the "name" attribute since it generally simplifies BUILD file management as well as large scale changes across the google monorepo.

Use of variables also generally tends to have poor effects on build health, are not really aligned with https://bazel.build/build/style-guide#prefer-damp-build-files-over-dry, and is something we generally discourage.

So it would be interesting to learn what the specific need is, from what I can tell `patches` would generally be part of a target with a `name` as well?

Additionally the change to support variables seems to add some non-insignificant complexity across many of the commands.

buildozer/README.md
```diff
@@ -253,7 +273,7 @@ buildozer 'new cc_binary new_bin before tests' //:__pkg__
 # Copy an attribute from `protolib` to `py_protolib`.
 buildozer 'copy testonly protolib' //pkg:py_protolib
 
-# Set two attributes in the same rule
+# Set two attributes in the same  rule
```
oreflow, line 276: please fix

edit/edit.go
```diff
@@ -38,6 +38,23 @@ var (
...
 
+func resolveBzlTarget(basePath, rule string) (bzlFile string, rulePart string, ok bool) {
+	if !strings.Contains(rule, ".bzl") {
+		return "", "", false
+	}
+	idx := strings.Index(rule, ".bzl")
+	bzlPart := rule[:idx+4]
+	rulePart = rule
+	if idx+4 < len(rule) && rule[idx+4] == ':' {
+		rulePart = rule[idx+5:]
```
oreflow, line 49: I believe this is safe since all index offsets are in the string match with ".bzl", but just checking if you have tried this with unicode characters in the label to see if it might require rune based manipulation instead of char-index?

oreflow (commented):
The .bzl expansion support LGTM as long as you add some documentation for that targets may be specified as such, but for the variable modifications we would require a strong reason to step back on that nodes are accessed and modified via the "name" attribute, if you have a specific case in mind it might even be better to support said case better via native rules instead of variables.

PR #1471, opened 2026-06-09
Revert "updates (#1386)"

fmeum:
## Buildtools PR checklist

## Description

This reverts commit 4006b543a694f6cf77c2d8cc188c5f53f3bac1d9. As reported in https://github.com/bazelbuild/buildtools/pull/1386#issuecomment-3838906721 and below, this has numerous false positives and causes a lot of churn. The checks could potentially still be useful if scoped far more tightly, but determining this scope isn't trivial, hence a revert.


fmeum:
@oreflow @Wyverald PTAL

fmeum:
@vladmos @oreflow Friendly ping

Ahajha:
+1 on this, I just filed https://github.com/bazelbuild/buildtools/issues/1492 which is yet another indication of the churn caused by `canonical-repository`.

fmeum:
@alexeagle Ping

alexeagle (approved):
Don't we have an option of keeping the code but making the checks no longer included in the default set?

fmeum:
We could do that, but even then we should probably work on avoiding some of the false positives. Bringing back a part of the code from git history would be easy at that point.

PR #1472, opened 2026-06-09
Fix another only written to unused variable case

keith:
If you have an array was used like:

```bzl
items = []
items.append(1)
```

This would be considered used even if it was never read. Now some known
functions are ignored so it can still be considered unused in this case.

### DIFF
--- a/warn/warn_control_flow.go
+++ b/warn/warn_control_flow.go
@@ -231,6 +231,7 @@ var mutatingMethodsReturningNone = map[string]bool{
 	"extend": true,
 	"insert": true,
 	"remove": true,
+	"sort":   true,
 	"update": true,
 }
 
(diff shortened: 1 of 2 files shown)


keith:
hrm this heuristic might not work:

```bzl
def foo(items):
    items.append(1)
```

this shows items as unused. let me see if i can exclude params

keith:
that case is potentially fixable, but this case i found in rules_go is more problematic:

```bzl
def apple_ensure_options(ctx, env, compiler_option_lists, linker_option_lists, target_gnu_system_name):
    """Returns environment and flags for Apple targets."""
    platform, platform_type = _PLATFORMS.get(target_gnu_system_name, (None, None))
    if not platform:
        return
    env.update(_apple_env(ctx, platform))
    min_version = _apple_version_min(ctx, platform, platform_type)
    for compiler_options in compiler_option_lists:
        compiler_options.append(min_version)
    for linker_options in linker_option_lists:
        linker_options.append(min_version)
```

here the logic diagnoses `compiler_options` as unused but it's mutating things again

keith:
i pushed a fix that seems to handle both of these cases

oreflow:

PR #1476, opened 2026-06-10
perf: reduce allocations in the printer and string-list sorting

jbedard:
## Buildtools PR checklist

## Description

Reduce array copying to only when necessary, which seems like the exception and not the norm.


oreflow:

PR #1479, opened 2026-07-01
Add config-driven attr-policy linting

### LINKED ISSUE #1480: Feature request: config-driven attribute-policy linting (attr-policy)
## Summary

Request a generalized, config-driven buildifier warning — `attr-policy` — that enforces declarative rules about BUILD attribute values without adding new Go code per policy.

Policies would live in `.buildifier.json` and cover constraints such as:

- forbidding specific scalar values (e.g. `timeout = "eternal"`) unless allow-listed
- list membership (e.g. no `exclusive` in test `tags`)
- boolean literals (e.g. `local = True`)
- dict entries (e.g. `execution_requirements = {"no-cache": "1"}`)
- numeric bounds (e.g. `shard_count` ≤ 50)

A related but separate tool (`testpolicy`) would read test-execution metrics and emit `buildozer` commands for timeout/flaky tuning; buildifier stays purely static. (similar to Google's http://go/attt )

## Design doc

Draft design (Workstream A: `attr-policy` in buildifier; Workstream B: empirical `testpolicy` tool):

https://github.com/alexeagle/buildtools/blob/docs/attr-policy-test-tuning-design/docs/attribute-policy-and-test-tuning-design.md

Happy to upstream this if there is interest.

alexeagle:
Design doc: https://github.com/alexeagle/buildtools/blob/docs/attr-policy-test-tuning-design/docs/attribute-policy-and-test-tuning-design.md

## Summary

- Adds `attr-policy`: config-driven BUILD attribute checks via `attrPolicy` in `.buildifier.json`, plus a JSON schema for config validation.
- **Default-on** when no config is set: enforces Bazel's `shard_count` ≤ 50 on tests, and flags deprecated `licenses` / `output_licenses` usage.
- `licenses` and `output_licenses` are legacy attrs from the pre-`rules_license` era ([bazel#188](https://github.com/bazelbuild/bazel/issues/188), [bazel#7444](https://github.com/bazelbuild/bazel/issues/7444)). Bazel still accepts them silently — no analysis error, no `--incompatible_*` flag (contrast `distribs`, which had `--incompatible_no_package_distribs`), and the built-in `deprecation` attribute only warns on deprecated *targets*, not deprecated attribute names. Existing buildifier `attr-licenses` only catches the name in rule *definitions* (`.bzl`), not on BUILD targets. These default checks are the practical way to surface and finish removal.
- Design doc covers a future `testpolicy` tool (not implemented here).

Closes #1480

## Test plan

- [x] `bazel test //warn:warn_test //buildifier/config:config_test //warn/docs:docs_test`


alexeagle:
@vladmos could you at least comment whether this FR is in a shape that can be accepted upstream?

buildifier/config/attrpolicy.go
```diff
@@ -0,0 +1,227 @@
...
+	Required           bool              `json:"required,omitempty"`
+	ForbidPresence     bool              `json:"forbidPresence,omitempty"`
+	Allowlist          []string          `json:"allowlist,omitempty"`
+	Suppressible       *bool             `json:"suppressible,omitempty"`
+	Message            string            `json:"message,omitempty"`
+}
+
+func compileAttrPolicy(policy *AttrPolicy) ([]warn.AttrPolicyRuleCompiled, error) {
+	if policy == nil || len(policy.Rules) == 0 {
+		return warn.DefaultAttrPolicyRules(), nil
```
vladmos, line 41: (moving the discussion from the private messaging to Github for visibility)

Should we provide a way to let people define their own rules and use the default at the same time, without copying and maintaining the list of default rules? My suggestion was to allow something like

    "attrPolicy": {
        "keepDefault": false,  // true by default
        "rules": [...],
    }

would it be a common usecase that a repo maintainer wants to define something custom for their repo without disabling the default checks?

alexeagle: I would guess the majority of corporate monorepo users benefit from creating some enforcement of their local conventions (for example, a lexicon of allowed `tags` values so they don't proliferate meaninglessly)

Personally, I'd like to keep the full list in one place, so I would just copy and modify the default settings. But I'm happy to follow your guidance.

Some complications with `keepDefault`:
1. To understand the policy, you have to look up the defaults - but if they change from one release to the next, it's easy to make the mistake of looking up the HEAD documentation and concluding you have something enforced.
2. have to explain and test for override situation where default has one value and user has another. Especially tricky for nested values: given a default

```
      {
        "name": "no-eternal-timeout",
        "ruleKinds": ["*_test"],
        "attr": "timeout",
        "forbidValues": ["eternal"],
        "allowlist": ["//slow/..."]
      }
```

what happens if a user does

```
    "keepDefault": true,
    ...
      {
        "name": "no-long-timeout",
        "ruleKinds": ["*_test"],
        "attr": "timeout",
        "forbidValues": ["long"],
        "allowlist": ["//other/..."]
      }
```

does the forbidValues compose, or is it overridden? How would a user say that no timeout values should be forbidden? What about "allowlist"? Can users provide a name that collides with a default?

alexeagle:
(note, I'm getting my CLA resolved, that check should be green early next week)

PR #1481, opened 2026-07-13
buildifier: make formatting idempotent for compact block statements followed by comments

### LINKED ISSUE #1470: Is buildifier expected to need two passes to reach a stable format? (single-line body + following comment)
# Is `buildifier` expected to need two passes to reach a stable format? (single-line body + following comment)

## Summary

When a single-line compound statement (for example `def f(): pass`) is immediately followed by a top-level comment, `buildifier` does not reach a stable result in one pass. The first pass expands the body to a block but does not insert the blank line that the formatter wants before the following comment; a second pass inserts that blank line. The output is valid Starlark after the first pass and converges after the second, so this is a convergence question rather than a correctness one, but it means a file that `buildifier` just wrote in fix mode is reported as still needing formatting by `buildifier --mode=check`.

Is this expected, or should a single format pass already be a fixpoint?

## Reproduction

`buildifier` built from `master` at commit `84fa6c3` (post `v8.5.1`).

```sh
$ printf 'def f(): pass\n#c\n' > f.bzl

# One write/fix pass:
$ buildifier f.bzl
$ cat f.bzl
def f():
    pass
#c

# The file buildifier just wrote is reported as needing reformatting:
$ buildifier --mode=check f.bzl
f.bzl # reformat
$ echo $?
4

# A second write pass adds the blank line before the comment and converges:
$ buildifier f.bzl
$ cat f.bzl
def f():
    pass

#c

$ buildifier --mode=check f.bzl
$ echo $?
0
```

## Expected vs actual

- Expected: `buildifier <file>` followed by `buildifier --mode=check <file>` passes (one pass is a fixpoint), as it is for already-expanded input.
- Actual: the first pass produces output that the check mode (and a second fix pass) still rewrites.

## Where it shows up

The reproducer is minimal, but the same shape occurs for `if x: pass`, `for i in x: pass`, and similar single-line compound bodies followed by a comment, in both BUILD and `.bzl` modes. The practical effect is in CI setups that run `buildifier` (or a pre-commit hook) to format and `buildifier --mode=check` to verify: a freshly auto-formatted file can fail the check on the first run and only pass after a redundant second format.

If the input already has the body expanded and the blank line present, the output is stable, so the trigger is specifically the body-expansion-plus-following-comment combination, where the blank-line-before-comment insertion appears to be applied only after the body has been expanded by a prior pass.

## Notes

- The output is always valid (it re-parses), so this is about idempotence/convergence, not corrupted output.
- I checked for duplicates: #70 (closed, 2017) is a different shape (a comment moving between a list and its last element when the list is exploded); #1408 is about buildozer exit codes. I did not find an existing report for the single-line-body + following-comment case, but please point me at one if it exists.

I am happy to send a PR if a one-pass fix is wanted here.

earfman:
## Buildtools PR checklist

## Description

Fixes #1470.

buildifier needs two passes to reach a stable format when a compact block statement is followed by a line comment:

```
def f(): pass
# comment
```

Pass 1 expands the body but omits the blank line before the comment, so `buildifier --mode=check` fails on freshly formatted files (breaking format-then-verify CI); pass 2 converges. Compact `if` and `for` bodies nested in a def reproduce the same way.

**Root cause:** the two ways of writing the same logical code parse to different ASTs. With an expanded def, a dedented trailing comment goes through `extractTrailingComments` and becomes a standalone `CommentBlock` statement, which the printer separates from the block with a blank line. With a compact def, the `stmts: stmts _COMMENT '\n'` rule attaches the comment to the def's `After` list, which the printer emits flush against the body — so pass 1 output differs from the stable form.

**Fix:** in the `stmts: stmts _COMMENT '\n'` rule, treat block statements (`DefStmt`/`ForStmt`/`IfStmt`) like a nil `lastStmt`: the comment becomes a standalone `CommentBlock`, matching the expanded-form parse. `parse.y.go` regenerated with goyacc at the repo-pinned golang.org/x/tools v0.15.0 (leading "Code generated" line stripped to match the checked-in convention).

### These changes were tested using the following steps

- `printf 'def f(): pass\n# hello\n' > t.bzl && buildifier t.bzl && buildifier --mode=check t.bzl` — fails (exit 4) on main after one pass, exits 0 with this change; same for nested compact `if`/`for` shapes.
- New testdata pair `078.in` / `078.golden` covers all three shapes plus the comment-then-statement case; `TestPrintRewrite` fails on main with this pair and passes with the fix (build, bzl, and default parser types share one golden).
- `go test ./...` green (with `TEST_SRCDIR`/`TEST_WORKSPACE` set for the runfiles-based tests; the pre-existing `differ` build failure on main is unchanged).
- Differential sweep: formatting the repo's 222 testdata/BUILD/bzl corpus files plus 26 adversarial cases (same-line suffix comments, consecutive comment lines, CRLF, `--type=build`/`module`, nested compact bodies, mixed if-elif chains) produces output byte-identical to what unpatched buildifier converges to after multiple passes — the change affects convergence only, never the stable format — and every case is now stable in a single pass.

---
*This change was prepared with AI assistance and independently verified before submission: the bug was reproduced on unmodified HEAD in a fresh clone, the parser was regenerated independently, and an adversarial review confirmed the one-pass output matches the previously-converged format across file types.*

### DIFF
--- a/build/parse.y
+++ b/build/parse.y
@@ -318,7 +318,14 @@ stmts:
 	{
 		$$ = $1
 		$<lastStmt>$ = $<lastStmt>1
-		if $<lastStmt>$ == nil {
+		if $<lastStmt>$ == nil || isBlockStmt($<lastStmt>$) {
+			// Comments after a block statement (e.g. a compact `def f(): pass`)
+			// must become a standalone CommentBlock, matching how the same
+			// comment is parsed when the block is written in its expanded,
+			// indented form (see extractTrailingComments). Attaching it to the
+			// block's After list instead makes the printer emit it without the
+			// blank line that separates a block from a trailing comment, so
+			// formatting would need a second pass to become stable.
 			cb := &CommentBlock{Start: $2}
 			$$ = append($$, cb)
 			$<lastStmt>$ = cb
@@ -1219,6 +1226,19 @@ func forceMultiLineComprehension(start Position, expr Expr, clauses []Expr, end 
 		previousEnd = clauseEnd
 	}
 	return previousEnd.Line != end.Line
+}
+
+// isBlockStmt reports whether x is a statement with an indentable body
+// (def, for, or if). Line comments that follow such statements should form
+// standalone CommentBlock statements rather than being attached to the block
+// as After-comments, matching how the same comments are parsed when the
+// block is written in its expanded form (see extractTrailingComments).
+func isBlockStmt(x Expr) bool {
+	switch x.(type) {
+	case *DefStmt, *ForStmt, *IfStmt:
+		return true
+	}
+	return false
 }
 
 // extractTrailingComments extracts trailing comments of an indented block starting with the first
--- a/build/parse.y.go
+++ b/build/parse.y.go
@@ -144,7 +144,7 @@ const yyEofCode = 1
 const yyErrCode = 2
 const yyInitialStackSize = 16
 
-//line build/parse.y:1059
+//line build/parse.y:1066
 
 // Go helper code.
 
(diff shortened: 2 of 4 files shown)


earfman:
@googlebot I signed it!

vladmos (approved):
Thanks!

PR #1482, opened 2026-07-15
Allow to disable multiple warnings on one line

matthiaskramm:
This allows to disable more than one warning per "disable" comment. So, instead of writing

```python
# buildifier: disable=warning1
# buildifier: disable=warning2
```

you can now write

```python
# buildifier: disable=warning1,warning2
```

### DIFF
--- a/edit/edit_test.go
+++ b/edit/edit_test.go
@@ -865,3 +865,35 @@ a = my_macro(
 		})
 	}
 }
+
+func TestContainsDisableComment(t *testing.T) {
+	testCases := []struct {
+		comment string
+		prefix  string
+		warning string
+		want    bool
+	}{
+		{"# buildifier: disable=warning1", "buildifier", "warning1", true},
+		{"# buildifier: disable=warning1", "buildifier", "warning2", false},
+		{"# buildifier: disable=warning1,warning2", "buildifier", "warning1", true},
+		{"# buildifier: disable=warning1,warning2", "buildifier", "warning2", true},
+		{"# buildifier: disable=warning1,warning2", "buildifier", "warning3", false},
+		{"# buildifier: disable=warning1, warning2", "buildifier", "warning2", true},
+		{"# buildifier: disable=warning1,warning2 (explanation)", "buildifier", "warning1", true},
+		{"# buildifier: disable=warn", "buildifier", "warning1", false},
+		{"# buildifier: disable=warning1", "buildifier", "warn", true},
+		{"# buildozer: disable=warning1,warning2", "buildozer", "warning2", true},
+		{"# buildozer: disable=warning1,warning2", "buildifier", "warning2", false},
+	}
+
+	for i, tc := range testCases {
+		expr := &build.Ident{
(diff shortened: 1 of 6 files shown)


warn/warn.go
```diff
@@ -312,8 +312,8 @@ func runWarningsFunction(category string, f *build.File, fct func(f *build.File,
 
 // HasDisablingComment checks if a node has a comment that disables a certain warning
 func HasDisablingComment(expr build.Expr, warning string) bool {
-	return edit.ContainsComments(expr, "buildifier: disable="+warning) ||
-		edit.ContainsComments(expr, "buildozer: disable="+warning)
+	return edit.ContainsDisableComment(expr, "buildifier", warning) ||
+		edit.ContainsDisableComment(expr, "buildozer", warning)
```
vladmos: I wonder if it should be a single call to `ContainsDisableComment(expr, warning)`, and the prefixes can be iterated directly there.

matthiaskramm: I tried that earlier. But it made the code more complicated (to me), and also made `ContainsDisableComment` less generic. That said, happy to go back to that version if desired. It's your call.

edit/edit_test.go
```diff
@@ -865,3 +865,35 @@ a = my_macro(
...
+		warning string
+		want    bool
+	}{
+		{"# buildifier: disable=warning1", "buildifier", "warning1", true},
+		{"# buildifier: disable=warning1", "buildifier", "warning2", false},
+		{"# buildifier: disable=warning1,warning2", "buildifier", "warning1", true},
+		{"# buildifier: disable=warning1,warning2", "buildifier", "warning2", true},
+		{"# buildifier: disable=warning1,warning2", "buildifier", "warning3", false},
+		{"# buildifier: disable=warning1, warning2", "buildifier", "warning2", true},
+		{"# buildifier: disable=warning1,warning2 (explanation)", "buildifier", "warning1", true},
```
vladmos: I don't think it's critical, but the explanation may accidentally contain some other warning categories. Maybe just add a test for it to document the current behavior:

    {"# buildifier: disable=warning1,warning2 (warning3)", "buildifier", "warning3", true},

warn/warn.go
vladmos: I think it makes sense, because

1. in general we only need `# buildifier:` comments, `# buildozer:` is only supported for historical reasons, and I don't see why any other prefix should be added.
2. as the name suggests, it's only about `disable=` comments for warnings, so even if we decide to support a comment for a different use-case (such as `# formatter: do not format`) it won't be using `edit.ContainsDisableComment` anyway. With `edit.ContainsComment` it made more sense because that method is a bit more generic.

edit/edit_test.go
matthiaskramm: Done, good idea.

warn/warn.go
matthiaskramm: SG! Done.

vladmos (approved):
Thanks!

PR #1483, opened 2026-07-17
Allow JSONC in .buildifier.json config parsing

alexeagle:
## Summary
- Parse `.buildifier.json` as JWCC/JSONC using [`github.com/tailscale/hujson`](https://github.com/tailscale/hujson), so configs can include `//` and `/* */` comments and trailing commas.
- Add unit tests covering comment stripping edge cases and end-to-end config loading.

## Justification
Editors and many tooling ecosystems treat `.json` config files as JSONC (VS Code, Cursor, etc.), but buildifier previously used `encoding/json` directly and rejected any comments. That mismatch forces users to choose between documented/annotated configs and a config that buildifier will actually load.

Using hujson gives spec-compliant JWCC parsing (RFC 8259 JSON plus comments and trailing commas) instead of a hand-rolled comment stripper, which is safer around strings containing `//` or `/*`.

## Implications

We might not be the only parser of this file. For example buildifier plugin for an IDE might have been written with a separate parser. I don't think we need to allow the file format to be a public API however.

## Test plan
- [x] `bazel test //buildifier/config:config_test`
- [ ] Verify a `.buildifier.json` with inline comments loads via `--config`

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -30,6 +30,7 @@ use_repo(
     go_deps,
     "com_github_golang_protobuf",
     "com_github_google_go_cmp",
+    "com_github_tailscale_hujson",
     "net_starlark_go",
     "org_golang_google_protobuf",
     "org_golang_x_tools",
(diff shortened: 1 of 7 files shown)


keith (approved):

fmeum (approved):

fmeum:
Could you resolve the conflicts?

PR #1488, opened 2026-08-18
Complete support for Starlark type annotations

tetromino:
This change introduces parsing, AST, formatting, and linting support for experimental Bazel Starlark type annotations
(https://github.com/bazelbuild/bazel/issues/27370) and Buck2 type syntax:

- Type alias statements (`type OptionalDict[T, U] = dict[T, U] | None`).
- Generic `def` statements (`def foo[T, U](x: T) -> U`).
- Var statements (`x: list[int]`)
- `type_*` AST nodes to cleanly distinguish type syntax from standard expressions. In particular, this allows us to improve formatting for complex type applications, and to allow the ellipsis (`...`) only in type expressions.

Not supported yet: `cast` and `isinstance` expressions - these will be added in a follow-up PR; supporting them is a bit tricky due to the limitations of yacc's LALR(1) grammar.

Disclosure: Gemini-assisted.

TAG=agy
CONV=ae2ff586-d57d-4fd4-a22a-883aec5d0802

## Buildtools PR checklist

- [ ] I have read Google Eng Practices on [Small Changes](https://google.github.io/eng-practices/review/developer/small-cls.html), this PR either follows these guidelines or the description provides reasoning for why they can not be followed.

@brandjon @susinmotion FYI


build/print.go
```diff
@@ -620,6 +620,43 @@ func (p *printer) expr(v Expr, outerPrec int) {
...
+			p.expr(x, precLow)
+		}
+
+	case *TypeAppExpr:
+		p.expr(v.Name, precLow)
+		if v.Args != nil {
+			p.expr(v.Args, precLow)
+		}
+
+	case *TypeListExpr:
```
fmeum: **[🤖 Claude]** **Formatter emits a trailing comma the parser rejects**

The formatter prints a trailing comma in multiline type lists inside `cast`/`isinstance` arguments, but the call-argument grammar (which parses `cast(...)` as a call before the transformation) rejects that trailing comma — so buildifier's own output fails to re-parse.

```python
y = cast(list[
    int,
    str
], x)
```

formats to

```python
y = cast(
    list[
        int,
        str,
    ],
    x,
)
```

and re-parsing that output fails with `5:6: syntax error near ]`. The inverse also holds: a user-written trailing comma in a multiline cast type argument is a syntax error, even though the same list parses in an annotation position.

build/lex.go
```diff
@@ -732,6 +741,33 @@ func (in *input) order(v Expr) {
...
+	case *TypeExpr:
+		for _, x := range v.List {
+			in.order(x)
+		}
+	case *TypeAppExpr:
+		in.order(v.Name)
+		if v.Args != nil {
+			in.order(v.Args)
+		}
+	case *TypeListExpr:
```
fmeum: **[🤖 Claude]** **Comments before a closing type bracket detach and corrupt the file**

`in.order()` for `TypeListExpr`/`CastExpr`/`IsInstanceExpr` registers no closing-bracket `End` anchor (unlike `CallExpr`, which does `in.order(&v.End)`), so a comment before the closing bracket re-attaches to the next pre-order node. In an annotated assignment this splits the statement and produces unparseable output while buildifier exits 0:

```python
x: dict[
    str,
    # keep me
] = {}
```

is rewritten to

```python
x: dict[
    str,
] =
# keep me
{}
```

and the next parse fails at 4:1. Similarly, in

```python
def f(x: list[
    int,
    # c
]): pass
```

the comment silently migrates into the function body, and comments before the `)` of a multiline cast/isinstance move out of the expression. Base handles the equivalent inputs correctly.

build/parse.y
```diff
@@ -538,6 +601,28 @@ primary_expr:
 	}
 |	primary_expr '(' arguments_opt ')'
 	{
+		if ident, ok := $1.(*Ident); ok {
```
fmeum: **[🤖 Claude]** **Loaded symbol named `cast` is flagged unused; fix mode deletes the load**

Because any 2-argument call to an identifier literally named `cast`/`isinstance` is rewritten into `CastExpr`/`IsInstanceExpr`, the AST no longer contains an `Ident` for the callee. `edit.UsedSymbols` never sees the name, the unused-load check flags a genuinely used loaded symbol, and fix mode deletes the load:

```python
load(":util.bzl", "cast")

x = cast(a, b)
```

→ `Loaded symbol "cast" is unused. Please remove it.` — and `buildifier --lint=fix` removes the load, leaving `cast` undefined at Bazel load time.

build/parse.y
```diff
@@ -538,6 +601,28 @@ primary_expr:
...
+			if ident.Name == "cast" && len($3) == 2 {
+				$$ = &CastExpr{
```
fmeum: **[🤖 Claude]** **cast/isinstance rewrite fires in BUILD files and on keyword arguments**

The rewrite also fires in contexts where it shouldn't:

- In a BUILD file, `cast(name = "foo", srcs = ["a"])` produces a target invisible to `Rules()` — `f.Rules("")` returns 0 and buildozer reports the target as not found.
- `cast(a = 1, b = 2)` becomes a `CastExpr` whose `Type` is an `AssignExpr` — a corrupt AST that also pollutes `GetTypes`.
- A statement-level call like `cast(x, "before")` inside a function draws a false `Expression result is not used.` warning (`CallExpr` is exempt in noEffectStatementsCheck; `CastExpr` isn't).
- `ForceCompact` (from `forceCompact($2, $3, $4)`) is never set on the new nodes, so a compact top-level `cast(a, b)` in BUILD mode is forced multiline.

Restricting the transformation to positional 2-argument calls in type-syntax contexts (or keeping the callee `Ident` in the AST) would avoid this class of issue.

fmeum (commented):
**[🤖 Claude]** This review was performed by Claude (Claude Code), at fmeum's request; fmeum triaged the findings but Claude wrote every comment. Each finding was reproduced locally at head 80211fc against base 5d08cfd — none are speculative. Inline comments cover the findings that anchor to diff lines; the ones below touch code this PR doesn't modify but whose behavior it changes.

### build/parse.y:1630 — Tuples in isinstance are silently rewritten to lists

`toTypeExpr` rewrites every non-empty `TupleExpr` into a `TypeListExpr`, so a tuple literal passed to `cast`/`isinstance` is silently reformatted into a list, changing program semantics:

```python
if isinstance(x, (int, str)):
```

formats to

```python
if isinstance(x, [int, str]):
```

For a user-defined `isinstance` helper doing membership/type checks, a tuple argument becomes a list.

### warn/warn_control_flow.go:729 — Bare `x: int` statement treated as a use, not a declaration

A bare var statement `x: int` (a `TypedIdent` used as a statement) is treated as a use rather than a declaration: `findUninitializedVariables` reports the annotation itself as reading an uninitialized variable, and the unused-variable collector marks the name as used.

```python
def f():
    x: int
    x = 1
    return x
```

→ false `Variable "x" may not have been initialized.` on the annotation line; conversely, `x: int` marking `x` as used suppresses legitimate unused-variable findings.

### warn/warn_control_flow.go:563 — Type alias names are never treated as definitions by the variable checks

The redefined-variable and unused-variable checks never treat `TypeAliasStmt` as defining its `Name` (their statement scans only match `AssignExpr`):

- `type A = int` followed by `type A = str` and `A = 5` produces no redefined-variable finding.
- In

```python
def f():
    a = 1
    type a = int
    return 2
```

no warning is emitted at all: `WalkOnce` exposes the alias `Name` as an ordinary `Ident`, which counts as a *use* of `a` and hides the genuinely unused `a = 1`.

### bzlenv/bzlenv.go:181 — bzlenv never declares type parameters or bare annotated variables

`declareLocalVariables` handles only `AssignExpr`/`IfStmt`/`ForStmt`, so a bare `x: int` statement inside a function declares nothing — while the same statement at top level *is* declared via `declareGlobals`, an asymmetric resolution. And `DefStmt`/`TypeAliasStmt` `TypeParams` are never declared in any scope, so `def foo[T](x: T) -> T` leaves `T` unresolvable for every bzlenv-based consumer (warn_bazel_api, type inference, naming checks), which misclassify it as an undeclared global.

build/parse.y
```diff
@@ -1325,3 +1549,80 @@ func getLastBody(stmt Expr) *[]Expr {
...
 }
+
+// toTypeExpr converts an Expr (such as IndexExpr, BinaryExpr, ListExpr, DictExpr, TupleExpr)
+// into a type_expr (TypeAppExpr, TypeExpr, TypeListExpr, TypeDictExpr).
+func toTypeExpr(e Expr) Expr {
+	if e == nil {
+		return nil
+	}
+	switch e := e.(type) {
+	case *IndexExpr:
```
fmeum: **[🤖 Claude]** **Exponential parse time from double toTypeExpr recursion**

The `IndexExpr` case calls `toTypeExpr(e.Y)` twice — once to test whether the conversion succeeds and once to build the fallback — doubling the work per nesting level, i.e. exponential parse time for nested subscripts inside cast arguments.

Measured on `x = cast(a[a[...b...]], y)`: depth 18 parses in 20 ms, depth 22 in 331 ms, depth 24 in 1.3 s (2× per level); ~40 levels hangs buildifier indefinitely — a DoS for CI running buildifier on untrusted input. Calling `toTypeExpr(e.Y)` once and reusing the result fixes it.

build/walk.go
```diff
@@ -143,6 +143,62 @@ func WalkOnce(v Expr, f func(x *Expr)) {
 	case *AssignExpr:
 		f(&v.LHS)
 		f(&v.RHS)
+	case *TypeAliasStmt:
```
fmeum, line 144: **[🤖 Claude]** **WalkOnce drops Edit() results on TypeAliasStmt; unchecked assertions panic**

Two issues in the new traversal code:

1. The `TypeAliasStmt` case passes the callback pointers to local copies of `Name` and `TypeParams` without writing the results back (unlike the sibling `DefStmt` case), so `Edit()` replacements are silently dropped: renaming every `Ident` on `type Foo[T] = dict[T, Foo]` renames only the right-hand side and leaves the declaration untouched.
2. The new write-backs elsewhere use unchecked type assertions: an `Edit` on `def foo(T: int)` that substitutes a `StringExpr` for the parameter `Ident` panics with `interface conversion: build.Expr is *build.StringExpr, not *build.Ident` at `v.Ident = ident.(*Ident)` — a crash class that didn't exist before, because `TypedIdent` children weren't walked pre-PR.

build/print.go
```diff
@@ -857,6 +894,16 @@ func (p *printer) expr(v Expr, outerPrec int) {
 	case *DefStmt:
 		p.printf("def ")
 		p.printf(v.Name)
+		if len(v.TypeParams) > 0 {
```
fmeum: **[🤖 Claude]** **Comments inside type parameter lists are displaced**

`DefStmt`/`TypeAliasStmt` print type parameters via a bare printf loop instead of `p.seq`, so comments attached inside `[T, U]` have no anchor and are displaced:

- `def foo[T,  # comment` (with `U](a):` on the next line) → the comment moves inside the parens, above `a`.
- A multiline type parameter list with a leading `# a comment` line → the comment lands at column 0 inside the brackets and the parameter is misindented to 8 spaces.
- `type Foo[T,  # comment` → the comment moves to the end of the statement.

All mangled forms persist on reformatting.

build/parse.y
```diff
@@ -513,6 +569,13 @@ primary_expr:
 	{
 		$$ = $1
 	}
+|	_ELLIPSIS
```
fmeum: **[🤖 Claude]** **`...` now parses as a general expression in every dialect**

`...` is lexed everywhere and accepted as a general `primary_expr` `LiteralExpr`: `x = ...`, `f(...)`, and `[...]` all parse without error in every dialect, including BUILD files — each was a syntax error before this PR — and the post-parse rejection of stray ellipses mentioned in the PR description doesn't appear to exist.

The token also gets two AST representations (`LiteralExpr` in expression position, `EllipsisExpr` in type arguments), so every consumer has to special-case both — and warn/types.go classifies the `LiteralExpr` form as Float, enabling bogus operator-type warnings.

build/parse.y
```diff
@@ -503,6 +523,42 @@ small_stmt:
 		}
 	}
 
+type_alias_stmt:
+	ident ident type_params_opt '=' type_expr
+	{
+		if $1.(*Ident).Name != "type" {
+			yylex.Error("syntax error near " + $1.(*Ident).Name)
```
fmeum: **[🤖 Claude]** **Misparsed alias statements report the wrong error position**

The manual validation in the `type_alias_stmt` action (`yylex.Error("syntax error near " + $1.(*Ident).Name)`) fires only after the whole rule is reduced, so a misparse reports a later position and a different token than before: `foo bar = baz` now reports `2:1: syntax error near foo` (past the end of the line, naming the first token) where base reports `1:8: syntax error near bar`. Editor and CI diagnostics point at the wrong place for every such error.

build/print.go
```diff
@@ -929,6 +976,29 @@ func (p *printer) expr(v Expr, outerPrec int) {
...
+			p.printf("[")
+			for i, param := range v.TypeParams {
+				if i > 0 {
+					p.printf(", ")
+				}
+				p.expr(param, precLow)
+			}
+			p.printf("]")
+		}
+		p.printf(" = ")
```
fmeum: **[🤖 Claude]** **Multi-line type aliases format with a double space before the backslash**

`TypeAliasStmt` printing emits `" = "` and `breakline()` then adds another space before the backslash continuation, so every multi-line alias formats to `=  \` with a double space — and since `type_expr` accepts neither a newline after `=` nor parentheses, every alias too long for one line hits it:

```python
type T = \
    int | float
```

formats (idempotently, verified byte-for-byte) to

```python
type T =  \
    int | float
```

build/parse.y
tetromino: Proper support for cast/isinstance deferred to follow-up PR.

build/parse.y
tetromino: cast/isinstance deferred to follow-up PR

build/parse.y
tetromino: cast/isinstance deferred to follow-up PR

build/parse.y
tetromino: This was very much by design :)

To support cast/isinstance, we had to support parsing `...` everywhere - and then, after converting cast/isinstance using toTypeExpr, catch any remaining `...` in non-type code.

In any case, we don't need ellipses as primary_expr until the follow-up PR.

build/print.go
tetromino: Good catch! Will fix in follow-up when I re-add cast expressions. (I think the only reasonable fix would be to check for trailing commas in non-type syntax in a post-parsing step, same as checking for ellipses literals.)

build/lex.go
tetromino: I think the conclusion is that TypeList and TypeDict were a bad idea: they don't add any value in the current state of the code, and are a route for bugs to enter due to differences in behavior from normal ListExpr/DictExpr. So I am removing them and using normal ListExpr/DictExpr instead.

And I'm updating TypeApplication to print like a CallExpr.

Fixed by a179a5888dcb5e0e85163530b6d782ec57627ed0

build/walk.go
tetromino: For the type params of TypeAliasStmt and DefStmt, I think to make them editable, we have to pass them as a single list node (otherwise, the editing function cannot add/remove type params). This means, naturally, storing them as a ListExpr. (Which incidentally helps with fixing the formatting of comments in type params!)

As for the *identifier* of TypeAliasStatement or TypedIdent, an editor function that changes the node to a non-identifier breaks Starlark syntax, and we *absolutely should* panic here, although of course with a nicer message. (Better would be to report a nice structured error, but alas WalkOnce is a stable API as far as I can tell, and has no error reporting provision...)

Fixed by d8d781a1262852d9cc51a9e41e99881ee94aacb1 and a01af04630962a324bb64e5dca4b284466ce64fe

build/print.go
tetromino: Fixed as a side effect of d8d781a1262852d9cc51a9e41e99881ee94aacb1

build/parse.y
tetromino: Nice catch! Fixed by b13f47f6a2ece3210634f034f9a28538e493aa48

build/print.go
tetromino: Fixed by b13f47f6a2ece3210634f034f9a28538e493aa48

fmeum (commented):
**[🤖 Claude]** Round 2, again authored by Claude (Claude Code) at fmeum's request, with fmeum triaging the findings. The first-round fixes hold up: removing cast/isinstance eliminated that entire class, and I verified the walk/bzlenv/printing/lint fixes landed (goyacc regeneration matches parse.y with 0 conflicts, and a corpus differential over the repo shows byte-identical formatting outside the intended grammar change). The findings below are new issues or edge-case gaps in the fix commits, verified by executing the PR code at head 7a57b19 (exceptions noted inline).

The following findings touch lines outside the diff hunks:

### build/parse.y:390 — Return types still parse as `test`, diverging from parameter annotations

`def` return types still parse as `test` instead of `type_expr`, so the two annotation positions diverge:

- `def h() -> struct[{"a": int}, ...]:` is a parse error while the identical annotation on a parameter parses.
- `x: T` and `-> T` for the same type produce different AST node shapes (`IndexExpr`/`BinaryExpr` vs `TypeAppExpr`/`TypeExpr`), so type-aware tooling must handle both, and buildozer edits that move an annotation between positions break.
- A `TypeAppExpr` placed programmatically in `DefStmt.Type` can print unparseable output.

### build/parse.y:1319 — isSimpleExpression not extended for type nodes: compact type applications explode in BUILD mode

`isSimpleExpression` doesn't handle the new type nodes (`EllipsisExpr`, `TypeAppExpr`, list/dict type args), so `forceCompact` is false for any type application with non-ident args, and in BUILD/WORKSPACE mode `useCompactMode` explodes an originally-compact application one-arg-per-line:

- BUILD file: compact `type T = Callable[[int], str]` reformats to multiline, while `type T = dict[str, int]` (all-ident args) stays compact.
- The same input in a .bzl file keeps the compact form.

079.build.golden encodes this divergence (a compact `Callable` becomes 7 lines), so it may be partially intentional — but the inconsistency between `Callable[...]` and `dict[...]`, and between BUILD and .bzl for the same construct, looks like a mechanism gap rather than a policy.

warn/warn_control_flow.go
```diff
@@ -456,7 +478,14 @@ func unusedVariableCheck(f *build.File, root build.Expr) (map[string]bool, []*Li
 						// even if not used.
 						suppressedWarnings[ident.Name] = true
 					}
+
+					// Parameters' types are defined in type params or in the outer scope.
+					typedIdent, ok := param.(*build.TypedIdent)
```
fmeum: **[🤖 Claude]** **unused-variable falsely flags types used in defaulted/starred parameter annotations**

The `param.(*build.TypedIdent)` assertion misses the `*AssignExpr` (`x: T = 1`) and `*UnaryExpr` (`*args: T`, `**kwargs: T`) parameter shapes, so their annotations are never passed to `collectUsedTypeSymbols` and a type used only there is reported as unused:

```python
type _Num = int | float

def foo(x: _Num = 1):
    pass
```

→ `Type "_Num" is unused. Please remove it.` (verified by running the PR's warn package); same false positive for `def foo[T](x: T = 1)` and `def foo[T](*args: T)`. Users following the warning would delete a used declaration.

build/print.go
```diff
@@ -1131,6 +1166,19 @@ func (p *printer) seq(brack string, start *Position, list *[]Expr, end *End, mod
 	}
 }
 
+// optionalTypeParams formats a ListExpr of a type alias's or generic function's type params,
+// or does nothing if given a nil.
+func (p *printer) optionalTypeParams(typeParams Expr) {
```
fmeum, line 1183: **[🤖 Claude]** **Suffix comment silently deleted on a generic def with an empty parameter list**

The comment attaches to the TypeParams `ListExpr`, but `optionalTypeParams` prints it via `p.seq` directly (not `p.expr`), which never emits the node's own suffix comments:

```python
def f[T]():  # c
    pass
```

formats to `def f[T]():` — the comment is permanently lost on every buildifier run (verified). Contrast: plain `def f():  # c` preserves the comment by moving it above the def, and `type X[T] = int  # c` keeps it.

build/parse.y
```diff
@@ -503,6 +523,49 @@ small_stmt:
 		}
 	}
 
+type_alias_stmt:
```
fmeum: **[🤖 Claude]** **Error-position regression for invalid input beginning `ident ident`**

The `type_alias_stmt` production shifts the second identifier expecting `[...]= type`, so when no `=` follows, the error is reported at the *next* token with an empty "near" string instead of at the offending identifier:

- `foo bar\n` → `t.bzl:2:1: syntax error near ` (empty, pointing past the line); base reports `t.bzl:1:8: syntax error near bar`.
- Running buildifier on a non-Starlark file hits the same path: `package main` → error at `2:1` instead of base's `1:13: syntax error near main`.

Verified against both base and PR head.

build/parse.y
```diff
@@ -1063,6 +1126,144 @@ for_clauses_with_if_clauses_opt:
...
+		}
+	}
+|	_ELLIPSIS
+	{
+		$$ = &EllipsisExpr{
+			Pos: $1,
+		}
+	}
+
+type_dict:
```
fmeum, line 1228: **[🤖 Claude]** **Existing dict lints edit type annotations: unsorted-dict-items reorders struct type fields**

`type_dict` produces a plain `*DictExpr`, so existing dict lints reach into type annotations. Verified by running FixWarnings on the PR tree:

```python
type _Big = struct[{"z": int, "a": str}, ...]
```

with `--lint=fix` and unsorted-dict-items enabled rewrites the annotation to `{"a": str, "z": int}` — the linter silently edits a type declaration, where field order can be meaningful (e.g. record constructors).

build/print.go
```diff
@@ -882,6 +897,7 @@ func (p *printer) expr(v Expr, outerPrec int) {
 	case *DefStmt:
 		p.prints("def ")
 		p.prints(v.Name)
+		p.optionalTypeParams(v.TypeParams)
 		p.seq("()", &v.StartPos, &v.Params, nil, modeDef, v.ForceCompact, v.ForceMultiLine)
```
fmeum: **[🤖 Claude]** **Multiline type params break open a compact parameter list**

The parameter seq is anchored at `Function.StartPos` (the `def` keyword) rather than the `(` position, so multiline type params make `useCompactMode` see the first parameter on a later line than `def` and spuriously reformat a compact parameter list as multiline:

```python
def foo[
    T,
    U,
](x: T) -> U:
```

reformats with the compact `(x: T)` broken open (verified by running build.Format on the PR tree). `StartPos` was a safe anchor only while nothing could intervene between the name and `(`.

warn/warn_control_flow.go
```diff
@@ -561,33 +636,56 @@ func unusedVariableWarning(f *build.File) []*LinterFinding {
...
-		if !ok {
-			continue
-		}
-		if !definedSymbols[left.Name] {
-			definedSymbols[left.Name] = true
-			continue
-		}
+		switch s := s.(type) {
+		case *build.AssignExpr:
+			left, ok := s.LHS.(*build.Ident)
```
fmeum: **[🤖 Claude]** **Type-redefinition detection misses TypedIdent and def redefinitions**

The `s.LHS.(*build.Ident)` assertion fails for typed assignments and var statements, and `DefStmt` isn't tracked at all, so:

- `type T = int` followed by `T: int = 5`, `T: int`, or `def T(): ...` produces no finding (verified),
- while `type T = int` + `T = 5` correctly reports `Variable "T" has already been defined as a type.`

The inconsistency undermines the warning's own "incompatible with static type checking" rationale.

warn/warn_control_flow.go
```diff
@@ -561,33 +636,56 @@ func unusedVariableWarning(f *build.File) []*LinterFinding {
...
+			if !ok {
+				continue
+			}
+			if _, ok := definedSymbols[left.Name]; !ok {
+				definedSymbols[left.Name] = Variable
+				continue
+			}
+
+			if s.Op == "+=" && (types[s.LHS] == List || types[s.RHS] == List) {
```
fmeum: **[🤖 Claude]** **`T += [1]` after a type alias escapes the redefinition check**

The `+=` list-append allowance runs before the Type-vs-Variable check, so augmenting a name defined by a type alias is silently accepted:

```python
type T = int

T += [1]
```

→ no redefined-variable finding (verified), although `T` is redefined from a type to a list — exactly what the new code flags as incompatible with static type checking for plain `T = 42`. The allowance's rationale (appending is not reassignment) can't apply to a type; checking the type-map first fixes it.

build/walk.go
```diff
@@ -143,6 +143,40 @@ func WalkOnce(v Expr, f func(x *Expr)) {
...
```
fmeum, line 144: **[🤖 Claude]** **WalkPointers hands out pointers to temporaries for alias/var-statement names**

`WalkOnce` passes callback pointers to function-local temporaries for `TypeAliasStmt.Name` and `TypedIdent.Ident`. Writes made *during* the walk are copied back, but pointers saved from `WalkPointers` and written *after* the walk — the standard `LinterReplacement` pattern in warn (`*r.Old = r.New` in warn.go, used by warn_bazel_api.go et al.) — are silently lost for these two slots, producing half-applied fixes for any fixer that touches an alias or var-statement name. No in-tree fixer hits this today, but the API contract (pointer into the AST) is violated, so it's a trap for the next lint with an autofix.

Relatedly, an external `build.Edit` user replacing one of these idents with e.g. a `DotExpr` now panics — I saw a01af04 makes that deliberate, but it converts a previously total API into a partial one; degrading by ignoring the invalid replacement would keep external callers crash-free.

build/print.go
```diff
@@ -1023,6 +1055,9 @@ func (p *printer) useCompactMode(start *Position, list *[]Expr, end *End, mode s
 			if isDifferentLines(&start, previousEnd) {
 				return false
 			}
+			if mode == modeTypeList && len(*list) > 1 && start.Line != 0 && end.Line != 0 && start.Line != end.Line {
```
fmeum: **[🤖 Claude]** **The modeTypeList multiline-element condition is dead code**

modeTypeList sequences are always type-parameter lists whose elements are single-line `Ident`s, so `start.Line != end.Line` can never hold. The multiline goldens this clause appears to serve are actually handled by the Before-comment check and `ForceMultiLine`. It adds a fourth clause to an already dense predicate and suggests per-element multiline detection that never occurs — delete it (or assert the invariant).

build/lex.go
```diff
@@ -297,17 +297,47 @@ func (in *input) parse() (f *File, err error) {
...
 
+	if err := checkTypeAlias(in.file); err != nil {
+		return nil, err
+	}
+
 	return in.file, nil
 }
 
+// checkTypeAlias checks that type alias statements occur only at top-level.
+func checkTypeAlias(file *File) error {
```
fmeum, line 308: **[🤖 Claude]** **checkTypeAlias walks the full AST of every parsed file**

`checkTypeAlias` runs a full-tree `WalkInterruptable` over every expression of every parsed file just to find nested `TypeAliasStmt` nodes, and after reporting an error it still visits every remaining node. This puts a whole-AST walk (all idents, strings, call args) in the hot parse path for a statement-level property — measurable on large BUILD files. `build.WalkStatements` (as warn uses) visits only statement nodes and supports the same early stop; alternatively, track nesting during the parse itself.

build/parse.y
```diff
@@ -41,6 +41,7 @@ package build
 	// partial syntax trees
 	expr      Expr
 	exprs     []Expr
+	idents    []*Ident
```
fmeum: **[🤖 Claude]** **Dead %union field and truncated comment**

Two leftovers:

- The `idents []*Ident` `%union` field is never referenced by any `%type` or rule (`type_params_idents` uses `<exprs>`).
- The `type_alias_stmt` error branch contains a comment `// Match ` cut off mid-sentence, which obscures the intent of the soft-keyword error path.

build/parse.y
tetromino: Not a bug. Reordering the keys doesn't change the semantics of Bazel's `struct` type annotation; and if some other type constructor is sensitive to the order of keys in a dictionary argument, I would argue that it is a bug in that type constructor :)

build/parse.y
tetromino: Fixed by d2cf96d4afd4ddc2554e12f074395718c79f1bef

build/lex.go
tetromino: Fixed by 236e427c06c29b915ecb5167183f9deac362d4bc

build/print.go
tetromino: Fixed by 1e3f1475615458219ee109c89c6c0378d37a4a0d

build/walk.go
tetromino: Meh. I still think that allowing syntactically invalid edits is madness, but Claude has a point that `walk` might save pointers after walking.... Fixed by 8c09447861f37b62242806b835a4778c64565ad4

warn/warn_control_flow.go
tetromino: Fixed by b2fcd8a908aba0e0e72f7b01d79622cc653baf85

build/print.go
tetromino: Oh, good catch. Fixed by be38048ada7944bea8fccfac0f508e3d8619c3fa

build/parse.y
tetromino: Interesting point. Fixed by 9a3b44b3d45bf89c75f0e8314f3adc7d048e7e83

build/print.go
tetromino: Great catch! Fixed by f37a5c202a29c96cf0e77f57ded957c33f20f269

warn/warn_control_flow.go
tetromino: Fixed by e219612ea538ac2a4ad4bd6ac648ef567efaa7d9

warn/warn_control_flow.go
```diff
@@ -561,33 +654,57 @@ func unusedVariableWarning(f *build.File) []*LinterFinding {
...
-		if !ok {
-			continue
-		}
-		if !definedSymbols[left.Name] {
-			definedSymbols[left.Name] = true
-			continue
-		}
+		switch s := s.(type) {
+		case *build.AssignExpr:
+			left, ok := s.LHS.(*build.Ident)
```
fmeum: **[🤖 Claude]** **Direct `LHS.(*build.Ident)` assertions are blind to TypedIdent — one shared helper would fix seven sites**

This is the dominant remaining issue, and it's systemic rather than local: an annotated assignment's LHS is now a `*TypedIdent` wrapping the ident. Code paths that go through `bzlenv.CollectLValues` or `build.GetParamIdent` handle that correctly, but every site that matches assignments with a direct `s.LHS.(*build.Ident)` assertion silently skips annotated assignments. Seven sites are affected — each reproduced at head d2cf96d against an unannotated baseline that behaves correctly:

1. **`warn/warn_control_flow.go:667`** (redefinedVariableWarning, this line): `x = 1` + `x: int = 2` → no warning; `x: int = 1` + `x += 1` → no warning; and `type T = int` + `T: int = 5` bypasses the new type-redefinition message — annotated globals are exactly the style this typed-code warning targets.
2. **`warn/types.go:182`** (DetectTypes): a typed assignment never records the variable's type, so `x: int = 3` + `y = x / 2` loses the `integer-division` warning that `x = 3` gets — annotating a variable makes the type-driven lints forget its type.
3. **`warn/warn_macro.go:153`**: `my_rule: MyType = rule()` isn't registered, so `unnamed-macro` disappears.
4. **`warn/warn_bazel_operation.go:166`**: an annotated depset variable loses `overly-nested-depset`.
5. **`edit/edit.go:1056`** (UsedSymbols; also feeds `edit/fix.go:352`): the `as.LHS == expr` skip misses because LHS is the TypedIdent wrapper while WalkOnce visits the inner ident, so the LHS counts as a *use*: `load(":a.bzl", "x")` + `x: int = 1` → no `unused-load` warning; automated load pruning silently stops working on annotated files.
6. **`edit/buildozer.go:1195`** (`-edit-variables`): with `V: list = ["a"]`, an add rewrites the attribute to `srcs = V + ["b"]` instead of editing `V` in place as it does for `V = ["a"]`.
7. **`edit/bzlmod/bzlmod.go:248`**: with `ext: T = use_extension("//:ext.bzl", "ext")`, `buildozer 'use_repo_add ext b'` exits 2 with `no use_extension assignment to variable "ext" found`.

Rather than seven point patches, a single shared lvalue-unwrap helper (e.g. `build.AssignLHSIdent(expr) *Ident`, or reusing the CollectLValues machinery) applied at these sites fixes the family at once — and gives future call sites an obviously-correct spelling. Site 5 additionally needs the `as.LHS == expr` identity check to match through the wrapper.

warn/warn_control_flow.go
```diff
@@ -473,6 +502,39 @@ func unusedVariableCheck(f *build.File, root build.Expr) (map[string]bool, []*Li
 					usedSymbolsFromOuterScope[ident.Name] = true
```
fmeum, line 509: **[🤖 Claude]** **A type param referenced only in a parameter default is falsely flagged unused**

Default-value idents go unconditionally to `usedSymbolsFromOuterScope` without consulting `definedTypes`, so:

```python
def foo[T](x = T):
    print(x)
```

→ `unused-variable: Type "T" is unused.` although `T` is referenced (reproduced at head). With a same-named top-level alias, attribution also flips between the two paths: the default variant flags the param while the annotation variant (test-blessed) flags the outer alias — incoherent in both directions.

warn/warn_control_flow.go
```diff
@@ -519,39 +584,67 @@ func unusedVariableCheck(f *build.File, root build.Expr) (map[string]bool, []*Li
...
-			makeLinterFinding(ident, fmt.Sprintf(`Variable %q is unused. Please remove it.`, ident.Name)))
 	}
+}
 
-	for name, def := range definedFunctions {
+// Format and append linter findings for unused variables to the findings slice.
+//
+// Helper function for unusedVariableCheck.
+func appendUnusedVariableLinterFindings(findings []*LinterFinding, definedSymbols map[string]build.Expr, symbolKind string, usedSymbols map[string]bool, suppressedWarnings map[string]bool, ignoreTopLevel bool) []*LinterFinding {
+	for name, expr := range definedSymbols {
```
fmeum, line 643: **[🤖 Claude]** **Same-line "Type X is unused" findings come out in nondeterministic order**

`definedTypes` is a map and the downstream `sort.Slice` is unstable and keyed on `Line` only — and same-line groups are the norm for `def f[A, B, C]`. Reproduced at head: 20 identical runs on `def _f[A, B, C](x: int) -> None` with unused params produced 3 distinct output orderings (15/3/2 split), which breaks CI output diffing. Iterating the `CollectLValues` slice instead of the map (or extending the sort key to the column) makes it deterministic.

build/print.go
```diff
@@ -642,6 +615,21 @@ func (p *printer) expr(v Expr, outerPrec int) {
 		p.prints(": ")
 		p.expr(v.Type, precLow)
 
+	case *TypeExpr:
```
fmeum: **[🤖 Claude]** **TypeExpr discards '|' positions, collapsing wrapped unions and mis-indenting comments**

parse.y's `type_expr` action ignores `$2`, so `TypeExpr` loses the operator positions/`LineBreak` that `BinaryExpr` preserved. Two consequences, both reproduced at head vs base:

- A long union the user wrapped at `|` collapses onto one >80-column line.
- `def f(x: int |` with a comment on the continuation line prints the comment at the statement margin (column 0) instead of aligned under the operand.

build/lex.go
fmeum, line 308: **[🤖 Claude]** **Printer has no guard for ASTs the parser would reject (nested alias, empty type application)**

The top-level-only restriction for type aliases exists only as this post-parse check, with no matching printer guard — so a programmatically built AST with a `TypeAliasStmt` inside a def body passes `build.Format` and produces output buildifier then refuses to parse (`type alias not at top level`). Likewise `TypeAppExpr{Args: nil}` prints as `T[]`, which is a syntax error on re-parse (print.go:626).

Given the position from the walk.go thread — syntactically invalid states should fail loudly rather than flow through — these two spots seem like they want the same treatment `optionalTypeParams` already gets (panic on the invalid state), rather than silently emitting unparseable output.

warn/warn_control_flow.go
```diff
@@ -380,6 +383,14 @@ func extractIdentsFromStmt(stmt build.Expr) (assigned, used map[*build.Ident]boo
 			}
 			used[expr] = true
 
+		case *build.TypedIdent:
```
fmeum, line 384: **[🤖 Claude]** **A dead bare declaration is invisible to every lint — intentional?**

A bare declaration `x: int` that is never assigned or used produces zero findings: its ident is excluded from the unused/uninitialized/no-effect analyses, while dead `x = 1` gets `unused-variable` (reproduced at head).

This may well be intentional — a declaration assigns nothing, so 'unused variable' arguably doesn't apply — but it means dead declarations accumulate silently. Flagging so it's a deliberate decision rather than a side effect of the blocked-nodes mechanism.

build/rewrite.go
```diff
@@ -1404,7 +1404,7 @@ func editFloats(f *File, _ *Rewriter) {
 		if !ok {
 			return
 		}
-		if !strings.ContainsRune(l.Token, '.') {
+		if l.Token == "..." || !strings.ContainsRune(l.Token, '.') {
```
fmeum: **[🤖 Claude]** **The editFloats `"..."` guard is dead code**

The lexer folds `...` into `_ELLIPSIS` before float assembly, and `LiteralExpr` is only built from the three `_INT`/`'.'` number productions — so `l.Token == "..."` can never be true here (two independent checks confirmed unreachability). The guard anchors a false invariant that will mislead the next editor of float rewriting; remove it, or replace it with a comment explaining why the case cannot occur.

build/parse.y
```diff
@@ -503,6 +523,60 @@ small_stmt:
 		}
 	}
 
+// Split out of type_alias_stmp to improve error handling:
```
fmeum: **[🤖 Claude]** **Leftovers: `type_alias_stmp` typo, and the truncated `// Match ` comment survived**

d2cf96d removed the dead `idents` union field, but two leftovers from that thread remain at head:

- The rule comment reads `type_alias_stmp` (typo), which defeats searching for the split-rule rationale.
- The dangling `// Match ` comment above the error branch is still cut off mid-sentence (with trailing whitespace that `git apply` flags), so the intent of the soft-keyword error path is still undocumented.

build/parse.y
```diff
@@ -503,6 +523,60 @@ small_stmt:
...
+// on two idents in sequence, e.g. "foo bar", we want to emit the error at "bar".
+type_alias_stmt_start:
```
fmeum, line 529: **[🤖 Claude]** **type_alias_stmt_start builds a throwaway node the next reduction discards**

`type_alias_stmt_start` allocates a `TypeAliasStmt` that `type_alias_stmt` immediately discards and rebuilds field-by-field, unlike the file's established in-place-mutation pattern (`def_header_type_opt`: `$1.Type = $3; $$ = $1`). Any future `TypeAliasStmt` field has to be remembered in the second action's copy list or it is silently dropped between the two reductions — `TypePos`/`Name` already have to be threaded by hand today. Mutating and forwarding the node from the first action removes the trap.

fmeum (commented):
**[🤖 Claude]** Round 3, authored by Claude (Claude Code) at fmeum's request, with fmeum triaging. All round-2 findings were verified fixed at head d2cf96d (parser tables regenerate byte-identically with 0 conflicts, comments on the new constructs round-trip idempotently, and ~1000-case fuzzing produced no crashes or idempotence failures). Per the thread replies, the unsorted-dict-items point is dropped as intended behavior, and BUILD-mode multiline formatting of type applications is accepted as consistent with call formatting.

This round's headline is a single systemic issue — TypedIdent-blind `LHS.(*build.Ident)` assertions — reported once below with all affected sites, rather than as per-site comments. The rest are smaller lint gaps and cleanups. Everything was reproduced at head except where a finding says otherwise.

Findings on lines outside the diff hunks:

### warn/warn_control_flow.go:97 — missing-return never consults the new return annotation

`missingReturnValueWarning` computes `hasNonEmptyReturns` solely from `ReturnStmt.Result`, so an annotated function that returns nothing goes unflagged:

```python
def f() -> int:
    print(1)
```

and

```python
def f(x) -> int:
    if x:
        return
```

produce no missing-return findings (reproduced at head) — a gap squarely in this PR's own type-aware-lint domain: `-> int` is a stronger statement of intent than any inferred non-empty return.

---
https://claude.ai/code/session_013eQCUiidFetaGHqfWkotSv

warn/warn_control_flow.go
tetromino: This is intentional; type parameters can only be used in annotations, not in the value domain.

warn/warn_control_flow.go
tetromino: Good point, fixed by 748b7be6ee33f5fdc5fa2f506a253bd7defb6905

warn/warn_control_flow.go
tetromino: Pre-existing issue, but sure, fixed by 75264e7c6c18c053e4f55e62a7be9fc8d44f320d

build/lex.go
tetromino: Intentional. Same reason why printer already has no guard for nonsense like AssignStmt with a DefStmt on the LHS etc. The user of the edit library must themselves ensure their edits don't violate Starlark syntax,

build/print.go
tetromino: Good point, fixed by ece03c1a0de93250daadb0bbe8983d5eca330420

warn/warn_control_flow.go
tetromino: Interesting observation. It's true that although `extractIdentsFromStmt` returns `assigned`, the only place we use that value is when searching for _declared_ (as contrasted with _assigned_) symbols. Fixed by 8270b125e826230ec25c4cfcf4f12f88eaf3da05

build/parse.y
tetromino: Fixed by bcc1629e977bee666682d06c3be5f3c2c7295e5c

build/parse.y
tetromino: Fixed by bcc1629e977bee666682d06c3be5f3c2c7295e5c

build/rewrite.go
tetromino: Fixed by bcc1629e977bee666682d06c3be5f3c2c7295e5c

edit/edit.go
```diff
@@ -1053,7 +1053,7 @@ func UsedSymbols(stmt build.Expr) map[string]bool {
 		// Check if we are on the left-side of an assignment
 		for _, e := range stack {
 			if as, ok := e.(*build.AssignExpr); ok {
-				if as.LHS == expr {
+				if ident, ok := as.LHSIdent(); ok && ident == expr {
```
fmeum: **[🤖 Claude]** **UsedSymbols counts declaration identifiers as uses, so unused-load misses loads shadowed by declarations**

The exclusion check in the stack walk only recognizes an enclosing `AssignExpr`, so a bare var statement's ident (`x: int`), a `TypeAliasStmt`'s name, and def/alias type-param idents are all counted as *uses* — and `unusedLoadWarning` plus the unusedLoads fix (edit/fix.go:352) silently keep genuinely dead loads. This is the mirror image of the round-2 bug (which deleted used loads); now dead ones are kept.

Verified at head: `load(":a.bzl", "x")` + `def _f(): x = 1` correctly yields `Loaded symbol "x" is unused`, but the same file with `x: int` instead yields no finding; likewise `load(":a.bzl", "T")` + `type T = int` yields nothing while `T = 5` is flagged.

Fix: in the stack walk, also skip an ident whose parent is a statement-level `TypedIdent` or a `TypeAliasStmt`'s `Name`/`TypeParams` — mirroring the `LHSIdent()` exclusion for assignments.

fmeum (commented):
**[🤖 Claude]** Round 4, authored by Claude (Claude Code) at fmeum's request, with fmeum triaging. Every "Fixed by" reply from the earlier threads was verified at head bcc1629 — all check out, including the regression inputs re-run empirically — and `AssignExpr.LHSIdent()` covers the assignment sites correctly (the remaining direct `.LHS` assertions are all call-kwarg/MODULE contexts where a TypedIdent cannot occur). Parser tables regenerate byte-identically, and ~500k fuzz/round-trip probes found no panics, comment loss, or idempotency regressions. Positions stated in the threads (dict reordering, BUILD-mode type-application formatting, cast/isinstance deferral, panic-on-invalid-edit) were respected and not re-reported.

The findings this round share a smaller sibling of the round-3 theme: the assignment fixes don't yet extend to *declarations* (bare var statements, alias names, type params).

Findings on lines outside the diff hunks:

### warn/warn_naming.go:98 — Var statements and type alias names escape the name-conventions check

`nameConventionsWarning` only matches `*build.AssignExpr` statements, so the new declaration forms escape the default-enabled check that identical annotated assignments get. Verified at head: `fooBarQux: int = 1` warns `Variable name should be lower_snake_case ...` but the var statement `fooBarQux: int` yields no finding, and `type fooAlias = int` is also unchecked.

Adding a `TypedIdent` branch restores consistency (`bzlenv.CollectLValues(stmt)` already handles the shape). Type-alias names are worth a deliberate decision though: applying this rule verbatim would reject UpperCamelCase alias names like `Numeric`, so they likely deserve their own convention rather than the variable rule.

### build/print.go:459 — Half-applied LHSIdent migration: printer proxy detection disagrees with edit/bzlmod

`parseUseExtension` (edit/bzlmod/bzlmod.go:248) was converted to `AssignExpr.LHSIdent()` so a typed `use_extension` assignment is recognized, but the printer's parallel proxy detection `usedModuleExtensionProxy` here still does `assign.LHS.(*Ident)` — so the two code paths disagree about what counts as an extension proxy.

With `foo_deps: Ext = use_extension("//:a.bzl", "foo")` + `use_repo(foo_deps, "bar")` in a MODULE.bazel, edit/bzlmod treats `foo_deps` as a proxy but `useSameModuleExtensionProxy`/`compactStmt` (print.go:291) does not, so the statement loses its blank-line grouping with its `use_repo`. Low severity (type annotations in MODULE files are unlikely), but one of the two twins was updated and the other was not — `LHSIdent()` belongs here too.

---
https://claude.ai/code/session_0115t3FEWJipiBKepUbLNL8R

edit/edit.go
fmeum: **[🤖 Claude]** **UsedSymbols counts declaration identifiers as uses, so unused-load misses loads shadowed by declarations**

The exclusion check in the stack walk only recognizes an enclosing `AssignExpr`, so a bare var statement's ident (`x: int`), a `TypeAliasStmt`'s name, and def/alias type-param idents are all counted as *uses* — and `unusedLoadWarning` plus the unusedLoads fix (edit/fix.go:352) silently keep genuinely dead loads. This is the mirror image of the round-2 bug (which deleted used loads); now dead ones are kept.

Verified at head: `load(":a.bzl", "x")` + `def _f(): x = 1` correctly yields `Loaded symbol "x" is unused`, but the same file with `x: int` instead yields no finding; likewise `load(":a.bzl", "T")` + `type T = int` yields nothing while `T = 5` is flagged.

Fix: in the stack walk, also skip an ident whose parent is a statement-level `TypedIdent` or a `TypeAliasStmt`'s `Name`/`TypeParams` — mirroring the `LHSIdent()` exclusion for assignments.

warn/warn_control_flow.go
```diff
@@ -561,33 +658,57 @@ func unusedVariableWarning(f *build.File) []*LinterFinding {
...
-		}
-		left, ok := as.LHS.(*build.Ident)
-		if !ok {
-			continue
-		}
-		if !definedSymbols[left.Name] {
-			definedSymbols[left.Name] = true
-			continue
-		}
+		switch s := s.(type) {
```
fmeum, line 683: **[🤖 Claude]** **Var statements bypass redefined-variable, including the new type-shadowing detection**

The statement switch has no `*build.TypedIdent` case, so var statements neither register a definition nor trigger a redefinition finding — bypassing the type-shadowing detection this PR adds. (This re-surfaces the round-3 thread on TypedIdent/def redefinitions, which never got a reply and is still unfixed at head.)

Verified at head: `type T = int` followed by `T: str` produces no redefined-variable warning, while the equivalent `T: str = 5` correctly warns `Variable "T" has already been defined as a type`; `x = 5` followed by `x: int`, and duplicate `x: int` / `x: str`, are also silently accepted.

A `TypedIdent` case that registers via `GetIdent()` and warns when the name already exists (especially as a type) closes the gap; def-statement redefinitions from the same thread also remain untracked.

bzlenv/bzlenv.go
```diff
@@ -156,6 +171,18 @@ func CollectLValues(node build.Expr) []*build.Ident {
 	return result
 }
 
+// If node is a single identifier or typed identifier, returns the identifier; otherwise, returns nil.
+func GetIdent(node build.Expr) *build.Ident {
```
fmeum: **[🤖 Claude]** **New exported bzlenv.GetIdent is dead code and a third overlapping unwrap helper**

`bzlenv.GetIdent` has zero callers (only its own recursive call), and it joins `TypedIdent.GetIdent` (panics) and `AssignExpr.LHSIdent` (ok-bool) as a third ident-unwrap helper, each with different nil/panic semantics. As exported API it would have to be kept compatible for nothing, and it invites confusion about which helper to use.

Suggestion: delete it — or make it the single canonical helper and implement the others in terms of it.

fmeum (commented):
**[🤖 Claude]** Round 4, authored by Claude (Claude Code) at fmeum's request, with fmeum triaging. (This replaces an earlier round-4 review posted with the wrong finding selection.)

First the verification results: every "Fixed by" reply from the earlier threads checks out at head bcc1629, including the regression inputs re-run empirically, and `AssignExpr.LHSIdent()` covers the assignment sites correctly — the remaining direct `.LHS` assertions are all call-kwarg/MODULE contexts where a TypedIdent cannot occur. Parser tables regenerate byte-identically, and ~500k fuzz/round-trip probes found no panics, comment loss, or idempotency regressions. Positions stated in the threads (dict reordering, BUILD-mode type-application formatting, cast/isinstance deferral, panic-on-invalid-edit) were respected and not re-reported.

The correctness findings this round share a smaller sibling of the round-3 theme: the assignment fixes don't yet extend to *declarations* (bare var statements, alias names, type params).

Findings on lines outside the diff hunks:

### warn/warn_naming.go:98 — Var statements and type alias names escape the name-conventions check

`nameConventionsWarning` only matches `*build.AssignExpr` statements, so the new declaration forms escape the default-enabled check that identical annotated assignments get. Verified at head: `fooBarQux: int = 1` warns `Variable name should be lower_snake_case ...` but the var statement `fooBarQux: int` yields no finding, and `type fooAlias = int` is also unchecked.

Adding a `TypedIdent` branch restores consistency (`bzlenv.CollectLValues(stmt)` already handles the shape). Type-alias names are worth a deliberate decision though: applying this rule verbatim would reject UpperCamelCase alias names like `Numeric`, so they likely deserve their own convention rather than the variable rule.

### build/print.go:459 — Half-applied LHSIdent migration: printer proxy detection disagrees with edit/bzlmod

`parseUseExtension` (edit/bzlmod/bzlmod.go:248) was converted to `AssignExpr.LHSIdent()` so a typed `use_extension` assignment is recognized, but the printer's parallel proxy detection `usedModuleExtensionProxy` here still does `assign.LHS.(*Ident)` — so the two code paths disagree about what counts as an extension proxy.

With `foo_deps: Ext = use_extension("//:a.bzl", "foo")` + `use_repo(foo_deps, "bar")` in a MODULE.bazel, edit/bzlmod treats `foo_deps` as a proxy but `useSameModuleExtensionProxy`/`compactStmt` (print.go:291) does not, so the statement loses its blank-line grouping with its `use_repo`. Low severity (type annotations in MODULE files are unlikely), but one of the two twins was updated and the other was not — `LHSIdent()` belongs here too.

---
https://claude.ai/code/session_0115t3FEWJipiBKepUbLNL8R

edit/edit.go
tetromino: Fixed in 1a04050acf07b52eda7dbd43e0fd9eb2be821fc0 - but note there seems to be a pre-existing problem where we ignore assignments with multiple idents on the LHS. (Added a TODO comment for it.)

warn/warn_control_flow.go
tetromino: I've added a warning for type alias after var statement; while ensuring that assignment after var statement is allowed - c584e7462474120d0266cef31aba59cfbc8a95d6

This raises the interesting question about var statement after an assignment (or a type alias statement). *That* should not be a warning at all, it should be a fatal error (as also explained in my proposed language spec PR), but we cannot reasonably check for it until we have a proper resolver stage after parsing. See https://github.com/bazel-contrib/buildtools/issues/1500 for further discussion.

bzlenv/bzlenv.go
tetromino: Thanks, fixed by 88682af017e3ee255fd5235c820bdfc2f60609c3

build/lex.go
```diff
@@ -850,9 +898,15 @@ func (in *input) order(v Expr) {
 			in.order(v.Result)
 		}
 	case *DefStmt:
+		if v.TypeParams != nil {
+			in.order(v.TypeParams)
+		}
 		for _, x := range v.Params {
 			in.order(x)
 		}
+		if v.Type != nil {
```
fmeum, line 908: **[🤖 Claude]** **A comment before the closing ')' of a def with a return type makes the formatter emit unparseable output**

`in.order()` for DefStmt now orders the return-type node, so a line comment before the closing `)` of a parameter list attaches as a Before-comment of the return type and gets printed mid-header. Valid pre-PR input:

```python
def f(
        x,
        # about return
) -> int:
    pass
```

formats to

```python
def f(
        x) ->
# about return
int:
    pass
```

which fails re-parse with a 3:1 syntax error (same with an empty parameter list; print site is print.go:875). Base moved the comment into the body instead. Since `-> type` is pre-existing syntax, the formatter now corrupts files that were fine before this PR.

edit/edit.go
```diff
@@ -1050,10 +1050,21 @@ func UsedSymbols(stmt build.Expr) map[string]bool {
...
 		for _, e := range stack {
-			if as, ok := e.(*build.AssignExpr); ok {
-				if as.LHS == expr {
+			switch e := e.(type) {
+			case *build.AssignExpr:
+				// TODO: should we be checking all CollectLValues here?
+				if ident, ok := e.LHSIdent(); ok && ident == expr {
+					return
+				}
+			case *build.TypedIdent:
```
fmeum, line 1066: **[🤖 Claude]** **unused-load still misses loads shadowed by type parameters — the 1a04050 fix is incomplete**

1a04050 excludes `TypedIdent` and `TypeAliasStmt` name idents from `UsedSymbols`, but not def/alias *type-param* idents (which the round-4 comment listed alongside them), so loads shadowed by type parameters still count as uses. Verified at head:

```python
load(":f.bzl", "T")

def foo[T](x: int) -> int:
    return x
```

→ 0 unused-load findings (likewise `load(":f.bzl", "K")` + `type _D[K] = dict[K, int]`), while `T = 5` correctly warns — dead loads are silently kept. The new test in warn_control_flow_test.go currently enshrines the gap by expecting `s7` to be kept.

warn/warn_control_flow.go
```diff
@@ -380,19 +381,33 @@ func extractIdentsFromStmt(stmt build.Expr) (assigned, used map[*build.Ident]boo
 			}
 			used[expr] = true
 
+		case *build.TypedIdent:
```
fmeum, line 384: **[🤖 Claude]** **The `_: Any` enable-type-checking marker is told to delete itself**

The `_: Any` marker — the idiom this PR's own testdata and buildozer test showcase — is flagged by the default-on unused-variable warning:

```python
_: Any  # Enable type checking
```

→ `:1: Variable "_" is unused. Please remove it.` A user following the advice silently disables type checking.

To be fair to the baseline: base flags `_ = 5` the same way, so this is pre-existing `_` handling extended verbatim to the new var-statement path. But since the PR promotes this exact statement as the opt-in mechanism, the new TypedIdent path (lines 384–397, which only consults `# @unused`) probably deserves a `_` special case — or the docs should bless `# @unused` as part of the idiom.

warn/warn_control_flow.go
```diff
@@ -473,6 +506,39 @@ func unusedVariableCheck(f *build.File, root build.Expr) (map[string]bool, []*Li
...
+				// The type alias name is defined in the current scope
+				typeName := expr.GetIdent().Name
+				if _, ok := definedTypes[typeName]; !ok {
+					definedTypes[typeName] = expr.Name
+				}
+				if edit.ContainsComments(expr, "@unused") {
+					suppressedWarnings[typeName] = true
+				}
+
+				usedSymbolsInTypeAlias, findingsInTypeAlias := unusedVariableCheck(f, expr)
```
fmeum, line 533: **[🤖 Claude]** **A self-referential type alias marks itself used**

The recursive call merges the alias's own used symbols into the parent scope without excluding the alias name, so a type alias whose definition mentions its own name is never reported unused. Verified at head:

```python
type _Tree = dict[str, _Tree]
```

(also `type _T = _T`, and the mutual pair `type _A = _B` / `type _B = _A`) → 0 unused-variable findings even though nothing else references the alias, while `type _T = int` alone correctly warns. Recursive aliases are the normal way to write tree-shaped types, so unused ones will be common.

warn/warn_control_flow.go
```diff
@@ -443,7 +463,13 @@ func unusedVariableCheck(f *build.File, root build.Expr) (map[string]bool, []*Li
 			}
 
 			// The function is a root for the current scope.
-			// Collect its parameters as defined in the current scope.
+			// Collect its type parameters.
+			if expr.TypeParams != nil {
+				for _, typeParam := range bzlenv.CollectLValues(expr.TypeParams) {
+					definedTypes[typeParam.Name] = typeParam
```
fmeum, line 469: **[🤖 Claude]** **Unused type-parameter findings have no conventional suppression path**

An underscore prefix does not suppress unused-type-param findings (unlike parameters and variables), and `# @unused` on the def or alias does not reach its type params. Verified at head:

- `def _f[_T](_x: int): pass` → `:1: Type "_T" is unused` while `_x` is exempt.
- `# @unused` before `def _f[T]()` or `type _Opt[_T] = int` still reports the type param.

Only the blunt `# buildifier: disable=unused-variable` silences it — inconsistent with every other symbol kind in this check.

warn/warn_naming.go
```diff
@@ -93,25 +93,40 @@ func nameConventionsWarning(f *build.File) []*LinterFinding {
 	var findings []*LinterFinding
 
 	build.WalkStatements(f, func(stmt build.Expr, stack []build.Expr) (err error) {
```
fmeum, line 107: **[🤖 Claude]** **name-conventions applies no convention to type parameters**

The new switch enforces UpperCamelCase on type alias names but nothing on type parameters of defs or aliases: `type my_int = int` → `Type name "my_int" should be UpperCamelCase`, while `def f[t_x](x: t_x) -> t_x:` and `type Foo[t_x] = list[t_x]` → 0 findings (verified at head). Type-name style is enforced for aliases yet unenforced for type params appearing in the same annotations — presumably the same UpperCamelCase rule should apply.

warn/warn_control_flow.go
```diff
@@ -561,33 +658,69 @@ func unusedVariableWarning(f *build.File) []*LinterFinding {
...
+			if _, ok := definedSymbols[ident.Name]; !ok {
+				definedSymbols[ident.Name] = DeclaredVariable
+				continue
+			}
+			// TODO(#1500): don't report a warning for a var statement that redeclares a global variable because
+			// 1. this cannot be a warning that could be ignored - it is a static fatal error; and
+			// 2. it must be a fatal error for local variables too (not just globals).
+			// To fix, we need a proper resolver stage after parsing. See https://github.com/bazel-contrib/buildtools/issues/1500
+
+		case *build.TypeAliasStmt:
```
fmeum, line 719: **[🤖 Claude]** **After a variable is redefined as a type, += still gets the list-append exemption**

The TypeAliasStmt redefinition path warns but never updates `definedSymbols` to TypeAlias, so the name keeps its AssignedVariable kind:

```python
T = []
type T = int
T += [1]
```

→ only line 2 warns; line 3 is silent because `definedSymbols["T"]` stayed AssignedVariable and the `+=` list-append exemption applies — whereas the PR's own test expects `type T = int` + `T += [42]` to warn. Updating the kind on the warning path (~line 713) makes line 3 consistent.

fmeum (commented):
**[🤖 Claude]** Round 5, authored by Claude (Claude Code) at fmeum's request, with fmeum triaging (correctness and documentation findings only this round; style/cleanup items were held back).

Fix verification at head c584e74: 88682af, 9d17571, 21b558c, and c584e74 all hold as described in the threads (the remaining `LHS.(*Ident)` assertions are kwarg-only contexts). 1a04050 is partial — see the unused-load finding below. Positions recorded in the threads (cast/isinstance follow-up, dict reordering, BUILD-mode type-application formatting, the Expr-widening/panic stance, value-domain type params, and the #1500 resolver-stage deferral) were respected and not re-reported.

The two findings at the top are file-corrupting regressions reachable from pre-existing syntax; the rest are lint gaps in the new type support plus one documentation gap.

Findings on lines outside the diff hunks:

### warn/warn_bazel_api.go:665 — The native.package_name() autofix rewrites idents inside type annotations into unparseable code

Because WalkOnce now traverses `TypedIdent.Type`/`DefStmt.Type`, `globalVariableUsageCheck` reaches idents inside type annotations and rewrites them: `buildifier --lint=fix` on

```python
def f(x: PACKAGE_NAME):
    return x
```

yields `def f(x: native.package_name()):` — a call expression the type grammar rejects, so re-parse fails with `1:30: syntax error near (`. Same for `x: REPOSITORY_NAME = 1` var annotations. This is a default-on fix corrupting the file; at base, annotations were never traversed.

Beyond the point fix (skip idents in type position here), this raises a broader question worth a systematic pass: every autofix that rewrites expressions now potentially reaches annotation position, and any that produces non-type_expr output there will corrupt the file the same way.

### warn/warn_naming.go:45 — confusing-name never sees names introduced by the new syntax

`collectLocalVariables` skips `TypeAliasStmt`, and the `DefStmt` handling checks `Params` but not `TypeParams`, so the banned ambiguous names `l`/`I`/`O` pass when introduced via type syntax. Verified at head: `I = 5` → confusing-name finding, but `type I = int` → 0 findings and `def f[l](x: l) -> l:` → 0 findings — the exact identifiers the lint exists to ban escape when declared as aliases or type params.

### WARNINGS.md:815 — WARNINGS.md not updated for the new type-aware lint behaviors

The new lint behaviors are undocumented: name-conventions' Type-name rule, redefined-variable's redefined-as-a-type semantics, and unused-variable's `Type ... is unused` findings — including their suppression story. A user hit by `Type name "my_int" should be UpperCamelCase` or `Type "_T" is unused` finds only the old variable/provider rules in the documented sections, with no mention of type aliases, type params, or how to suppress the new findings. (Documenting the suppression story would also force a decision on the `# @unused`/underscore gaps reported above.)

---
https://claude.ai/code/session_0115t3FEWJipiBKepUbLNL8R

build/lex.go
tetromino: This is a variation of pre-existing issue: buildifier already misformats

```
def f(
        x,
        # about return
):
    pass
```

as

```
def f(
        x):
    # about return
    pass
```

I'll see if there's a simple fix though, since until now, it didn't lead to invalid syntax.

tetromino: Tricky issue, but fixed by 558b083b30ac0ac303ade1511bdf7ff337e84cd1

edit/edit.go
tetromino: Fixed by 04f12461c5f8f66bb848d66b05a2618f46173ff9

warn/warn_control_flow.go
tetromino: Good call, fixed by daab8eec7c2ca9a3dd56605752db102dd016837f

warn/warn_control_flow.go
tetromino: Bazel doesn't support recursive type aliases, and there are no plans to add such support. I'll add a TODO if this ever becomes a problem in the future.

warn/warn_control_flow.go
tetromino: I originally thought that unused type parameters are always a mistake, but I suppose they could be useful for backwards compat reasons. Fixed by c3b9e97051eab0d01f7a8a589473205f74802f12

warn/warn_naming.go
tetromino: We don't enforce any convention for function parameter names, so I wouldn't worry about type parameter names for now.

warn/warn_control_flow.go
tetromino: IMHO the first warning is sufficient.

fmeum (commented):
**[🤖 Claude]** Round 6, authored by Claude (Claude Code) at fmeum's request, with fmeum triaging. This round focused on lint/format operations emitting invalid or semantically different Starlark. Good news first: a systematic sweep of every autofix against every node kind the type grammar can produce — plus empirical `--lint=fix --warnings=all` runs over kitchen-sink corpora and repo-wide base-vs-head differentials — found **no** remaining autofix that corrupts or changes the semantics of type syntax beyond the package-name case already fixed in 213f073. Of the eight fix commits, six hold as claimed (558b083's original repro, daab8ee, c3b9e97, 42db8f4, 828f564's lint half, 213f073); the two findings below are regressions introduced by 558b083's comment-anchor mechanism itself, both re-verified at head 97d8894.

---
https://claude.ai/code/session_017ByRvLmLeDbRZJkxKv6xdd

build/lex.go
```diff
@@ -850,9 +898,17 @@ func (in *input) order(v Expr) {
...
+			in.order(v.TypeParams)
+		}
 		for _, x := range v.Params {
 			in.order(x)
 		}
+		in.order(&v.ParamsEnd)
+		if v.Type != nil {
+			in.order(v.Type)
+		}
+		in.order(&v.ColonPos)
```
fmeum: **[🤖 Claude]** **`# buildifier: disable=...` on a def line silently stops suppressing warnings**

558b083 anchors def-header end-of-line comments to `DefStmt.ParamsEnd`/`ColonPos` — but those `End`/`Position` nodes are never visited by `build.Walk`/`WalkOnce` (the DefStmt case in build/walk.go:213 visits only TypeParams/Params/Type/Body), so `warn.DisabledWarning`, `skylarkCommentWarning`, and `edit.ContainsComments` no longer see such comments. On a plain type-free file:

```python
def foo(x):  # buildifier: disable=unused-variable
    pass
```

base: no findings; head: `Variable "x" is unused` — the suppression is silently ignored (same for `disable=function-docstring`, `disable=unnamed-macro`, etc.), so existing repos' suppressed warnings resurface on upgrade. Likewise `def f(x):  # skylark thing` loses the skylark-comment warning and its autofix.

Re-verified at 97d8894. Fix: have WalkOnce (or the comment consumers) visit the ParamsEnd/ColonPos comments.

build/print.go
```diff
@@ -882,12 +862,21 @@ func (p *printer) expr(v Expr, outerPrec int) {
...
-		p.seq("()", &v.StartPos, &v.Params, nil, modeDef, v.ForceCompact, v.ForceMultiLine)
+		p.optionalTypeParams(v.TypeParams)
+
+		// Line anchor for the opening '(' of the params list
+		paramsAnchor := v.StartPos
+		if v.TypeParams != nil {
+			_, paramsAnchor = v.TypeParams.Span()
+		}
+
+		p.seq("()", &paramsAnchor, &v.Params, &v.ParamsEnd, modeDef, v.ForceCompact, v.ForceMultiLine)
```
fmeum: **[🤖 Claude]** **Defs with the closing ')' on its own line no longer collapse to compact form**

A side effect of the same comment-anchor change: passing `&v.ParamsEnd` as the seq end anchor changes `useCompactMode`'s keep-original-style decision for defs whose closing `)` is on its own line (the end anchor now participates in the same-line comparison at print.go:~1070):

```python
def f(x,
      ):
    pass
```

base: `def f(x):`; head (re-verified at 97d8894):

```python
def f(
        x):
    pass
```

The output is valid and idempotent, and already-formatted files are unaffected, so this only churns first-time formatting — but it's an untested behavior change with no golden covering the shape. Either add a golden enshrining the new behavior as intended, or exclude modeDef's end anchor from the same-line comparison.

build/lex.go
tetromino: Fun fact: many pre-existing syntax nodes already use `End` fields (in ListExpr, DIctExpr, etc.) which are ignored by walkers and therefore by warn.DisabledWarning and skylarkCommentWarning.

The only way to fix, as far as I see, is to turn End-s into Expr-s, and then assert that they are of the correct type when trying to use them...

tetromino: Fixed DefStmt by 5aa9ed6d1c394563db1755769f3551d44058c29f

Pre-existing issues in the other syntax nodes are imho out of scope for this pr...

build/print.go
tetromino: Fair point, since in modeDef we don't print a trailing comma.

Fixed by f9a7512aa491189bb3256f455a7a8075b46e0dc5 without a test (we cannot test for this in print_test's testdata because compactifying a def statement's header changes ForceMultiline on reparsing and therefore triggers an eqchecker failure; and it's not clear to me that eqchecker should always ignore ForceMultiline).

fmeum (approved):

PR #1494, opened 2026-08-19
Add exports-files-discouraged buildifier lint check

alexeagle:
## Summary

`exports_files` is a design smell: it leaks an InputArtifact to be in the srcs of an unrelated Bazel package. This breaks the principle of locality that allows BUILD files to be maintained in isolation.

With agents writing more BUILD files I've noticed this pattern increasingly used, when the right answer is to move the rule that needs to consume the source file.

## Detail

- Adds a new default-on buildifier lint check `exports-files-discouraged` that flags `exports_files()` as a package encapsulation violation ("feature envy").
- When `exports_files()` is present, the check also warns if visibility is missing or not scoped to a package (`:__pkg__` / `:__subpackages__`) or `package_group`, and if there is no comment explaining why encapsulation is being violated.
- Suggests exposing files via a rule (e.g. `filegroup`) so the owning package can govern access and later swap the label for a generated artifact.

## Test plan

- [x] `bazel test //warn:warn_test --test_filter=TestExportsFilesDiscouraged`
- [x] `bazel test //warn/docs:docs_test`
- [x] Verified disable comment `# buildifier: disable=exports-files-discouraged` suppresses all findings


warn/warn_bazel_test.go
```diff
@@ -262,6 +262,63 @@ py_binary(
...
+exports_files(["foo.txt"])
+`,
+		[]string{
+			`:1: exports_files() violates package encapsulation`,
+			`:1: exports_files() defaults to //visibility:public`,
+			`:1: exports_files() should include a comment explaining why package encapsulation is violated`,
+		},
+		scopeBuild)
+
+	checkFindings(t, "exports-files-discouraged", `
```
alexeagle, line 276: I think in this case, the comment and visibility should be enough that buildifier exits zero

warn/warn_bazel.go
```diff
@@ -278,6 +278,100 @@ func externalPathWarning(f *build.File) []*LinterFinding {
...
+	if len(rule.Call.List) >= 2 {
+		if _, ok := rule.Call.List[1].(*build.AssignExpr); ok {
+			return nil
+		}
+		return build.Strings(rule.Call.List[1])
+	}
+	return nil
+}
+
+func isPackageOrPackageGroupVisibility(label string) bool {
```
alexeagle, line 305: the root package would have to use `visibility = //:__subpackages__` - that's probably okay.

Alternatively maybe this rule should not evaluate the root BUILD file since it's common that linters and other repo-wide tooling need their config files in the root?

PR #1505, opened 2026-09-03
Add cast and isinstance keywords

tetromino:
Finishing support for the type syntax proposed in https://github.com/bazelbuild/starlark/pull/362

After some reflection, I think we should always parse cast/isinstance as keywords (requiring 2 arguments, one of which is a value expression and the other a type expression) since permitting pre-type Starlark syntax is likely to only lead to confusion, and would be quite difficult to get right.

Note that this is an **incompatible change**: forms like `cast(x, y, z)` or `cast = foo` will now be a parse error. However, in practice, I observe no occurrences of `cast` or `isinstance` used as identifiers in any Starlark code in Google's monorepo, so I believe that it's rather unlikely to break anyone.

Store the parsed cast/isinstance exprs as `CallExpr`-s (to avoid diverging their formatting/walking/etc. support from call expressions) with an enum field to mark that they are of a special kind.

## Buildtools PR checklist

- [X] The code in this PR is covered by unit/integration tests.
- [X] I have tested these changes and provide testing instructions below.
- [X] I have read Google Eng Practices on [Small Changes](https://google.github.io/eng-practices/review/developer/small-cls.html), this PR either follows these guidelines or the description provides reasoning for why they can not be followed.


fmeum (approved):

fmeum:
I unfortunately missed that this didn't end up merging for v10. I will send a v10.0.2 with this and argue for it to be a bugfix for type support, but call out the incompatible change.

fmeum (approved):

fmeum:
@tetromino I rebased the PR onto main and will merge it now.

PR #1510, opened 2026-09-11
buildozer: do not remove visibility from macros or loaded rules in fix

kyledobitz:
Previously, buildozer's `removeVisibility` routine stripped visibility attributes (such as `"//visibility:private"` or matching the package default_visibility) from any rule whose visibility matched `defaultVisibility`.

However, Starlark macros and loaded rules frequently implement custom fallback visibility logic (e.g. falling back to a non-private default if visibility is omitted). Stripping visibility from such macros causes them to silently broaden their visibility.

This change skips visibility removal if a rule's kind contains a dot (indicating a dotted macro or module) or matches any symbol loaded via a `load()` statement in the BUILD file.

Unit tests are included in `edit/fix_test.go` covering:
- Redundant visibility removal for native rules
- Preserving visibility for dotted macros
- Preserving visibility for loaded macros
- Package `default_visibility` handling for native rules vs loaded macros


edit/fix.go
```diff
@@ -125,6 +132,28 @@ func removeVisibility(f *build.File, r *build.Rule, pkg string) bool {
...
+		return false
+	}
+	kind := r.Kind()
+	if strings.Contains(kind, ".") {
+		return true
+	}
+	for _, stmt := range f.Stmt {
+		if load, ok := stmt.(*build.LoadStmt); ok {
+			for _, to := range load.To {
+				if to.Name == kind {
```
fmeum, line 147: With native rules being Starlarkified, wouldn't this match almost every target?

fmeum (commented):
Do you have an example of a macro that would benefit from this change in behavior? Symbolic macros do not need it and legacy macros should use private visibility for inner targets.

PR #1512, opened 2026-09-14
only remove visibility matching explicit default_visibility

oreflow:
## Buildtools PR checklist

## Description

Previously, buildozer's fix.go:105 routine stripped `visibility` attributes whenever they matched an explicit `default_visibility` or defaulted to `["//visibility:private"]` when `default_visibility` was omitted.

However, Starlark macros and custom rules often implement their own default visibility logic (e.g., defaulting to public or custom targets when `visibility` is omitted). Stripping `visibility = ["//visibility:private"]` when no package-level default is declared can silently broaden visibility.

Rather than relying on fragile heuristics to identify macros and loaded rules, this change refactors fix.go:105 to only strip visibility attributes when `default_visibility` is **explicitly defined** in a `package()` declaration:

 - **Explicit `default_visibility` required**: If `default_visibility` is not explicitly set on the package, rule `visibility` attributes are preserved.
 - **Unordered string list matching**: Compares string lists using `slices.Sorted`, allowing redundant visibility to be removed even if list elements appear in a different order.
 - **Identifier matching**: Supports matching identifier expressions (e.g. `package(default_visibility = PUBLIC)` and `visibility = PUBLIC`).

### (optional) These changes were tested using the following steps

Ran unit tests covering all cases:
```bash
    go test ./edit -run TestFix
```

Unit test cases in fix_test.go verify:

  • Preserving visibility when default_visibility is not set
  • Removing redundant visibility with differing slice order
  • Removing redundant visibility matching identifiers
  • Preserving non-matching visibility


fmeum (approved):

fmeum:
Let's wait on merging this until we have fixed #1518.

PR #1518, opened 2026-09-15
Fix @unused for last argument of function

keith:
Regressed in d2a20dc9d54727597d66fd091cb16711375cac2b


keith:
cc @tetromino in case you have thoughts of if this behavior should change because of the types stuff

fmeum:
Looks like current main may emit this by turning:
```
def f(
      x,  # @unused
  ):
      pass
```

into:

```
  def f(
          x):  # @unused
      pass
```

I will send a fix for that.

fmeum:
I see how this is a regression, but the previous behavior is actually surprising to me. If anything, I would have expected that to declare the function itself as unused? With types trailing the function definition, this would be even less consistent.

Did you encounter this pattern in the real world or only due to buildifier rewriting it first?

keith:
it's in the real world: https://github.com/search?type=code&q=language%3Astarlark+%2F%5C)%3A.*%40unused%2F

keith:
i think if we want to go the other way we probably want to format `):` onto a new line always? and during that formatting today it would move the unused to the wrong spot, but then once the user moved it at least it would be fixed

fmeum:
Okay, with that amount of real world usage of this pattern I don't think we can realistically change this behavior.

warn/warn_control_flow_test.go
```diff
@@ -654,6 +654,23 @@ foo()
 		},
 		scopeEverywhere)
 
+	checkFindings(t, "unused-variable", `
+def foo(x):  # @unused
+  pass
+
+def bar(
+    x,
+    y):  # @unused
```
fmeum, line 663: Could you add another test in which this also has a trailing return type?

fmeum (approved):

PR #1519, opened 2026-09-16
Preserve comments on the final function parameter

fmeum:
Buildifier currently moves a trailing comment on the final function parameter past the closing `)` and any return annotation. For example:

```python
def f(
    x,  # @unused
) -> int:
    return 1
```

becomes:

```python
def f(
        x) -> int:  # @unused
    return 1
```

Since #1488, these two spellings have different AST attachments: the first comment belongs to `x`, while the second belongs to the function header's colon. #1518 intentionally interprets an `@unused` comment on the header as applying to the final parameter, preserving the longstanding behavior of existing files. However, buildifier should not convert the preferred parameter-attached form into that legacy header-attached form. Keeping the forms distinguishable lets users write the clearer form today and leaves room to change or deprecate the legacy interpretation in the future.

This PR keeps signatures with parameter suffix comments multiline and emits a pending final parameter comment before the closing parenthesis:

```python
def f(
        x  # @unused
) -> int:
    return 1
```

Both forms remain supported, with `@unused` applying only to the final parameter:

```python
# Preferred: the comment is attached to y.
def f(
        x,
        y  # @unused
):
    pass

# Legacy compatibility: a header comment applies to y, not x.
def g(x, y):  # @unused
    pass
```

The lexer previously treated a comment following a line containing only `):` as a body comment. That exception was added in #276 before block-header comments had their own AST node. Remove it so comments visibly written after the colon remain attached to the header, including multiline `def`, `if`, and `for` headers.

Finally, flush queued end-of-line comments before reducing the indentation margin. This keeps continuation comments on a final parameter aligned with that parameter and makes formatting idempotent.

The `unused-variable` documentation now recommends the parameter-attached form and documents the header-attached form as backward compatibility behavior.


fmeum:
@tetromino

build/print_test.go
```diff
@@ -528,6 +528,51 @@ func (eq *eqchecker) checkValue(v, w reflect.Value) error {
 	return nil
 }
 
+func TestPrintDefParameterComments(t *testing.T) {
+	for _, param := range []string{"x", "x = None", "x: int", "x: int = 0", "*args", "**kwargs"} {
+		for _, returnType := range []string{"", " -> int"} {
+			headerComment := ""
+			if returnType != "" {
+				headerComment = "  # header"
+			}
```
tetromino: We should test for the header comment even if there's no return type annotation.

tetromino (approved):
LGTM for the printing fix - thank you for that!

What I'm a bit concerned about is the semantics of the comments - is the `@unused` on the last parameter treated as applying only to that parameter or all parameters? (I would expect only to that parameter, but I worry it might be applying to all.)

fmeum:
I worked on this a bit more and massaged it into a form that keeps the clearer "parameter-attached" annotation as is without dropping support for the "function-attached" annotation.

@keith What do you think?

tetromino (approved):
Thanks! Finally we are formatting post `:` comments as expected :)

The backwards-compatible behavior re. applying to last comment is certainly a wart, but we have to live with it because a fair bit of code I see relies on it...

